PR #14, opened 2020-08-27
Initial Plan Page

gildub:
Addresses https://github.com/konveyor/virt-ui/issues/8

### DIFF
--- a/src/app/Plans/PlansPage.tsx
+++ b/src/app/Plans/PlansPage.tsx
@@ -16,6 +16,7 @@ import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import { PlusCircleIcon } from '@patternfly/react-icons';
 import PlansTable from './components/PlansTable';
 import { IAddPlanDisabledObjModel } from './types';
+import { Provider } from '@app/Providers/types';
 import { ProviderType } from '@app/common/constants';
 
 // TODO replace these with real state e.g. from redux
@@ -26,8 +27,8 @@ const migplans = MOCK_PLANS;
 const providers = MOCK_PROVIDERS;
 
 const PlansPage: React.FunctionComponent = () => {
-  const vmwareList = providers.map((x) => x.spec.type === ProviderType.vsphere);
-  const cnvList = providers.map((x) => x.spec.type === ProviderType.cnv);
+  const vmwareList: Provider[] = providers.filter((x) => x.spec.type === ProviderType.vsphere);
+  const cnvList: Provider[] = providers.filter((x) => x.spec.type === ProviderType.cnv);
 
   const [isWizardOpen, toggleWizard] = React.useReducer((isWizardOpen) => !isWizardOpen, false);
 
--- a/src/app/Plans/PlansPage.tsx
+++ b/src/app/Plans/PlansPage.tsx
@@ -27,7 +27,7 @@ const PlansPage: React.FunctionComponent = () => {
   const vmwareList = providers.map((x) => x.spec.type === ProviderType.vsphere);
   const cnvList = providers.map((x) => x.spec.type === ProviderType.cnv);
 
-  const [isWizardOpen, toggleWizard] = React.useReducer(() => !isWizardOpen, false);
+  const [isWizardOpen, toggleWizard] = React.useReducer((isWizardOpen) => !isWizardOpen, false);
 
   let addPlanDisabledObj: IAddPlanDisabledObjModel = {
     isAddPlanDisabled: false,
--- a/src/app/Plans/PlansPage.tsx
+++ b/src/app/Plans/PlansPage.tsx
@@ -1,14 +1,125 @@
 import * as React from 'react';
-import { PageSection, Title } from '@patternfly/react-core';
+import {
+  Card,
+  PageSection,
+  CardBody,
+  EmptyState,
+  EmptyStateIcon,
+  Title,
+  Button,
+  Bullseye,
+  Spinner,
+  Tooltip,
+} from '@patternfly/react-core';
+import { AddCircleOIcon } from '@patternfly/react-icons';
+import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
+import PlansTable from './components/PlansTable';
+import { IAddPlanDisabledObjModel } from './types';
 
-const PlansPage: React.FunctionComponent = () => (
-  <>
-    <PageSection>
-      <Title headingLevel="h1" size="lg">
-        Plans Page Title
-      </Title>
-    </PageSection>
-  </>
-);
+// TODO replace these with real state e.g. from redux
+import { MOCK_PLANS } from './mocks/plans.mock';
+import { MOCK_PROVIDERS } from '@app/Providers/mocks/providers.mock';
+const IsFetchingInitialPlans = false; // Fetching for the first time, not polling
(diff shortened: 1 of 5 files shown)


src/app/Plans/PlansPage.tsx
```diff
@@ -1,14 +1,125 @@
...
+
+// TODO replace these with real state e.g. from redux
+import { MOCK_PLANS } from './mocks/plans.mock';
+import { MOCK_PROVIDERS } from '@app/Providers/mocks/providers.mock';
+const IsFetchingInitialPlans = false; // Fetching for the first time, not polling
+const migplans = MOCK_PLANS;
+const providers = MOCK_PROVIDERS;
+
+const PlansPage: React.FunctionComponent = () => {
+  const VMWareList = providers.map((x) => x.spec.type === 'vsphere');
```
mturley, line 27: The capitalization of `VMWareList` and `CNVList` here is odd for an array of data. I think we should reserve capitalized names for modules like components and types. So these could be `vmwareList` and `cnvList`.

Also you can `import { ProviderType } from '@app/common/constants';` and use `ProviderType.vsphere` and `ProviderType.cnv` instead of the plain strings, if you want. Although I think TS validates those either way because of the type of `providers`.

src/app/Plans/PlansPage.tsx
```diff
@@ -1,14 +1,125 @@
...
+  const CNVList = providers.map((x) => x.spec.type === 'cnv');
+
+  const [isWizardOpen, toggleWizard] = React.useReducer(() => !isWizardOpen, false);
```
mturley, line 30: This is really nice and clever!

src/app/Plans/PlansPage.tsx
```diff
@@ -1,14 +1,125 @@
...
+
+  const [addPlanDisabledObj, setAddPlanDisabledObj] = React.useState<IAddPlanDisabledObjModel>({
```
mturley, line 32: I know this stuff was copied from mig-ui, but this is one of the `useEffect` calls that isn't necessary. The only place this state changes is immediately after it is initialized. So `addPlanDisabledObj` doesn't need to be state, it can just be a variable that is initialized on each render. So we could replace lines 32-58 here with this:

```ts
let addPlanDisabledObj: IAddPlanDisabledObjModel = {
  isAddPlanDisabled: false,
  disabledText: 'Create a migration plan to select VMs to migrate to OpenShift virtualization.',
};

if (vmwareList.length < 1) {
  addPlanDisabledObj = {
    isAddPlanDisabled: true,
    disabledText: 'At least 1 VMWare provider is required to create a plan.',
  }
}

if (cnvList.length < 1) {
  addPlanDisabledObj = {
    isAddPlanDisabled: true,
    disabledText: 'At least 1 OpenShift provider is required to create a plan.',
  }
}
```

mturley: Note that I also changed the capitalization of `VMWare` to `VMware` (lowercase W), apparently we've been getting that wrong :)

gildub: Great, I love this. I effectively had been wondering about the state of `addPlanDisableObj` and assumed there were cases where the backend would be changed behind the curtain and we would need this to be reflected in the displayed screen. Thanks for clarifying.

The VMware capitalization has been hunting everyone (I know it's not just me) for years. It has to stop!!! ;)

src/app/Plans/PlansPage.tsx
gildub: Thanks for `@app/common/constants` pointer.

src/app/Plans/PlansPage.tsx
gildub: Thank you!

vconzola:
Hi @gildub A couple comments...
1. Remove the border around the empty state component.
2. There should be 24px padding between icon and heading text
3. Missing the subtext in the mockup that goes below the heading
4. There should be 16px padding between the heading text and the subtext
5. There should be 32px padding between the subtext and the icon.

Thanks.

mturley:
Ah sorry I missed the visual things, I had just quickly reviewed the code. @gildub let me know if you need help with the padding stuff, we shouldn't hard-code pixel sizes, the pixel sizes Vince describes correspond to spacers in PatternFly: https://www.patternfly.org/v4/design-guidelines/styles/spacers You can see where I use those in some places via `import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing'`. 

@vconzola, does this look right for the empty state of the Providers page? If so, @gildub you can imitate the code of my ProvidersPage file when `providers.length === 0`. I think you need to put an `<EmptyStateBody>` above the button, with the body text from [this slide of the mockups](https://docs.google.com/presentation/d/1YtiJKf5FyDhBf_eIc6pXbHloCSidDmBKATgYXRtPh-o/edit#slide=id.g878b3b7f84_0_68) ("Create a migration plan to select VMs..."), and that should fix the spacing above and below that text since spacing is built in to `<EmptyStateBody>`.

mturley:
> @vconzola  does this look right for the empty state of the Providers page?

Forgot my screenshot.

Screen Shot 2020-08-28 at 12 28 42 PM

mturley:
@gildub I'm also about to implement basically this same PR for the mappings page, once I have that ready we can compare and maybe identify common components to abstract (although it's pretty lightweight).

gildub:
@vconzola, @mturley thank you for you reviews, I probably should have pushed this one as WIP (draft) meanwhile the back and forth about the look & feel is likely normal UI work flow. Anyway, I'll get back onto it first thing on Monday.

@vconzola, one thing missing are the pagination details, although it probably doesn't matter when there are no objects displayed, but that's what the UI Mockups is showing, so is it needed or not (of course outside of the case when there mig plans)?

@mturley, I'll readjust from your code.

mturley:
@gildub the pagination is something Vince and I discussed that we should have surfaced somewhere. We decided the pagination controls should be shown only if there are providers/mappings/plans to show, so they shouldn't be in this empty state, but they could still appear with different empty state (TableEmptyState from mig-ui) if the user added filters that don't match anything. That will only become relevant once we get around to using FilterToolbar in this page (I need to finish moving that into lib-ui as well).

mturley:
@gildub here's my stuff for the mappings pages. You can pretty much copy the structure of NetworkMappingsPage or StorageMappingsPage from here: https://github.com/konveyor/virt-ui/pull/16 you're pretty close, though. I think it's just the color variants on the PageSections and the missing EmptyStateBody.

src/app/Plans/PlansPage.tsx
```diff
@@ -1,14 +1,109 @@
...
+          <Card>
+            <CardBody>
+              {!migplans ? null : migplans.length === 0 ? (
+                <EmptyState className={spacing.my_2xl}>
+                  <EmptyStateIcon icon={PlusCircleIcon} />
+                  <Title size="lg" headingLevel="h2">
+                    No migration plans
+                  </Title>
+                  <EmptyStateBody>
+                    Create a migration plan to select VMs to migrate to OpenShift virtualization.
```
mturley, line 81: @vconzola told me should now be capitalizing the V in "OpenShift Virtualization".

vconzola:
@gildub Sorry for the slow reply - I was on PTO Friday. The Mappings empty states look good now. Thanks for making the changes.

PR #15, opened 2020-08-28
Initial Plan wizard

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/9

### DIFF
--- a/src/app/Plans/components/Wizard/GeneralForm.tsx
+++ b/src/app/Plans/components/Wizard/GeneralForm.tsx
@@ -1,0 +1,92 @@
+import * as React from 'react';
+import { Form, FormGroup, TextArea, TextInput, Title } from '@patternfly/react-core';
+import SimpleSelect, { OptionWithValue } from '@app/common/components/SimpleSelect';
+import { Provider } from '@app/Providers/types';
+
+interface IGeneralFormComponent {
+  srcProviders: Provider[];
+  tgtProviders: Provider[];
+}
+
+const GeneralForm: React.FunctionComponent<IGeneralFormComponent> = ({
+  srcProviders,
+  tgtProviders,
+}: IGeneralFormComponent) => {
+  const [planName, setPlanName] = React.useState<string>('');
+  const [planDesc, setPlanDesc] = React.useState<string>('');
+  const [srcProvider, setSrcProvider] = React.useState<Provider | null>(null);
+  const [tgtProvider, setTgtProvider] = React.useState<Provider | null>(null);
+
+  const srcProvidersOptions = Object.values(srcProviders).map((provider) => ({
+    toString: () => provider.metadata.name,
+    value: provider,
+  })) as OptionWithValue<Provider>[];
+
+  const tgtProvidersOptions = Object.values(tgtProviders).map((provider) => ({
+    toString: () => provider.metadata.name,
+    value: provider,
+  })) as OptionWithValue<Provider>[];
+
+  return (
+    <Form>
+      <Title headingLevel="h1" size="md">
+        Give your plan a name and a description
+      </Title>
+      <FormGroup
+        label="Plan name"
+        isRequired
+        fieldId="planName"
+        helperTextInvalid="TODO"
+        validated="default" // TODO add state/validation/errors to this and other FormGroups
+      >
+        <TextInput id="planName" value={planName} type="text" onChange={setPlanName} />
+      </FormGroup>
+
+      <FormGroup
+        label="Plan description"
+        fieldId="planDesc"
+        helperTextInvalid="TODO"
+        validated="default" // TODO add state/validation/errors to this and other FormGroups
+      >
+        <TextArea id="planDesc" type="text" value={planDesc} onChange={setPlanDesc} />
+      </FormGroup>
+
+      <Title headingLevel="h3" size="md" /* className={styles.fieldGridTitle} */>
+        Select source and target providers
+      </Title>
+      <FormGroup
+        label="Source Provider"
+        isRequired
+        fieldId="sourceProvider"
+        helperTextInvalid="TODO"
+        validated="default" // TODO add state/validation/errors to this and other FormGroups
+      >
+        <SimpleSelect
+          id="sourceProvider"
+          options={srcProvidersOptions}
+          value={[srcProvidersOptions.find((option) => option.value === srcProvider)]}
+          onChange={(selection) => setSrcProvider((selection as OptionWithValue<Provider>).value)}
+          placeholderText="Select a provider"
+        />
+      </FormGroup>
+
+      <FormGroup
+        label="Target provider"
+        isRequired
+        fieldId="tgtProvider"
+        helperTextInvalid="TODO"
+        validated="default" // TODO add state/validation/errors to this and other FormGroups
+      >
+        <SimpleSelect
+          id="tgtProvider"
+          options={tgtProvidersOptions}
+          value={[tgtProvidersOptions.find((option) => option.value === tgtProvider)]}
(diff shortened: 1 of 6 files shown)


mturley:
Since we're using squash and merge it might be easiest to rebase this manually by cherry-picking the new commits onto a fresh branch. That's the only thing I don't like about squash-and-merge! 🙃

gildub:
@mturley, cherry-pick is definitely the easiest in such case!

src/app/Plans/components/Wizard/GeneralForm.tsx
```diff
@@ -0,0 +1,92 @@
+import * as React from 'react';
+import { Form, FormGroup, TextArea, TextInput, Title } from '@patternfly/react-core';
+import SimpleSelect, { OptionWithValue } from '@app/common/components/SimpleSelect';
+import { Provider } from '@app/Providers/types';
+
+interface IGeneralFormComponent {
+  srcProviders: Provider[];
+  tgtProviders: Provider[];
```
ibolton336, line 8: We should change "tgt" to "dest". After discussing with Mike, we have come to the consensus that "src" and "dest" work best for code legibility going forward. 

NOTE: The markup for ui fields will continue to read "Source" and "Target"

gildub: @ibolton336, no worries I'll change it.

src/app/Plans/components/Wizard/GeneralForm.tsx
```diff
@@ -0,0 +1,92 @@
...
+}
+
+const GeneralForm: React.FunctionComponent<IGeneralFormComponent> = ({
+  sourceProviders,
+  targetProviders,
+}: IGeneralFormComponent) => {
+  const [planName, setPlanName] = React.useState<string>('');
+  const [planDesc, setPlanDesc] = React.useState<string>('');
+  const [srcProvider, setSrcProvider] = React.useState<Provider | null>(null);
+  const [tgtProvider, setTgtProvider] = React.useState<Provider | null>(null);
```
ibolton336, line 18: Still a few more tgt abbreviations here.

src/app/Plans/components/Wizard/GeneralForm.tsx
```diff
@@ -0,0 +1,92 @@
...
+  sourceProviders: Provider[];
+  targetProviders: Provider[];
+}
+
+const GeneralForm: React.FunctionComponent<IGeneralFormComponent> = ({
+  sourceProviders,
+  targetProviders,
+}: IGeneralFormComponent) => {
+  const [planName, setPlanName] = React.useState<string>('');
+  const [planDesc, setPlanDesc] = React.useState<string>('');
```
ibolton336, line 16: can we change 'Desc' to 'Description' for clarity? Unless it is an extremely long name, I think it would be best to have intentionally descriptive variable names. This way our code is self documenting.

src/app/Plans/components/Wizard/GeneralForm.tsx
mturley: Agreed, let's use `sourceProvider` and `targetProvider` for readability

src/app/Plans/components/Wizard/WizardComponent.tsx
```diff
@@ -0,0 +1,126 @@
+import * as React from 'react';
+import WizardStepContainer from './WizardStepContainer';
+import { Wizard } from '@patternfly/react-core';
+import Review from './Review';
+import GeneralForm from './GeneralForm';
+import { Provider } from '@app/Providers/types';
+
+interface IWizardComponent {
```
mturley, line 8: For consistency, let's put `Props` at the end of this interface

src/app/Plans/components/Wizard/WizardComponent.tsx
```diff
@@ -0,0 +1,126 @@
...
+  isOpen: boolean;
+  onHandleWizardClose: () => void;
+  sourceProviders: Provider[];
+  targetProviders: Provider[];
+}
+
+const WizardComponent: React.FunctionComponent<IWizardComponent> = ({
```
mturley, line 15: Maybe we should call this PlanWizard instead of WizardComponent, in case we eventually have another wizard? WizardComponent was only really necessary to distinguish from WizardContainer in mig-ui, which we don't need here. Totally not important, I'm fine either way.

src/app/Plans/components/Wizard/WizardComponent.tsx
```diff
@@ -0,0 +1,126 @@
...
+    {
+      id: stepId.Review,
+      name: 'Review',
+      component: <Review />,
+      nextButtonText: 'Finish',
+    },
+  ];
+
+  const handleClose = () => {
+    onHandleWizardClose();
```
mturley, line 106: The `onHandleWizardClose` naming is a bit redundant, I think we can just call the prop `onClose` and pass it directly into `<Wizard>` instead of having the `handleClose` function in the middle.

mturley (changes requested):
This is looking good! Just a couple of minor things.

src/app/Plans/components/Wizard/WizardComponent.tsx
gildub: Sorry, I know better.

src/app/Plans/components/Wizard/WizardComponent.tsx
gildub: That's a good idea, changing it.

src/app/Plans/components/Wizard/GeneralForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+const GeneralForm: React.FunctionComponent<IGeneralFormComponentProps> = ({
+  sourceProviders,
+  targetProviders,
+}: IGeneralFormComponentProps) => {
+  const [planName, setPlanName] = React.useState<string>('');
+  const [planDescription, setPlanDescription] = React.useState<string>('');
+  const [sourceProvider, setSourceProvider] = React.useState<Provider | null>(null);
+  const [targetProvider, setTargetProvider] = React.useState<Provider | null>(null);
+
+  const sourceProvidersOptions = Object.values(sourceProviders).map((provider) => ({
```
mturley, line 20: We can probably move these provider options / select stuff to a common `<ProviderSelect>` or something eventually, just making a note to myself for that. I have similar code in #24 / #25.

src/app/Plans/components/Wizard/PlanWizard.tsx
```diff
@@ -0,0 +1,122 @@
...
+}
+
+const PlanWizard: React.FunctionComponent<IPlanWizardProps> = ({
+  isOpen,
+  onClose,
+  sourceProviders,
+  targetProviders,
+}: IPlanWizardProps) => {
+  enum stepId {
+    General = 1,
```
mturley, line 22: Not a request, just a question: what's the deal with the `= 1` on General but none on the other steps? Looking in the debugger it seems that sets all of these to numeric values, which is pretty cool, I just wasn't familiar with this pattern.

mturley (approved):
LGTM! @ibolton336 do you have any thoughts on this before we merge it?

ibolton336 (approved):
LGTM

src/app/Plans/components/Wizard/PlanWizard.tsx
gildub: @mturley, I actually copied that one from mig-ui, interestingly this pattern is similar in Golang.

mturley: Cool! Good to know.

PR #20, opened 2020-09-01
Update plan and hook data structure

gildub:
Updates the interfaces to get closer to API data.

### DIFF
--- a/src/app/Plans/types.ts
+++ b/src/app/Plans/types.ts
@@ -1,4 +1,5 @@
 import { Hook } from '@app/Hooks/types';
+import { CR } from '@app/common/types';
 import { Mapping } from '@app/Mappings/types';
 import { IVMwareProvider, ICNVProvider, IHost } from '@app/Providers/types';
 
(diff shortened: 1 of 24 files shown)


src/app/Plans/types.ts
```diff
@@ -1,81 +1,48 @@
...
   isAddPlanDisabled: boolean;
   disabledText: string;
 }
 
-export interface IMigHooks {
-  before: boolean;
-  beforeHook: IHook;
-  after: boolean;
-  afterHook: IHook;
+export interface PlanVM {
```
mturley, line 11: Just for consistency and to distinguish between components and interfaces in imports, let's put an `I` prefix on all our interfaces.

Honestly I wonder if we should be putting `E` in front of enums and `T` or `U` in front of union types. I sometimes have to remember what is and isn't a component when I look at imports. I was guilty of that here with `Mapping`.

src/app/common/types.ts
```diff
@@ -0,0 +1,23 @@
+// TODO: lib-ui candidate
+export interface MetaTypeMeta {
+  apiVersion: string;
+  kind: string;
+}
```
mturley, line 5: Moving common TS types to lib-ui is a great idea! Hmm, I wonder if that would work since we compile to JS with Rollup. It exports `.d.ts` files too, maybe that would be sufficient.

src/app/Plans/types.ts
ibolton336: +1 I agree with the I in front of interfaces. Keeps things simple & allows for quick identification of TS specific pieces.

ibolton336:
@gildub Can you please provide a description for your PRs stating their intent & the motivation behind them. Also link any specific issues they aim to resolve. Thanks

mturley (commented):
If we put the `I` back in front of these interfaces I think this will be good to go.

src/app/Plans/types.ts
gildub: @mturley, @ibolton336, yeah I guess I got excited with the `Mapping` ;)
I'm fine with using `I` for interface so we're consistent.
For `E`and others, I'll tend to not using it unless it really help when reading the code. But if the TS trend is going that I won't be able to resist it.

src/app/common/types.ts
gildub: @mturley CRs are so omnipresent that any generalization could help. Not sure that's the best way but would help if we could go in that direction.

ibolton336:
LGTM - this will likely keep changing but going to go ahead and merge this version.

ibolton336 (approved):

PR #21, opened 2020-09-01
PlanWizard step 2: Initial Filter & Select VMs 

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/17


package.json
```diff
@@ -67,7 +67,7 @@
   },
   "dependencies": {
     "@konveyor/lib-ui": "^1.4.0",
-    "@patternfly/react-core": "^4.40.4",
+    "@patternfly/react-core": "4.47.1",
```
ibolton336, line 70: Please split the Patternfly update into a separate PR. Thanks

gildub: https://github.com/konveyor/virt-ui/pull/26

src/app/Plans/components/Wizard/FilterVMs.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+// TODO replace with redux or equiv.
+import { VMsOptions } from './mocks/VMs.mock';
+
+interface IFilterVMsProps {
+  Inventory: [];
+}
+
+const FilterVMs: React.FunctionComponent<IFilterVMsProps> = ({ Inventory }: IFilterVMsProps) => {
+  const [activeItems, setItems] = React.useState(VMsOptions);
```
ibolton336, line 12: @mturley What do you think about inferring types here from the VMsOptions vs defining an interface & conditionally accepting an undefined type? Would we always know the VMsOptions list of types or is this coming from the API where we would need to check for some error when fetching types?

src/app/Plans/components/Wizard/FilterVMs.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const [checkedItems, setCheckedItems] = React.useState([]);
+
+  const onSelect = (event, treeViewItem, parentItem) => {
+    return;
+  };
+
+  const onCheck = (event, treeViewItem) => {
```
ibolton336, line 19: Is this a potential use case for a reusable hook?

src/app/Plans/components/Wizard/FilterVMs.tsx
```diff
@@ -0,0 +1,55 @@
...
+        <TreeView
+          data={VMsOptions}
+          activeItems={activeItems}
+          hasChecks
+          onSearch={() => {
+            return;
+          }}
+          onSelect={onSelect}
+          onCheck={onCheck}
+          searchProps={{
```
ibolton336, line 44: We could perhaps pull this out of the render logic & define as a const.

src/app/Plans/components/Wizard/SelectVMs.tsx
```diff
@@ -0,0 +1,141 @@
...
+  wrappable,
+} from '@patternfly/react-table';
+import tableStyles from '@patternfly/react-styles/css/components/Table/table';
+
+import { IVMs } from '../../types';
+import { useSortState, usePaginationState, useSelectionState } from '@app/common/hooks';
+import { StatusIcon, StatusType } from '@konveyor/lib-ui';
+
+interface ISelectVMsProps {
+  VMs: IVMs[];
```
ibolton336, line 21: Is the the naming pattern we would like to settle on for VMs? This seems like a tricky one with it pertaining to a plural abbreviation. Anyone have any opinions about how to name stuff like this?

src/app/Plans/components/Wizard/SelectVMs.tsx
```diff
@@ -0,0 +1,141 @@
...
+    sortedItems
+  );
+
+  const { selectedItems: expandedVMs, toggleItemSelected: toggleVMExpanded } = useSelectionState<
+    IVMs
+  >(sortedItems);
+
+  const columns: ICell[] = [
+    {
+      // Using a custom column instead of Table's onSelect prop due to issues
```
ibolton336, line 43: What issues came up here?

ibolton336 (changes requested):
See comments. Thanks

mturley (commented):
@gildub, let me know if my feedback about the `VMsOptions` / `TreeViewDataItem` objects makes sense. Basically, I don't think we should be storing those in state at all. They're a render implementation detail. The only state we actually change is which items are selected, so let's just keep that in state.

Then, we can create an immutable set of TreeViewDataItems with `checked: false` on each item by traversing the API data, and then map over that to create a copy of that data with each of the `checked` booleans can be based on the `checkedItems` state array.

In order to represent the parent items being checked if the children are all checked, we can use helpers in that map to recursively look at the children when determining each `checked` value. See the `areAllDescendantsChecked` and `areSomeDescendantsChecked` helpers in this JSX example in the docs: https://www.patternfly.org/v4/documentation/react/components/treeview#with-checkboxes
, they use the `checkedItems` array to check ids of the whole subtree. See the `this.mapTree` method in those docs for an example of how to map over the tree and add the right `checked` values.

So we could do something like this, assuming we have `datacenters` API data in scope:
```ts
const baseOptions: TreeViewDataItem[] = [
  {
    name: 'All datacenters',
    id: 'all',
    checkProps: { 'aria-label': 'all-check', checked: false },
    children: datacenters.map((datacenter) => ({
      // Whatever we need here based on the API data structure
      name: datacenter.metadata.name,
      id: datacenter.id
      checkProps: { 'aria-label': `${datacenter.metadata.name}-check`, checked: false },
      children: ... // Another map... probably we'd refactor that into a helper we can use recursively
    }),
  },
]
```

And then possibly in the next PR when we add the checkbox state, we can do:

```ts
const options = baseOptions.map((option) => ({
  ...option,
  checkProps: { ...option.checkProps, checked: areAllDescendantsChecked(option) },
  children: ... // Again, map over the children recursively, so this all should probably be a helper function
});
```

And in fact, for `checkedItems` we could use the `useSelectionState` hook I wrote to manage adding and removing items to it instead of `useState` directly. https://konveyor.github.io/lib-ui/?path=/docs/hooks-useselectionstate--checkboxes

As you can see I'm coming up with this as I go, I just confused myself a little bit, let me know if you want me to try and hack this together myself as a followup PR. My point is that we should get rid of your `onCheck` for now, and we shouldn't keep these options in state. Let's at least make those changes in this PR and then handle the second `checked` map call in another PR.

src/app/Plans/components/Wizard/SelectVMs.tsx
mturley: I agree, the interface represents one VM, it should be called `IVM` so that `IVM[]` represents an array of VMs.

mturley: Also, and this is more of a nitpick, prop names shouldn't be capitalized since that is reserved for components and types. Let's call this `vms`.

src/app/Plans/components/Wizard/SelectVMs.tsx
mturley: That's duplicated from my comment in the providers table. Basically, there is a bug in Table where if you use `onSelect` together with the expandable row, it adds an empty column to the left of your expandable row because of its code for adding checkbox columns. This indents the expanded content by a bunch. I plan to open that as an issue in PF and we can eventually switch back to onSelect. We should do that because I'm pretty sure doing this breaks mobile friendliness.

src/app/Plans/components/Wizard/SelectVMs.tsx
```diff
@@ -0,0 +1,141 @@
...
+        vm.Name,
+        vm.Datacenter,
+        vm.Cluster,
+        vm.Host,
+        vm.FolderPath,
+      ],
+    });
+    if (isExpanded) {
+      rows.push({
+        parent: 1,
```
mturley, line 103: This `parent` value needs to be the index of the parent row, not `1`. See the one I use in the providers table, you can use `rows.length - 1`.

src/app/Plans/components/Wizard/SelectVMs.tsx
```diff
@@ -0,0 +1,141 @@
...
+        compoundExpand: 1,
```
mturley, line 104: Per @vconzola , we need to replace this compound expansion with regular expansion, which I think just means removing the `compoundExpand` property. See this example: https://www.patternfly.org/v4/documentation/react/components/table#compact-expandable (but obviously with our hooks approach instead of the class stuff)

src/app/Plans/components/Wizard/mocks/VMs.mock.ts
```diff
@@ -0,0 +1,146 @@
+import { IVMs } from '../../../types';
+
+export const VMsOptions = [
```
mturley, line 3: As mentioned above, instead of putting the tree view item structure in here, let's put mock data in here of datacenters and clusters and folders the way we expect them to be in the API. Here is an example you can look at in Jeff's demo API server: https://inventory-openshift-migration.apps.cluster-jortel.v2v.bos.redhat.com/namespaces/openshift-migration/providers/vsphere/test/tree/vm (requires VPN)

mturley: Actually let's talk on Slack about this, Jeff is telling me that might not make sense:

> The datacenters/clusters/folders are not the same tree.
> it's either datacenter/clusters (by folder)
> or
> datacenter/folders/vms

src/app/Plans/types.ts
```diff
@@ -46,3 +46,13 @@ export interface IMigration {
     StorageReady: boolean;
   };
 }
+
+export interface IVMs {
```
mturley, line 50: As mentioned above, let's name this interface `IVM` since it represents a singular VM. Multiple VMs are represented by `IVM[]`.

mturley:
Sorry for all the thinking-out-loud, guys. The more I think about this and discuss the data structure with Jeff, the more I think maybe we should leave this PR with the `VMsOptions` mock data in place in that `TreeViewDataItem` structure, and save the creation of it based on API data for another PR. @ibolton336 thoughts?

src/app/Plans/components/Wizard/FilterVMs.tsx
ibolton336: Good point - we need to make sure we are operating in a way that is react friendly.

ibolton336:
> Sorry for all the thinking-out-loud, guys. The more I think about this and discuss the data structure with Jeff, the more I think maybe we should leave this PR with the `VMsOptions` mock data in place in that `TreeViewDataItem` structure, and save the creation of it based on API data for another PR. @ibolton336 thoughts?

+1 I agree - lets open an issue for that & implement in a separate PR. But the rest of the comments should be easy enough to address I think.

src/app/Plans/components/Wizard/FilterVMs.tsx
ibolton336: +1 for suggestion to move work to implement the correct data structure to a new PR/Issue.

src/app/Plans/components/Wizard/SelectVMs.tsx
ibolton336: Got it, thanks!

vconzola:
@gildub Link to latest mockups for this step: [](https://docs.google.com/presentation/d/1xpOlK_OpCDAnylVEL7bNH_UIRsEx89N1wfZKHVQySbk/edit?usp=sharing) Slides 29-31. Let me know if you have any questions.

src/app/Plans/components/Wizard/SelectVMs.tsx
mturley: Sorry, disregard my comment here. That was just for the mappings tables.

src/app/Plans/components/Wizard/SelectVMs.tsx
```diff
@@ -0,0 +1,141 @@
...
+          }}
+          checked={areAllSelected}
+        />
+      ),
+      columnTransforms: [classNamesTransform(tableStyles.tableCheck)],
+    },
+    {
+      title: 'Migration Analysis',
+      transforms: [sortable, wrappable],
+      cellTransforms: [compoundExpand],
```
mturley, line 59: We'll also need to remove `compoundExpand` here and remove the import.

vconzola:
@mturley One thing I show in the mockups that I'd like to include if it's not too much work is the ability to expand all of the rows with a single click. Possible? What do you think?

mturley:
@vconzola Yeah, that's easy with my state abstraction :) it's the same code that drives our checkboxes, so it already has a "select all" function. Where should the button go?

vconzola:
@mturley Nice! I'd like it to be as shown in the mockups. Caret icon (https://fontawesome.com/icons/caret-square-right?style=regular) in the table toolbar that changes to caret down when it's selected and expands all of the rows.

mturley:
Ah, gotcha. Sorry I missed that.
@gildub, if you add a button for that, you can add `selectAll` to your destructured object from `useSelectionState` and call it with a boolean for whether to expand all or collapse all.

mturley:
Ah, dammit, sorry. I was thinking this was another PR.. the tree state is a bit different. Let me take a closer look tomorrow

src/app/Plans/components/Wizard/SelectVMs.tsx
gildub: @mturley are going for a Collapsible Table?

@vconzola, Initially I tried an Expandable Table but that didn't fit the purpose so I thought about a Compound Expandable Table.

src/app/Plans/types.ts
gildub: This one is obsolete, superseded by  https://github.com/konveyor/virt-ui/pull/20/files

src/app/Plans/components/Wizard/SelectVMs.tsx
gildub: @ibolton336, @mturley, IVMs was a bad choice.

src/app/Plans/components/Wizard/SelectVMs.tsx
gildub: I Initially used an Expandable Table but that didn't work as I explained during scrum last week.
Effectively the (Compact) Expandable format the second row with the same number of columns which limits the width of the text we have to display.
So I thought about a Compound Expandable Table was better look & feel.

mturley: Let's talk later so you can show me what you mean. You should be able to just use a colSpan to get around that.

mturley:
Ok @gildub apparently I was very confused yesterday. Vince was in fact referring to the table, so my prior advice stands. You can add `selectAll` to your destructuring here: https://github.com/konveyor/virt-ui/pull/21/files#diff-b61187f9e8bf2d5d8a64f3f8f6dd97fbR38 and then call it in your expand/collapse-all button. You'll also want to change the icon between `<AngleRightIcon />` and `<AngleDownIcon />` based on `areAllSelected`, so add that to the destructure too (or don't destructure it like we discussed on slack).

As for how to render the expand-all button in the header of the expand-caret column that is generated by Table... hm. Dammit, maybe that needs a PF change as well. @ibolton336 what do you say we handle that as a separate issue as well?

ibolton336:
> As for how to render the expand-all button in the header of the expand-caret column that is generated by Table... hm. Dammit, maybe that needs a PF change as well. @ibolton336 what do you say we handle that as a separate issue as well?

I am on board with that @mturley

src/app/Plans/components/Wizard/FilterVMs.tsx
```diff
@@ -0,0 +1,55 @@
...
+// TODO replace with redux or equiv.
+import { VMsOptions } from './mocks/VMs.mock';
+
+interface IFilterVMsProps {
+  Inventory: [];
+}
+
+const FilterVMs: React.FunctionComponent<IFilterVMsProps> = ({ Inventory }: IFilterVMsProps) => {
+  const [activeItems, setItems] = React.useState(VMsOptions);
+  const [checkedItems, setCheckedItems] = React.useState([]);
```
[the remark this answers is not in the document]
ibolton336: @gildub Can we add a comment and an issue link for this work? Thx.

mturley:
Ok. I opened #35, #36 and #38 to follow up on later.

src/app/Plans/components/Wizard/FilterVMs.tsx
mturley: @gildub let's make this `onCheck` function a no-op for now, I don't want to leave the imperative mutation here just in case we forget about it.

src/app/Plans/components/Wizard/FilterVMs.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { Form, FormGroup, TreeView, Title } from '@patternfly/react-core';
+
+// TODO replace with redux or equiv.
+import { VMsOptions } from './mocks/VMs.mock';
+
+interface IFilterVMsProps {
+  Inventory: [];
+}
+
+const FilterVMs: React.FunctionComponent<IFilterVMsProps> = ({ Inventory }: IFilterVMsProps) => {
```
ibolton336, line 11: Update to FilterVMsForm to keep consistent with GeneralForm

mturley:
Since the search state is also not connected, I think we can address that as a separate issue too so we don't hold this up any longer. Opened #39 for that. We probably want to address #38 and #39 at the same time since they will involve the same traversal of the generated tree items.

src/app/Plans/components/Wizard/mocks/VMs.mock.ts
```diff
@@ -0,0 +1,146 @@
...
+    defaultExpanded: true,
+  },
+];
+
+export const MOCK_VMS: IVM[] = [
+  {
+    Name: 'vm1',
+    Datacenter: 'Prod center 1',
+    Cluster: 'Cluster 1',
+    Host: 'host test1',
```
ibolton336, line 132: Just curious as to why these properties are capitalized? Not a big deal since they are going to be replaced with real data soon.

src/app/Plans/components/Wizard/SelectVMs.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { StatusIcon, StatusType } from '@konveyor/lib-ui';
+
+interface ISelectVMsProps {
+  vms: IVM[];
+}
+
+const SelectVMs: React.FunctionComponent<ISelectVMsProps> = ({ vms }: ISelectVMsProps) => {
+  const getSortValues = (vm: IVM) => {
+    return ['', vm.Name, vm.Datacenter, vm.Cluster, vm.Host, vm.FolderPath, ''];
+  };
```
mturley, line 28:
```suggestion
  const getSortValues = (vm: IVM) => {
    return [
      '', // Expand control column
      '', // Checkbox column
      vm.MigrationAnalysis,
      vm.Name,
      vm.Datacenter,
      vm.Cluster,
      vm.Host,
      vm.FolderPath,
      '', // Actions column
    ];
  };
```

src/app/Plans/components/Wizard/FilterVMs.tsx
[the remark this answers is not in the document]
mturley: I opened #38 for it, but we may want to add a comment there too.

mturley:
FYI @gildub now that https://github.com/konveyor/virt-ui/pull/32 is merged, when you rebase this you'll need to switch your `useSelectionState` imports to use the lib-ui version, and in the process change the params you pass to it into an object. We should also use the new `isEqual` and `isItemSelected` parts of it as described here: https://github.com/konveyor/virt-ui/pull/32/files#r485879159

src/app/Plans/components/Wizard/mocks/VMs.mock.ts
gildub: Good question, maybe old habit from different language.

src/app/Plans/components/Wizard/SelectVMs.tsx
gildub: Right.

src/app/Plans/components/Wizard/FilterVMsForm.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const [activeItems, setItems] = React.useState(VMsOptions);
+  const [checkedItems, setCheckedItems] = React.useState([]);
+
+  const onSelect = (event, treeViewItem, parentItem) => {
+    return;
+  };
+
+  const onCheck = (event, treeViewItem) => {
+    const checked = event.target.checked;
+    treeViewItem.checkProps.checked = checked;
```
mturley, line 21: @gildub , sorry if my earlier comment got lost in the mess, can you remove this `checked` assignment here and just have `onCheck` be a no-op `return;` like `onSelect` for now? Mutating state in-place like this is causing weird side-effects when trying to use the checkboxes, we should just have them totally do nothing until we implement #38.

mturley (changes requested):
One little thing

src/app/Plans/components/Wizard/FilterVMsForm.tsx
gildub: @mturley, no worries, done.

mturley (approved):
LGTM for now, @ibolton336 I think we can address any other nitpicks as we move forward with the followup issues

PR #23, opened 2020-09-02
Add network/storage mappings table component

### LINKED ISSUE #31: Replace compound expansion with simple expansion for mappings tables
Since we separated out network mappings from storage mappings we don't need the "xxx mappings" count column in the tables. User can see mapping details by expanding the whole table row. Updated mockups here: (https://docs.google.com/presentation/d/1xpOlK_OpCDAnylVEL7bNH_UIRsEx89N1wfZKHVQySbk/edit?usp=sharing)  Slides 22 and 25.

ibolton336:
- Adds necessary logic to populate the shared mappings table with network or storage mappings. 
- Adds Add mapping button to the top of the shared table. This button opens the shared mappings modal
- Uses mock data in local storage for testing. There are a few throwaway functions added to /mocks/helpers for this purpose. We can use these until we are ready for redux implementation. 

Closes #31

### DIFF
--- a/src/app/Mappings/Network/NetworkMappingsPage.tsx
+++ b/src/app/Mappings/Network/NetworkMappingsPage.tsx
@@ -16,19 +16,20 @@ import { PlusCircleIcon } from '@patternfly/react-icons';
 import { INetworkMapping, MappingType } from '../types';
 import MappingsTable from '../components/MappingsTable';
 import AddEditMappingModal from '../components/AddEditMappingModal';
+import { fetchMockStorage } from '../mocks/helpers';
 
 // TODO replace these with real state e.g. from redux
 const isFetchingInitialNetworkMappings = false; // Fetching for the first time, not polling
 // const networkMappings: INetworkMapping[] = [];
 
 const NetworkMappingsPage: React.FunctionComponent = () => {
-  const mockMapObj = localStorage.getItem('mappingsObject');
   const [networkMappings, setNetworkMappings] = React.useState([]);
 
+  //TODO replace with real state from redux
+  const mockMapObj = localStorage.getItem('networkMappingsObject');
   React.useEffect(() => {
     console.log(`TODO: fetch network mapping items`);
-    const mappingsKey = localStorage.getItem('mappingsObject' || '{}');
-    const currentMappings = mappingsKey !== null ? JSON.parse(mappingsKey).mappings : [];
+    const currentMappings = fetchMockStorage(MappingType.Network);
     setNetworkMappings(currentMappings || []);
   }, [mockMapObj]);
 
--- a/src/app/Mappings/Network/NetworkMappingsPage.tsx
+++ b/src/app/Mappings/Network/NetworkMappingsPage.tsx
@@ -18,15 +18,13 @@ import MappingsTable from '../components/MappingsTable';
 import AddEditMappingModal from '../components/AddEditMappingModal';
 import { fetchMockStorage } from '../mocks/helpers';
 
-// TODO replace these with real state e.g. from redux
 const isFetchingInitialNetworkMappings = false; // Fetching for the first time, not polling
-// const networkMappings: INetworkMapping[] = [];
 
 const NetworkMappingsPage: React.FunctionComponent = () => {
-  //TODO replace with real state from redux
-  const [networkMappings, setNetworkMappings] = React.useState([]);
+  //TODO: replace with real state from redux
+  const [networkMappings, setNetworkMappings] = React.useState<INetworkMapping[]>([]);
 
-  //TODO replace with real state from redux
+  //TODO: replace with real state from redux
   const mockMapObj = localStorage.getItem('networkMappingsObject');
   React.useEffect(() => {
     console.log(`TODO: fetch network mapping items`);
--- a/src/app/Mappings/Network/NetworkMappingsPage.tsx
+++ b/src/app/Mappings/Network/NetworkMappingsPage.tsx
@@ -57,7 +57,7 @@ const NetworkMappingsPage: React.FunctionComponent = () => {
                   </Button>
                 </EmptyState>
               ) : (
-                <MappingsTable mappings={networkMappings} />
+                <MappingsTable mappings={networkMappings} mappingType={MappingType.Network} />
               )}
             </CardBody>
           </Card>
--- a/src/app/Mappings/Network/NetworkMappingsPage.tsx
+++ b/src/app/Mappings/Network/NetworkMappingsPage.tsx
@@ -19,9 +19,19 @@ import AddEditMappingModal from '../components/AddEditMappingModal';
 
 // TODO replace these with real state e.g. from redux
 const isFetchingInitialNetworkMappings = false; // Fetching for the first time, not polling
-const networkMappings: INetworkMapping[] = [];
+// const networkMappings: INetworkMapping[] = [];
 
 const NetworkMappingsPage: React.FunctionComponent = () => {
+  const mockMapObj = localStorage.getItem('mappingsObject');
+  const [networkMappings, setNetworkMappings] = React.useState([]);
+
+  React.useEffect(() => {
+    console.log(`TODO: fetch network mapping items`);
+    const mappingsKey = localStorage.getItem('mappingsObject' || '{}');
+    const currentMappings = mappingsKey !== null ? JSON.parse(mappingsKey).mappings : [];
+    setNetworkMappings(currentMappings || []);
+  }, [mockMapObj]);
+
   const [isAddEditModalOpen, toggleAddEditModal] = React.useReducer((isOpen) => !isOpen, false);
   return (
     <>
--- a/src/app/Mappings/Network/NetworkMappingsPage.tsx
+++ b/src/app/Mappings/Network/NetworkMappingsPage.tsx
@@ -23,6 +23,7 @@ const isFetchingInitialNetworkMappings = false; // Fetching for the first time, 
 // const networkMappings: INetworkMapping[] = [];
 
 const NetworkMappingsPage: React.FunctionComponent = () => {
+  //TODO replace with real state from redux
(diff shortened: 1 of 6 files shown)


src/app/Mappings/mocks/helpers.ts
```diff
@@ -0,0 +1,32 @@
+import { MappingType } from '../types';
+
+export const updateMockStorage = (params: any) => {
```
mturley, line 3: We might want to add a comment here that this is temporary, so we don't invest too much time making it better / spreading it.

src/app/Mappings/Network/NetworkMappingsPage.tsx
```diff
@@ -16,12 +16,23 @@ import { PlusCircleIcon } from '@patternfly/react-icons';
...
 import AddEditMappingModal from '../components/AddEditMappingModal';
+import { fetchMockStorage } from '../mocks/helpers';
 
 // TODO replace these with real state e.g. from redux
 const isFetchingInitialNetworkMappings = false; // Fetching for the first time, not polling
-const networkMappings: INetworkMapping[] = [];
+// const networkMappings: INetworkMapping[] = [];
 
 const NetworkMappingsPage: React.FunctionComponent = () => {
+  const [networkMappings, setNetworkMappings] = React.useState([]);
```
mturley, line 26: Let's throw a comment here too saying this is temporary, so we don't end up duplicating redux state here in react land

src/app/Mappings/Network/NetworkMappingsPage.tsx
```diff
@@ -16,12 +16,23 @@ import { PlusCircleIcon } from '@patternfly/react-icons';
...
+
+  //TODO replace with real state from redux
+  const mockMapObj = localStorage.getItem('networkMappingsObject');
+  React.useEffect(() => {
```
mturley, line 30: So I'm not sure I get the useEffect code here, why does it have `mockMapObj` in its dependency array? That won't change here I don't think, so the useEffect will only be triggered once. I think we can get rid of the `mockMapObj` and have `[]` as the deps array for the effect?

src/app/Mappings/Storage/StorageMappingsPage.tsx
```diff
@@ -16,13 +16,24 @@ import { IStorageMapping, MappingType } from '../types';
...
 import AddEditMappingModal from '../components/AddEditMappingModal';
+import { fetchMockStorage } from '../mocks/helpers';
 
 // TODO replace these with real state e.g. from redux
 const isFetchingInitialStorageMappings = false; // Fetching for the first time, not polling
-const storageMappings: IStorageMapping[] = [];
+// const storageMappings: IStorageMapping[] = [];
 
 const StorageMappingsPage: React.FunctionComponent = () => {
+  const [storageMappings, setStorageMappings] = React.useState([]);
```
mturley, line 26: Same comments here: comment about temporary state, remove mockMapObj (I think?)

src/app/Mappings/components/AddEditMappingModal.tsx
```diff
@@ -88,7 +89,15 @@ const AddEditMappingModal: React.FunctionComponent<IAddEditMappingModalProps> =
       isOpen
       onClose={onClose}
       actions={[
-        <Button key="confirm" variant="primary" onClick={() => alert('TODO')}>
+        <Button
+          key="confirm"
+          variant="primary"
+          onClick={() => {
+            //TODO: Replace with a real redux call for adding a mapping
+            updateMockStorage({ mappingName, sourceProvider, targetProvider, mappingType });
```
mturley, line 97: We'll need to change this after we merge #25, we can generate the whole mapping object with a helper function I wrote there

src/app/Mappings/components/MappingsTable.tsx
```diff
@@ -1,13 +1,119 @@
...
+        },
+        {
+          title: <MappingsActionsDropdown />,
+        },
+      ],
+    });
+    if (isExpanded) {
+      rows.push({
+        parent: rows.length - 1,
+        compoundExpand: 4,
```
mturley, line 102: @vconzola was saying we should use regular expandable tables instead of compound expandable here. I think the only difference is getting rid of this `compoundExpand` property?

src/app/Mappings/components/MappingsTable.tsx
```diff
@@ -1,13 +1,119 @@
...
+                {/* <MappingDetailsTable {...props} /> */}
+              </div>
+            ),
+            props: { colSpan: columns.length, className: tableStyles.modifiers.noPadding },
+          },
+        ],
+      });
+    }
+  });
   // TODO: the storage and network mappings tables seem similar enough that we
```
mturley, line 117: We can get rid of my TODO comment here I think, now that we have some code for this.

src/app/Mappings/components/MappingsTable.tsx
```diff
@@ -31,7 +137,45 @@ const MappingsTable: React.FunctionComponent<IMappingsTableProps> = ({
...
-  return <h1>TODO: table here</h1>;
+  return (
+    <>
+      <Level>
+        <LevelItem>
+          <Button
+            key="confirm"
+            variant="primary"
+            onClick={() => {
+              //TODO: Replace with a real redux call for adding a mapping
```
mturley, line 148: This is just for opening the modal, so we can remove this comment

src/app/Mappings/components/MappingsTable.tsx
```diff
@@ -1,13 +1,119 @@
...
+  mappingType: string;
+  toggleAddEditModal: () => void;
 }
 
 const MappingsTable: React.FunctionComponent<IMappingsTableProps> = ({
   mappings,
+  mappingType,
+  toggleAddEditModal,
 }: IMappingsTableProps) => {
+  const getSortValues = (mapping: ICommonMapping) => {
```
mturley, line 42: I think we should use `Mapping` instead of `ICommonMapping` here. `Mapping` is a union of the two specific mapping interfaces, so a mapping with network-specific or storage-specific properties is a valid `Mapping` but not a valid `ICommonMapping` (which is just what those two interfaces extend).

src/app/Mappings/components/MappingsTable.tsx
```diff
@@ -1,13 +1,119 @@
...
+  };
+
+  const { sortBy, onSort, sortedItems } = useSortState(mappings, getSortValues);
+  const { currentPageItems, setPageNumber, paginationProps } = usePaginationState(sortedItems, 10);
+  React.useEffect(() => setPageNumber(1), [sortBy, setPageNumber]);
+
+  const {
+    selectedItems: expandedMappings,
+    toggleItemSelected: toggleMappingExpanded,
+  } = useSelectionState<ICommonMapping>(sortedItems);
```
mturley, line 54: Same here, `Mapping` instead of `ICommonMapping`. Maybe I shouldn't have even exported it.

src/app/Mappings/components/MappingsTable.tsx
```diff
@@ -1,13 +1,119 @@
...
+      transforms: [sortable],
+      cellTransforms: [compoundExpand],
+    },
+    { title: '', columnTransforms: [classNamesTransform(tableStyles.tableAction)] },
+  ];
+
+  const rows: IRow[] = [];
+  currentPageItems.forEach((mapping: ICommonMapping) => {
+    const { name, sourceProvider, targetProvider, items } = mapping;
+    const isExpanded = expandedMappings.includes(mapping);
```
mturley, line 71: This will be fine for now, but I'm actually realizing that if the mappings objects in redux are updated, this referential equality will fail to match and we'll lose our expanded state. We might need to use the `isItemSelected` function exported by the useSelectionState hook, and pass an `isEqual` function into it that uses `id`s or something. I do that in my examples here: https://konveyor.github.io/lib-ui/?path=/docs/hooks-useselectionstate--checkboxes 

I probably need to make the same change elsewhere, that'll become apparent as we start to use redux for these

src/app/Mappings/components/MappingsTable.tsx
```diff
@@ -1,13 +1,119 @@
...
+  } = useSelectionState<ICommonMapping>(sortedItems);
+
+  const columns: ICell[] = [
+    { title: 'Name', transforms: [sortable] },
+    { title: 'Source provider', transforms: [sortable] },
+    { title: 'Target provider', transforms: [sortable] },
+    {
+      title: <>{mappingType === MappingType.Network ? 'Network mappings' : 'Storage mappings'}</>,
+      transforms: [sortable],
+      cellTransforms: [compoundExpand],
```
mturley, line 63: This is another place where we need to remove `compoundExpand`, this and the import.

src/app/Mappings/types.ts
```diff
@@ -28,7 +28,12 @@ export interface IStorageMappingItem {
 export interface ICommonMapping {
   type: MappingType;
   name: string;
-  provider: {
+  sourceProvider: {
```
mturley, line 31: This might conflict with #25, I think I made the same change there.

mturley (commented):
Looking good! Just a couple things.

src/app/Mappings/Network/NetworkMappingsPage.tsx
ibolton336: The mockMapObj is required in the dependency array because when I added an item to the local storage, there was no rerender triggered. This will for sure go away when we move to real redux state but it is needed temporarily for the mock localstorage code to work for now.

src/app/Mappings/components/MappingsTable.tsx
ibolton336: Added a comment for now. We should open a PR to pull in the useSelection state from lib-ui also.

src/app/Mappings/Network/NetworkMappingsPage.tsx
mturley: got it. ok.

mturley (approved):
LGTM 👍

PR #24, opened 2020-09-02
Prototype network/storage mapping builder (multi-select source variant)

mturley:
Closes #7.

This version uses a typeahead multi-select for the sources, and groups multiple sources to single targets in each arrow-box group.

Only one of #24 or #25 should be merged, the other will be closed after we review the two designs.

Screen Shot 2020-09-03 at 1 23 31 PM

PR #25, opened 2020-09-03
Prototype network/storage mapping builder (single-select source variant)

### LINKED ISSUE #7: Build prototype of mapping creation widget
Mockups: https://docs.google.com/presentation/d/1YtiJKf5FyDhBf_eIc6pXbHloCSidDmBKATgYXRtPh-o/edit#slide=id.g8b7ac84433_0_57

Originally this was designed as a drag-and-drop tool, I want to try something simpler before we get too far with that plan.

mturley:
Closes #7.

Based on #24, this converts the mapping arrow-box groups so that each maps only a single source to a single target, and the same target can be selected in more than one group. This addresses some feedback that @vconzola had about the multi-select source control being confusing. This also simplifies the state management a bit, since it is closer to the actual API data structure.

Only one of #24 or #25 should be merged, the other will be closed after we review the two designs.

Screen Shot 2020-09-03 at 1 20 45 PM


src/app/Mappings/components/MappingBuilder/MappingBuilder.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  return (
+    <>
+      <TextContent>
+        <Text component="p">{instructionText}</Text>
+      </TextContent>
+      {builderItems.map((item, itemIndex) => {
+        let key = '';
+        if (mappingType === MappingType.Network) {
+          const t = item.source as IVMwareNetwork | null;
```
ibolton336, line 71: :+1: Very cool use of typescript type assertion.

mturley: Honestly though I can't remember why I used the variable name `t`. I think it came from using it on the target side, but this one should be `s` or I should use longer names. I'll add a commit.

src/app/Mappings/components/MappingBuilder/MappingSourceSelect.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <SimpleSelect
+      id={id}
+      className="mapping-item-select"
+      variant="typeahead"
+      isPlain
+      options={options}
+      value={[selectedOption]}
+      onChange={(selection) => {
+        setSource((selection as OptionWithValue<MappingSource>).value);
```
ibolton336, line 50: :+1: Looking good

src/app/Mappings/components/MappingBuilder/MappingTargetSelect.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const targetOptions: OptionWithValue<MappingTarget>[] = availableTargets.map((target) => ({
+    value: target,
+    toString: () => {
+      if (mappingType === MappingType.Network) {
+        return (target as ICNVNetwork).name;
+      }
+      if (mappingType === MappingType.Storage) {
+        return target as string;
+      }
+      return '';
```
ibolton336, line 41: I wonder if there is a cleaner way to code this? I have used a similar pattern in the past but tried recently to switch to ternary expressions for these. It would probably be overkill for a switch statement. This works fine but I am always wishing there was some sort of middle ground.

src/app/Mappings/components/MappingBuilder/MappingBuilder.tsx
mturley: Actually, turns out I didn't need this at all. Both types of sources have an `id` property. I'll remove the whole condition

src/app/Mappings/components/MappingBuilder/MappingTargetSelect.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <SimpleSelect
+      id={id}
+      className="mapping-item-select"
+      variant="typeahead"
+      isPlain
+      options={targetOptions}
+      value={[selectedOption]}
+      onChange={(selection) => {
+        setTarget((selection as OptionWithValue<MappingTarget>).value);
```
ibolton336, line 57: :+1:

src/app/Mappings/components/MappingBuilder/MappingTargetSelect.tsx
mturley: Yeah, I was thinking of a ternary expression, but that requires choosing either network or storage as the condition and then having "other" be assumed to be the other. i.e. if it's a network mapping do this, else do this. I don't know if we'll ever need to add a third kind of mapping, but this allows for that. Could always refactor that later, so if you think it should be a ternary I'm fine with that.

src/app/Mappings/components/MappingBuilder/helpers.ts
```diff
@@ -0,0 +1,71 @@
...
+  INetworkMappingItem,
+  IStorageMappingItem,
+} from '../../types';
+import { IMappingBuilderItem } from './MappingBuilder';
+import { IVMwareProvider, ICNVProvider } from '@app/Providers/types';
+
+export const getSourceById = (sources: MappingSource[], id: string): MappingSource | undefined =>
+  sources.find((source) => source.id === id);
+
+export const getBuilderItemsFromMapping = (
```
ibolton336, line 16: Is this ever called?

src/app/Mappings/components/MappingBuilder/MappingTargetSelect.tsx
mturley: If I had proper generics support in SimpleSelect this wouldn't even be necessary 😄

src/app/Mappings/components/MappingBuilder/helpers.ts
```diff
@@ -0,0 +1,71 @@
...
+  MappingItem,
+  MappingType,
+  MappingTarget,
+  INetworkMappingItem,
+  IStorageMappingItem,
+} from '../../types';
+import { IMappingBuilderItem } from './MappingBuilder';
+import { IVMwareProvider, ICNVProvider } from '@app/Providers/types';
+
+export const getSourceById = (sources: MappingSource[], id: string): MappingSource | undefined =>
```
ibolton336, line 13: Is this ever called?

ibolton336 (changes requested):

src/app/Mappings/components/MappingBuilder/MappingTargetSelect.tsx
ibolton336: I am fine with if statements if there are more than 3 statements. & maybe we switch to switch statements if there are more than 3. Its more of a gripe with JS than with your PR :)

src/app/Mappings/components/MappingBuilder/helpers.ts
mturley: Not yet, but it can be used to pre-populate the state here from API mappings when we add edit functionality.

src/app/Mappings/components/MappingBuilder/helpers.ts
mturley: It's called in `getBuilderItemsFromMapping`.

ibolton336 (approved):
LGTM :100:

mturley:
Awesome! Thanks @ibolton336

PR #26, opened 2020-09-03
Bump PF

gildub:
Bumps `@patternfly/react-core` to 4.47.0 needed for Treeview.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
   },
   "dependencies": {
     "@konveyor/lib-ui": "^1.4.0",
-    "@patternfly/react-core": "^4.40.4",
+    "@patternfly/react-core": "^4.47.0",
     "@patternfly/react-icons": "^4.7.2",
     "@patternfly/react-styles": "^4.7.2",
     "@patternfly/react-table": "^4.15.5",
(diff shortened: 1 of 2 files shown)


ibolton336:
LGTM

package.json
```diff
@@ -67,7 +67,7 @@
   },
   "dependencies": {
     "@konveyor/lib-ui": "^1.4.0",
-    "@patternfly/react-core": "^4.40.4",
+    "@patternfly/react-core": "4.47.1",
```
ibolton336, line 70: Should be "@patternfly/react-core": "^4.47.0" 
Looks like 4.47.0 is the latest version https://www.npmjs.com/package/@patternfly/react-core.
Also, we need to include the caret as we do for the other versions. This will allow for small version bumps & keep things consistent w respect to semantic versioning. Thx

ibolton336 (changes requested):

package.json
gildub: 4.47.0 is latest published version on npm and it works too with Treeview.
Adding the caret too. 
Thanks

ibolton336 (approved):

PR #27, opened 2020-09-04
Initial redux scaffolding

ibolton336:
This PR is intended to start the discussion around what patterns we want to settle on for redux. Questions that need answering:

- What middleware do we plan on using? (redux-saga, redux-logger, connected-router, redux-persist, etc)
- What directory structure would we feel comfortable moving forward with?
[directory structure suggestions](https://redux.js.org/faq/code-structure#what-should-my-file-structure-look-like-how-should-i-group-my-action-creators-and-reducers-in-my-project-where-should-my-selectors-go)
- Where do we plan on centralizing types for redux? Where would we like to keep our types.ts files & how to we remain consistent with this?
- Do we plan on using union types to infer the type of redux actions? examples in this PR are pulled from
[typescript redux example](https://redux.js.org/recipes/usage-with-typescript )


gildub:
* What middleware do we plan on using?
  - `redux-saga`: sagas or async/await? Sagas brings on-the-fly cancel feature.
  - `redux-logger`: Is it mainly used for developments only? If yes then arent't devtools (redux-toolkit) enough?
  - `connected-router`: I suppose it's `connected-react-router`?
  - redux-persist: What are the use cases ? Can't they be addressed using local storage (JS)?
  - etc...
     - `reduxsauce`: If using `redux-toolkit` then won't be needed.

src/app/redux/common/actions.ts
```diff
@@ -0,0 +1,22 @@
+import { createAction } from '@reduxjs/toolkit';
+import { IMessage, SEND_MESSAGE, DELETE_MESSAGE, IChatActionTypes } from './types';
+
+// TypeScript infers that this function is returning SendMessageAction
```
mturley, line 4: I guess this is fine, maybe smart? Definitely weird. Whatever the redux team recommends, I think is probably good.

src/app/redux/common/index.ts
```diff
@@ -0,0 +1,6 @@
+import commonReducer from './reducers';
+export default commonReducer;
+
+// TODO: Discuss how would we like our directory structure to look for Redux? The redux docs have a 3 recommended approaches:
```
mturley, line 4: I think if we plan to try the `createSlice` approach provided by redux-toolkit, that lends itself to the "ducks" approach. The question is where do our ducks go. Do we want to have all the redux stuff in one place, and have e.g. `src/app/redux/plans/` and `src/app/redux/mappings/` etc, each containing maybe `reducer.ts` and `types.ts`? Maybe we just have `src/app/redux/plans.ts` with the slice and types and everything in one place?

Or maybe we pair the redux code with the relevant react code? We already have e.g. `src/app/Plans/` and `src/app/Mappings/`, which already each contain a `types.ts` file. We could just add `src/app/Plans/reducer.ts` or `src/app/Plans/redux/*` or whatever. The only issue with that is if we end up having more than one "duck" related to each page, or more than one page relating to the same "duck". I guess they could import across those boundaries.

I'm split between these two, leaning towards the `src/app/redux/` folder with each slice/resource-type/duck or whatever in there. I do think we should probably move a lot of our `types.ts` stuff into that area, at least the types that describe API objects. Maybe the types that are specific to UI components can stay near them.

src/app/redux/common/types.ts
```diff
@@ -0,0 +1,27 @@
...
+
+interface IDeleteMessageAction {
+  type: typeof DELETE_MESSAGE;
+  meta: {
+    timestamp: number;
+  };
+}
+
+//TODO: Discuss how do we feel about union types/inferring types?
+export type IChatActionTypes = ISendMessageAction | IDeleteMessageAction;
```
mturley, line 27: This is probably ok? Not sure until I play with it a little. I wonder if `createSlice` would take care of this for us though.

mturley (commented):
Here are my opinions. This looks good so far, I say let's play with [`createSlice`](https://redux-toolkit.js.org/api/createSlice).

src/app/redux/common/index.ts
ibolton336: I think I like keeping the redux stuff separate & following the ```src/app/redux/```. I don't think it really did us any favors trying to cram the redux stuff into the feature component directories on MTC. Having somewhere to store the api types globally would be a plus to that approach also.

ibolton336:
For several reasons, we have decided to use [react-query](https://react-query.tanstack.com/docs/videos) instead of react-redux. Closing this PR as we start to ramp up with react-query.

PR #28, opened 2020-09-04
PlanWizard: Initial Storage (step 3) and Network mapping (step 4)

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/18
Resolves https://github.com/konveyor/virt-ui/issues/29


src/app/Plans/components/Wizard/StorageMapping.tsx
```diff
@@ -0,0 +1,123 @@
...
+      </TextContent>
+      <Radio
+        id="existing-mapping"
+        label="Use an existing storage mapping"
+        name="radio-existing"
+        onChange={onChange}
+        isChecked={isUseExistingMappingChecked}
+      />
+      {isUseExistingMappingChecked && (
+        <FormGroup
```
mturley, line 68: If you wrap this whole thing in a `<Form>` the CSS will add some spacing between your `<FormGroup>s`

gildub: @mturley, thanks.

gildub:
@mturley, `StorageMappingForm` is now calling `AddEditMappingModal`.
I believe we need to break the latter so we could reuse some part of it.

mturley:
@gildub I may be wrong, but since we don't want the mapping builder to appear in a modal-in-a-modal and we don't need the other parts of its form (source and target provider fields before the builder), we should just import and use `MappingBuilder` instead of `AddEditMappingModal`. It sort of stands alone as its own big form field. Then we can add a separate name field below when the checkbox to save it is checked, since the layout of that is different.

vconzola:
@mturley @gildub I think I found a bug in the Mapping Builder when playing around with adding a new mapping in the wizard. Once the list of Source datastores is exhausted the Add button should be disabled. In the example in the preview there are only two source datastores. If you add both of them to a mapping and then click Add again you get an empty dropdown. No bueno.

vconzola:
@gildub I just noticed something else when adding a mapping. If the user selects the "Save mapping to use again" checkbox the text field that's displayed needs a label as shown in the [mockups] (https://docs.google.com/presentation/d/1xpOlK_OpCDAnylVEL7bNH_UIRsEx89N1wfZKHVQySbk/edit?usp=sharing), slide 34.

@mturley  I do LOVE this preview feature.

gildub:
@vconzola, @mturley, the preview is a killer feature!!!
I don't understand how 2 component's instances could share a variable (saved mapping name).

gildub:
@vconzola, I added the label to the mapping name although it's above because of the grid but I can `flex`. I think it actually looks better because of the alignment with the mapping boxes above. But that's your call.

mturley:
@vconzola 

> Once the list of Source datastores is exhausted the Add button should be disabled.

Ah, good catch! I'll open an issue for followup on that, I think it's out of the scope of this PR since it's a pre-existing component. https://github.com/konveyor/virt-ui/issues/42

@gildub 

> I don't understand how 2 instances of a component could share a variable (saved mapping name).

I'm not sure what you mean, can you elaborate? It might be something we need to figure out once we address the form validation issue (which we'll need to do before we use these values in API calls). I'm tracking that in #40.

mturley:
FYI @gildub in the PR description if you use the word "Closes" or "Fixes" instead of "Solves" before the issue number, it'll automatically close the issue when the PR is merged. I don't think Solves is one of their magic words. You can tell because it'll have a dotted underline and indicate on hover that it is recognized as such.

gildub:
@mturley, thank you for pointing it. You're right `resolves` is a key word not `solves`, I looked at the list but took a wrong shortcut:  https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords.

vconzola:
@gildub I agree - the label for the new mapping name looks better above the field. Thanks. Also noticed, because the name is required, you need to add validation that the field is not empty and has a valid format when the user clicks Next. I'm sure there's already code for this someplace. @mturley can probably point you to it.

gildub:
@vconzola, thanks for the review.
For validation we are in progress to identify the best approach, so this is pending. For more details https://github.com/konveyor/virt-ui/issues/40

mturley:
yep @vconzola we are revisiting our form validation approach in #40 to prevent going down the same path that led to some issues last time. For now we're merging all these forms with no validation, we'll loop back and add it all.

mturley:
Copying over from Slack discussion for posterity: the weird state retention between the two MappingForm instances can be fixed with unique `key`s:

```diff
diff --git a/src/app/Plans/components/Wizard/PlanWizard.tsx b/src/app/Plans/components/Wizard/PlanWizard.tsx
index 1f365ac..d054d61 100644
--- a/src/app/Plans/components/Wizard/PlanWizard.tsx
+++ b/src/app/Plans/components/Wizard/PlanWizard.tsx
@@ -83,6 +83,7 @@ const PlanWizard: React.FunctionComponent<IPlanWizardProps> = ({
       component: (
         <WizardStepContainer title="Map Storage">
           <MappingForm
+            key="mapping-form-storage"
             mappingType={MappingType.Storage}
             mappingList={MOCK_STORAGE_MAPPINGS}
             availableSources={MOCK_STORAGE_MAPPING_SOURCES}
@@ -98,6 +99,7 @@ const PlanWizard: React.FunctionComponent<IPlanWizardProps> = ({
       component: (
         <WizardStepContainer title="Network Mapping">
           <MappingForm
+            key="mapping-form-network"
             mappingType={MappingType.Network}
             mappingList={MOCK_NETWORK_MAPPINGS}
             availableSources={MOCK_NETWORK_MAPPING_SOURCES}
```

gildub:
@mturley, ouch! React basics. Thank you.

src/app/Mappings/Storage/mocks/storage_mappings.mock.ts
```diff
@@ -0,0 +1,132 @@
+import { IStorageMapping, INetworkMapping, MappingType } from '../../types';
+import { ProviderType } from '@app/common/constants';
+import { MappingSource, MappingTarget } from '@app/Mappings/types';
+import { NetworkType } from '@app/Providers/types';
+
+const storageMapping1: IStorageMapping = {
```
mturley, line 6: This filename is `storage_mappings.mock.ts`, it contains both storage and network mappings, so maybe we should call it `mappings.mock.ts`

src/app/Mappings/components/MappingBuilder/MappingBuilder.tsx
```diff
@@ -8,6 +8,7 @@ import LineArrow from '@app/common/components/LineArrow/LineArrow';
 import MappingSourceSelect from './MappingSourceSelect';
 import MappingTargetSelect from './MappingTargetSelect';
 import './MappingBuilder.css';
+import { debug } from 'webpack';
```
mturley, line 11: Can you remove this unused import?

src/app/Plans/components/Wizard/MappingForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+    (isCreateMappingChecked) => !isCreateMappingChecked,
+    false
+  );
+
+  const [isSaveNewMapping, toggleSaveNewMapping] = React.useReducer(
+    (isSaveNewMapping) => !isSaveNewMapping,
+    false
+  );
+
+  const [mapping, setMapping] = React.useState<IStorageMapping | INetworkMapping | null>(null);
```
mturley, line 49: There's a `Mapping` union type that covers storage and network, this could be `<Mapping | null>`, not really important though.

src/app/Plans/components/Wizard/MappingForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+              helperTextInvalid="TODO"
+              // TODO add state/validation/errors to this and other FormGroups
+              validated="default"
+            >
+              <SimpleSelect
+                id="mappingSelect"
+                options={mappingOptions}
+                value={[mappingOptions.find((option) => option.value === mapping)]}
+                onChange={(selection) =>
+                  setMapping((selection as OptionWithValue<IStorageMapping>).value)
```
mturley, line 85: This should probably be `as OptionWithValue<Mapping>` or `as OptionWithValue<IStorageMapping | INetworkMapping>`, since it's storage-specific now. Eventually I hope to add generics to `SimpleSelect` though so we don't have to assert this type at all.

src/app/Plans/components/Wizard/MappingForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+            <Checkbox
+              label="Save mapping to use again"
+              aria-label="save mapping checkbox"
+              id="save-mapping-check"
+              isChecked={isSaveNewMapping}
+              onChange={toggleSaveNewMapping}
+            />
+            {isSaveNewMapping && (
+              <Grid className={spacing.mbMd}>
+                <GridItem sm={12} md={5} className={spacing.mbMd}>
```
mturley, line 117: Nice, I like the alignment with the builder grid.

mturley (commented):
A couple of tiny things and then I think this is good to go!

mturley (approved):
LGTM 👍  nice work here @gildub

PR #30, opened 2020-09-07
PlanWizard: Initial step 4 network mapping

gildub:
Solves https://github.com/konveyor/virt-ui/issues/29


gildub:
NetworkMapping has been replaced with a generic component handling both storage and network mappinps, see https://github.com/konveyor/virt-ui/pull/28

PR #32, opened 2020-09-09
Update lib-ui version and bring in useSelectionState hook

ibolton336:
Pulls in useSelectionState from lib-ui. Includes some small changes to parameters passed into the useSelectionState hook & a version bump for lib-ui.


src/app/Providers/components/ProvidersTable/CNV/CNVProvidersTable.tsx
```diff
@@ -13,11 +13,12 @@ import {
 import { DatabaseIcon } from '@patternfly/react-icons';
 import tableStyles from '@patternfly/react-styles/css/components/Table/table';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
-import { useSortState, usePaginationState, useSelectionState } from '@app/common/hooks';
+import { useSortState, usePaginationState } from '@app/common/hooks';
 import CNVProviderActionsDropdown from './CNVProviderActionsDropdown';
 import { ICNVProvider } from '@app/Providers/types';
 import ProviderStatus from '../ProviderStatus';
 import './CNVProvidersTable.css';
+import { useSelectionState } from '@konveyor/lib-ui';
```
mturley, line 21: It doesn't really matter, but out of personal preference I like to try and keep the npm imports above the local imports. We can leave it though, this isn't a hill I need to die on

src/app/Providers/components/ProvidersTable/VMware/VMwareProviderHostsTable/VMwareProviderHostsTable.tsx
```diff
@@ -35,7 +36,9 @@ const VMwareProviderHostsTable: React.FunctionComponent<IVMwareProviderHostsTabl
   };
   const { sortBy, onSort, sortedItems } = useSortState(hosts, getSortValues);
 
-  const { selectedItems, toggleItemSelected, selectAll } = useSelectionState<IHost>(sortedItems);
+  const { selectedItems, toggleItemSelected, selectAll } = useSelectionState<IHost>({
+    items: sortedItems,
```
mturley, line 40: To be safe, let's get in the habit of also passing an `isEqual` param here. I think all hosts will have a `metadata.uid` property, which we should add to our `IHost` interface and our mock data. Then we can use:
```
isEqual: (a, b) => a.metadata.uid === b.metadata.uid
```

And then below when we define the rows, we can use `isItemSelected(host)` from the hook in place of `selectedItems.includes(host)`.

This way if the hosts array changes in redux due to polling, we won't lose our selection state (since `includes` depends on the redux objects being referentially equal (`===`) to the ones that were present when the user selected them.

Honestly, this may be enough of a gotcha that we might want to make `isEqual` a required param. I could see this becoming a sneaky bug in places we neglect it.

src/app/Providers/components/ProvidersTable/VMware/VMwareProvidersTable.tsx
```diff
@@ -50,11 +51,11 @@ const VMwareProvidersTable: React.FunctionComponent<IVMwareProvidersTableProps>
 
   const { selectedItems, toggleItemSelected, areAllSelected, selectAll } = useSelectionState<
     IVMwareProvider
-  >(sortedItems);
+  >({ items: sortedItems });
   const {
     selectedItems: expandedProviders,
     toggleItemSelected: toggleProviderExpanded,
-  } = useSelectionState<IVMwareProvider>(sortedItems);
+  } = useSelectionState<IVMwareProvider>({ items: sortedItems });
```
mturley, line 58: Let's use `isEqual` and `isItemSelected` here as well. Providers already have `metadata.uid` in them in our mocks... oh, but they duplicate the same uid in the mock files. when we make the mock providers with the different names in the mocks let's also give them unique uids. Unless you want to use names as the identifier?

mturley (commented):
Looking good, just some notes about future-proofing with `isEqual` and `isItemSelected`.

src/app/Providers/components/ProvidersTable/VMware/VMwareProvidersTable.tsx
ibolton336: Good call - I am just going to use name for now & open an issue to circle back on generating UIDs for providers. Also I think we may need to consider the same for mappings.

mturley: Sounds good!

mturley (approved):
LGTM 🚀

PR #48, opened 2020-09-14
Add MappingDetailView and use in expanded content of mappings table rows

### LINKED ISSUE #43: Implement read-only mapping view for expanded table and plan wizard results step
We need a grouped read-only view of the mapping "box-and-arrow" visualization. We can reuse some of the structure from the grouped variant of my mapping builder that we abandoned (#24), so I have a good sense of how to do this, I'll focus on it next.

mturley:
Closes #43.

This implements the `MappingDetailView` component, which uses the same box-and-arrow layout from the `MappingBuilder` but in a read-only form with sources grouped by target.

Originally I was planning to use MappingBuilder with a prop to disable the editing controls, but this turned out to be different enough that it seemed cleaner to make it a separate component.

This PR adds it to the mappings table, we should also use it in the results step of the plan wizard in a future PR.

Screen Shot 2020-09-14 at 6 34 54 PM

### DIFF
--- a/src/app/Mappings/Network/NetworkMappingsPage.tsx
+++ b/src/app/Mappings/Network/NetworkMappingsPage.tsx
@@ -29,7 +29,7 @@ const NetworkMappingsPage: React.FunctionComponent = () => {
   React.useEffect(() => {
     console.log(`TODO: fetch network mapping items`);
     const currentMappings = fetchMockStorage(MappingType.Network);
-    setNetworkMappings(currentMappings || []);
+    setNetworkMappings((currentMappings as INetworkMapping[]) || []);
   }, [mockMapObj]);
 
   const [isAddEditModalOpen, toggleAddEditModal] = React.useReducer((isOpen) => !isOpen, false);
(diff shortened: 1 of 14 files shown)


mturley:
Damn, @ibolton336, I ended up pushing this branch upstream instead of my fork right after calling you out for it 😆  oops.

mturley:
Per @vconzola's comment here: https://github.com/konveyor/virt-ui/issues/31#issue-696183236

We forgot to remove the column with the number of mapped items. I'll do it in this PR since I'm already in there.

ibolton336:
image

Seeing an error when sorting inside the mappings table. I think this is because we dont have validation in place yet. We are trying to map through "mappings" when a network/storage mapping has been added with an empty "items" array. Would this ever be a valid use case? Do we need an empty state for the expanded mappings items section?

vconzola:
@ibolton336 I don't believe this is a valid use case. The create mapping button should be disabled in the modal until a source-target map has been added. So there should never be an "empty" mapping. Do you agree @mturley ?

mturley:
That sounds correct to me @vconzola . @ibolton336 this shouldn't be possible to hit once we have proper validation in the create mapping modal. If you want to add a condition for it to be safe though, I wouldn't be opposed to that.

ibolton336:
I am ok with just taking a note & moving on.

ibolton336 (approved):
LGTM +1 nicely done

PR #49, opened 2020-09-16
Common AddTooltip component

gildub:
Closes https://github.com/konveyor/virt-ui/issues/46

### DIFF
--- a/src/app/Plans/PlansPage.tsx
+++ b/src/app/Plans/PlansPage.tsx
@@ -33,14 +33,14 @@ const PlansPage: React.FunctionComponent = () => {
 
   const [isWizardOpen, toggleWizard] = React.useReducer((isWizardOpen) => !isWizardOpen, false);
 
-  let addPlanDisabledObj: Pick<IAddTooltipProps, 'isEnabled' | 'content'> = {
-    isEnabled: false,
+  let addPlanDisabledObj: Pick<IAddTooltipProps, 'isTooltipEnabled' | 'content'> = {
+    isTooltipEnabled: false,
     content: '',
   };
 
   if (vmwareList.length < 1 || cnvList.length < 1) {
     addPlanDisabledObj = {
-      isEnabled: true,
+      isTooltipEnabled: true,
       content: 'At least 1 source and 1 target providers are required to create a plan.',
     };
   }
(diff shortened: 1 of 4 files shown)


src/app/Plans/PlansPage.tsx
```diff
@@ -33,22 +33,15 @@ const PlansPage: React.FunctionComponent = () => {
...
-    };
-  }
-
-  if (cnvList.length < 1) {
+  if (vmwareList.length < 1 || cnvList.length < 1) {
     addPlanDisabledObj = {
-      isAddPlanDisabled: true,
-      disabledText: 'At least 1 OpenShift provider is required to create a plan.',
+      isEnabled: true,
+      content: 'At least 1 source and 1 target providers are required to create a plan.',
```
mturley, line 44: @vconzola this text will be used as a tooltip on the Create Plan button when it's disabled. What do you want it to say? Didn't see this in the mockup.

src/app/Plans/PlansPage.tsx
```diff
@@ -33,22 +33,15 @@ const PlansPage: React.FunctionComponent = () => {
...
-      disabledText: 'At least 1 VMware provider is required to create a plan.',
-    };
-  }
-
-  if (cnvList.length < 1) {
+  if (vmwareList.length < 1 || cnvList.length < 1) {
     addPlanDisabledObj = {
-      isAddPlanDisabled: true,
-      disabledText: 'At least 1 OpenShift provider is required to create a plan.',
+      isEnabled: true,
```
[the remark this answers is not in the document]
mturley: Oh, I misunderstood that `isEnabled` means the tooltip is enabled, so the button is disabled. That's a bit confusing, hmm. Maybe we should pick one and include the name of the thing in the boolean, e.g. `isButtonDisabled`, `isTooltipEnabled`, something like that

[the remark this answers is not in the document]
gildub: Yeah I was thinking about that too!

src/app/Plans/PlansPage.tsx
vconzola: @mturley First, I don't think we need any tooltip when the Create plan button is enabled - it's self-explanatory what it will do. When the button is disabled, I think it should say, "You must add at least VMware provider and one OpenShift Virtualization provider in order to create a migration plan." Will need to re-word when we have more sources than VMware, but we haven't surfaced the "source" and "target" terms up to this point in the workflow, so I don't want to introduce them here.

mturley: @vconzola did you mean "at least one VMware provider" instead of "at least VMware provider"?

gildub: @vconzola, please see the Tooltip message when no providers are available.

plan-wizard-add-disabled

vconzola:
@gildub Please spell out the word "one". Numbers should only be used in text if two digits or more.

gildub:
@vconzola, you probably meant the opposite, numbers should be spelled in numbers if two digits or more.
Please re-review. Thanks

mturley (approved):
LGTM

PR #51, opened 2020-09-16
Fix Service account token label and fieldId

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/50


mturley (approved):
Oops! Good catch, thanks @gildub

PR #61, opened 2020-09-21
Move hosts table to its own page

### LINKED ISSUE #11: Move host details to its own page in MTV UI
We should move host details from a row expansion under providers to its own page that can be drilled down to from the providers page. Reasons: scaling - we need to support 50 or more hosts on a single provider and row expansion does not work well.

ibolton336:
For VMware providers, there is now an additional nested route accessible via the hosts column in the providers table. 
Closes #11


package.json
```diff
@@ -80,6 +80,7 @@
     "@patternfly/react-styles": "^4.7.3",
     "@patternfly/react-table": "^4.16.7",
     "@patternfly/react-tokens": "^4.9.6",
+    "@types/react-router": "^5.1.8",
```
gildub, line 83: This can go into the `devDependencies` section.

gildub (commented):
LGTM besides a detail about package.json, see inline.

mturley:
@ibolton336 is this ready for review or still a draft?

ibolton336:
> @ibolton336 is this ready for review or still a draft?

Still a draft @mturley . Wrapping up the hosts Query stuff.

mturley:
Sounds good. I know you're busy with MTC as well, let me know if you'd like a hand.

ibolton336:
export interface IHostsByProvider {
  // [providerName: string]: IHost[];
  id: string;
  name: string;
  productName: string;
}

> Sounds good. I know you're busy with MTC as well, let me know if you'd like a hand.

@mturley Ready for review. Removed the IHostsByProvider in favor of IHost[] types. This is all still going to change probably so not sure where to leave things at the moment. Also, The only field available for IHost is the name field from the api.

mturley:
@vconzola can you give this a visual review? You can click through to the hosts of a VMware provider now. I think we maybe need to remove the `variant="compact"` on the table to give it some more room to breathe.

src/app/Providers/components/VMwareProviderHostsTable/VMwareProviderHostsTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const { selectedItems, toggleItemSelected, selectAll } = useSelectionState<IHost>({
+    items: sortedItems,
+  });
+
+  const rows: IRow[] = sortedItems.map((host: IHost) => {
+    const { name } = host;
+    return {
+      meta: { host },
+      selected: selectedItems.includes(host),
+      cells: [name, 'network N/A', 'bandwidth N/A', 'mtu N/A'],
```
mturley, line 48: Yeah, we need to get in touch with Jeff about how to get this data for hosts. Opened an issue: https://github.com/konveyor/virt-ui/issues/85

src/app/Providers/HostsPage.tsx
```diff
@@ -0,0 +1,84 @@
...
+          <Card>
+            <CardBody>
+              {!match?.params.providerId ? (
+                <EmptyState className={spacing.my_2xl}>
+                  <EmptyStateIcon icon={PlusCircleIcon} />
+                  <Title headingLevel="h2" size="lg">
+                    No hosts
+                  </Title>
+                  <EmptyStateBody>No hosts available for this provider.</EmptyStateBody>
+                </EmptyState>
```
mturley, line 74: This "no hosts" message only appears if the route doesn't have a providerId, but not if the hosts are loaded and that list is empty. We should probably handle that case, and maybe even just have the `providerId` prop of VMwareProviderHostsTable be nullable/optional, pass in whatever we've got here, and have a condition in there that will show "no hosts" if either the `providerId` is undefined or if the hosts array is empty.

src/app/Providers/components/VMwareProviderHostsTable/VMwareProviderHostsTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const [isSelectNetworkModalOpen, setIsSelectNetworkModalOpen] = React.useState(false);
+
+  return (
+    <>
+      {hostsQuery.isLoading ? (
+        <LoadingEmptyState />
+      ) : hostsQuery.isError ? (
+        <Alert variant="danger" title="Error loading hosts" />
+      ) : (
```
mturley, line 62: There could be another condition here for showing the "no hosts" empty state message if `hostsQuery.data.length === 0` or `!providerId`

src/app/queries/hosts.ts
```diff
@@ -0,0 +1,37 @@
...
+import { MOCK_HOSTS_BY_PROVIDER } from './mocks/hosts.mock';
+import { IHost } from './types';
+import { useProvidersQuery } from '.';
+
+export const HOSTS_QUERY_KEY = 'hosts';
+
+// TODO handle error messages? (query.status will correctly show 'error', but error messages aren't collected)
+export const useHostsQuery = (providerName: string): QueryResult<IHost[]> => {
+  const { isPollingEnabled } = usePollingContext();
+  const providersQuery = useProvidersQuery();
```
mturley, line 14: Awesome.

src/app/queries/mocks/hosts.mock.ts
```diff
@@ -1,41 +1,19 @@
...
-// if (process.env.NODE_ENV === 'test' || process.env.DATA_SOURCE === 'mock') {
-const host1: IHost = {
-  metadata: {
-    name: 'host1',
-    network: {
-      name: 'management network',
-      address: '192.168.0.0/24',
-      isDefault: true,
+if (process.env.NODE_ENV === 'test' || process.env.DATA_SOURCE === 'mock') {
+  MOCK_HOSTS_BY_PROVIDER = [
```
mturley, line 7: Since these aren't indexed by provider anymore let's just rename it to `MOCK_HOSTS`. Nitpick, I know.

mturley (changes requested):
Looking good, just some minor stuff!

src/app/Providers/components/VMwareProviderHostsTable/VMwareProviderHostsTable.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  const [isSelectNetworkModalOpen, setIsSelectNetworkModalOpen] = React.useState(false);
+
+  return (
+    <>
+      {hostsQuery.isLoading || providersQuery.isLoading ? (
+        <LoadingEmptyState />
+      ) : hostsQuery.isError ||
+        providersQuery.isError ||
+        hostsQuery?.data?.length === 0 ||
```
mturley, line 63: Isn't is possible to have a provider with no hosts? That wouldn't be an error, we'd just have the "no hosts" empty state appear here in that case right?

mturley: @ibolton336 if it makes the logic easier you could even lift the `useHostsQuery` up into the HostsPage and handle all the loading states / errors etc there, and only render the table component if there are defined hosts

vconzola:
@ibolton336 @mturley LGTM. The only thing I'm not sure about is whether the table should be compact. I think that's a simple change though, so let's leave it for now and we can see how many hosts per provider is "typical". If it's alot (> 10) we can make the table compact.

src/app/queries/hosts.ts
```diff
@@ -0,0 +1,35 @@
...
+        enabled: !!currentProvider,
+        refetchInterval: isPollingEnabled ? POLLING_INTERVAL : false,
+      },
+    },
+    MOCK_HOSTS
+  );
+  return {
+    ...result,
+    data: result.data,
+  };
```
mturley, line 34: @ibolton336 sorry to bug you with one last change but just in case, can we return `sortResultsByName(result)` here? (imported from `@app/queries/helpers`)? Considering the order of API results doesn't seem stable we need to sort all of our query data I think.

ibolton336: No prob

ibolton336: @mturley Should be ready now

mturley: Haha I saw the `sortIndexedResultsByName`.. maybe those helper names are confusing

mturley (approved):
LGTM! 🎉

src/app/queries/hosts.ts
ibolton336: I am ok with the names. I was just caught being lazy copying what you had for providers.

PR #65, opened 2020-09-22
Add mock and remote data source modes, install react-query, add useMockableQuery and useProvidersQuery, fetch providers from API

mturley:
This is a big one.

* Adds an .env file for use by webpack's Dotenv plugin to define build environment variables
  * `DATA_SOURCE`: `mock` or `remote`.  Whether to use static mock data or real data from a remote cluster. The `mock` name leaves open the opportunity to add a `local` mode to make real API requests on local data in the future.
  * `REMOTE_API_URL`: sets the base URL to be used for API calls in remote mode.
* Adds new package.json scripts to run in various modes
  * Running the development server:
    * `yarn start:dev` will use `NODE_ENV=development` and whatever value of `DATA_SOURCE` is in your .env file
    * `yarn start:dev:remote` will use `NODE_ENV=development` and `DATA_SOURCE=remote`
    * `yarn start:dev:mock` will use `NODE_ENV=development` and `DATA_SOURCE=mock`
  * Running a production build:
    * `yarn build` will use `NODE_ENV=production` and whatever value of `DATA_SOURCE` is in your .env file
    * `yarn build:remote` will use `NODE_ENV=production` and `DATA_SOURCE=remote`
    * `yarn build:mock` will use `NODE_ENV=production` and `DATA_SOURCE=mock`.
  * For local development with a real API, use `yarn start:dev:remote`.
  * For local development of layouts/views only, you can still use `yarn start:dev:mock`, which will continue to work offline.
  * The surge.sh preview links will use `yarn build:mock`.
* Installs react-query and react-query-devtools (dev dependency). Dev tools appear with a toggle in the lower-left when running in dev mode.
* Adds a new directory at `@app/queries` with files for custom query hooks at the root, and `mocks/` and `types/` subdirectories for centralized mock data and API-object-related TS types. Moves all existing mocks and relevant types into these directories.
* Adds a new custom hook `useMockableQuery` which wraps react-query's `useQuery` and checks the `DATA_SOURCE` variable, if you're in mock mode, it simulates a 1 second loading time and returns mock data instead of running a real API request.
* Replaces all references to "CNV" with "OpenShift" across the UI, to match the `"openshift"` provider type we see in the API data, since it would be annoying to have both names.
* Changes the structure of the `IVMwareProvider` and `IOpenShiftProvider` TS interfaces to match the real API data structure, and changes all usages of these objects throughout the UI to match.
* Adds a custom hook `useProvidersQuery` which calls `useMockableQuery` with the URL for listing providers.
* Uses `useProvidersQuery` in the providers page, plan wizard, and mappings modal to load real providers from the API! :tada:
* Adds error conditions to render alerts when providers query fails
* Puts mock provider data behind an env variable condition so webpack can tree-shake them away when built in prod mode :sparkles:
* Switches PR and master surge.sh preview scripts to use `yarn build:mock` :rocket:
* Adds default sorting to our tables since API data order is not guaranteed to be stable

We'll need the `REMOTE_API_URL` to be configurable at run-time instead of build-time, so I'll follow up with that in another PR since that'll require some changes to the Express server code.


src/app/queries/helpers.ts
```diff
@@ -4,13 +4,25 @@ import { UseQueryObjectConfig, QueryResult, useQuery } from 'react-query';
 // TODO what about usePaginatedQuery, useInfiniteQuery?
 // TODO what about alternate param signatures for useQuery?
 
+const mockPromise = <TResult>(data: TResult, timeout = 1000, success = true) => {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => {
```
ibolton336, line 9: Sweet! This is looking great

.env.example
```diff
@@ -0,0 +1,3 @@
+NODE_ENV=production
```
gildub, line 1: Might be a personal view but what about creating a .env file with local options, and add remote examples in comments. So we don't bother with the example file.

So we don't get annoyed with a locally (on dev side) changed file, would the following work: Commit the .env file and then add it to .gitignore ?

ibolton336: I think its easy enough to create a .env file & the process is well enough documented in the README.md. Also, I think the .env file is already added to the .gitignore.

mturley: That's fair, yeah. I'll do that. Also I honestly don't think we need `NODE_ENV` in it at all since webpack sets that for us iirc

mturley: oh, @gildub the problem with committing .env is that when you modify your local one to add the real API URL it hangs around as an uncommitted change, and we may accidentally commit that. I think I'm going to keep the .env.example for that reason. When we move that to a run-time thing maybe we can change it.

ibolton336 (approved):
LGTM! Really well done.

PR #66, opened 2020-09-22
Initial Plans table

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/56


vconzola (dismissed):
@gildub There's something funky going on with the table row layout. @mturley which layout should be used here - Flex or Grid, to get the different width columns as shown in the mockups? If grid, I can provide the appropriate span for each column.

src/app/Plans/components/PlanStatus.tsx
```diff
@@ -0,0 +1,35 @@
...
+  } else {
+    // TODO: This is only a placeholder more status check needs to be done once clarity with statuses.
+    const totalVMs = plan.spec.vmList.length;
+    const percentVMsDone = totalVMs > 0 ? (migration.status.nbVMsDone * 100) / totalVMs : 0;
+    const label = 'Running';
+
+    return (
+      <>
+        <StatusIcon status={StatusType.Warning} label={label} />
+        <Progress
```
mturley, line 26: I think to get the status icon and the "X of Y migrated" label on the same line like the mockups, you'll need to wrap the StatusIcon in a `<Level>` layout with the StatusIcon and the label as `<LevelItem>`s. I don't think we can use the `label` prop of Progress like this, not sure.

Alternatively if there's some way to get the `label` for `Progress` to appear above the bar, we could stick the `Level` with both things in there.

src/app/Plans/components/PlansTable.tsx
```diff
@@ -1,19 +1,119 @@
...
+          ),
+        },
+        plan.spec.provider.sourceProvider.name,
+        plan.spec.provider.destinationProvider.name,
+        plan.spec.vmList.length,
+        {
+          title: <PlanStatus plan={plan} migration={getMigration(plan)[0]} />,
+        },
+        {
+          title: buttonText ? (
```
mturley, line 82: I think the Start Migration button is intended to be aligned right next to the PlanActionsDropdown, it's sort of like a "pinned action". We can probably just have the last column be a fragment containing both.
```tsx
{
  title: (
    <>
      {buttonText ? <Button [......]>{buttonText}</Button>: null}
      <PlanActionsDropdown />
    </>
  )
}
```

You might have to put some right-margin on the button if they are too close together. For that you can use the `spacing` utility classes

src/app/Plans/components/PlanStatus.tsx
gildub: @mturley, yeah this one definitely needs more work, I was going to leave it to cover it with another PR once we have more details about the plan migration statuses from API. That said the I'll fix it with `Level` in the meantime.

src/app/Plans/components/PlansTable.tsx
gildub: Thanks, I was wondering about that second last column. I'll regroup them.

mturley:
@vconzola we can't really use flex or grid, it's just a table row which sort of behaves like flex. The column widths will adapt to the sizes of their contents, so let's make sure we get those just right first, then we can maybe set minimum widths on the contents of certain cells to persuade it a little.

src/app/Plans/components/PlanStatus.tsx
```diff
@@ -0,0 +1,39 @@
...
+  } else {
+    // TODO: This is only a placeholder, more statuses check needs to be done.
+    const totalVMs = plan.spec.vmList.length;
+    const percentVMsDone = totalVMs > 0 ? (migration.status.nbVMsDone * 100) / totalVMs : 0;
+    const label = 'Running';
+
+    return (
+      <>
+        <Level>
+          <LevelItem>
```
mturley, line 26: @gildub hey, sorry I steered you in the wrong direction here. There is still some weird spacing around this `<Level>`, I checked the docs for the Progress component and realized there are props to display this stuff for us. We can use the `label` prop for the status icon and the `valueText` prop for "VMs migrated" text, like in this example: https://www.patternfly.org/v4/components/progress/#finite-step

src/app/Plans/components/PlansTable.tsx
```diff
@@ -1,19 +1,119 @@
...
+        <LevelItem>
+          <Button variant="secondary" onClick={() => alert('TODO')} isDisabled={false}>
+            Create
+          </Button>
+        </LevelItem>
+        <LevelItem>
+          <Pagination {...paginationProps} widgetId="plans-table-pagination-top" />
+        </LevelItem>
+      </Level>
+      <Table
```
mturley, line 105: Let's also add a CSS file for the PlansTable so we can apply some custom alignment. If you look at the mockups, the cell contents are vertically centered, and in this code the columns in the middle are aligned with the top of the cell, which looks a little odd.

I think if we put a `className="plans-table"` on here and then add this CSS, it will fix it:

```css
.pf-c-table.plans-table tbody > tr > * {
  vertical-align: center
}
```

This will override the `vertical-align: baseline` style that PF provides here (a weird choice IMO). There might be a more React-y way to do this, not sure.

src/app/Plans/components/PlansTable.tsx
```diff
@@ -1,19 +1,119 @@
...
+    }
+
+    rows.push({
+      meta: { plan },
+      cells: [
+        {
+          title: (
+            <>
+              <a href="#">{plan.metadata.name}</a>
+              <div>{plan.spec.description}</div>
```
mturley, line 69: Also, the plan description text should be a little smaller than the plan name. We can use `<Text component="small">` for that, it has styles which will also make it a shade of gray.

mturley:
@gildub you have a rebase conflict here, and I have some final minor feedback above. After that I think this is ready to go!

mturley:
We also need to follow up after this to add the filters above the table, but I think it's ok to leave that for another PR. https://github.com/konveyor/virt-ui/issues/76

src/app/Plans/components/PlanStatus.tsx
gildub: @mturley, I actually initially used Progress props but there is an issue with title prop which doesn't like passing it a react element. I was thinking about creating an issue on PF project.

mturley: Ah, ok, I mixed up the `title` and `label` props. Ok, yeah, we can just put a `className={spacing.mbXs}` on the Level I guess to give it some bottom margin. Doing this though I notice there is still some weird white space to the right of the progress bar / the label hangs over the edge of it. Not sure what to do about that.

Yes, let's open an issue on patternfly-react for it, that `title` prop should be a `React.ReactNode` instead of a `string`.

mturley: It looks like we can solve the weird overhang by also putting `spacing.prMd` on the Level as well, at least until we address the upstream PF issue. So this should make it look right:

```ts
...
className={`${spacing.mbXs} ${spacing.prMd}`}
```
```

Comment (gildub) [reply src/app/Plans/components/PlanStatus.tsx:26]: Thanks @mturley, I've created https://github.com/patternfly/patternfly-react/issues/4875
Using the spacing work arounds in the meantime.

Comment (mturley) [reply src/app/Plans/components/PlanStatus.tsx:26]: Perfect, thanks for opening that!

File: src/app/Plans/components/PlanStatus.tsx [hunk 6]
```diff
@@ -0,0 +1,40 @@
...
+            <StatusIcon status={StatusType.Warning} label={label} />
+          </LevelItem>
+          <LevelItem className={`${spacing.mbXs} ${spacing.prMd}`}>
```
mturley, line 30: Sorry to keep nitpicking @gildub but can you put these classes on the whole `<Level>` instead of just the `<LevelItem>`? the first item isn't vertically aligned anymore with this

gildub: No worries, done.
It makes sense but I don't see much of a difference. Maybe I'm missing something.

mturley (approved):
Looks good to me! @vconzola, want to give this a final review? Note that it's just the Plan level table, migrations will come in another PR. Also the filter bar at the top of the table will be in another PR. As for column widths, if there is a longer plan name the columns will readjust.

Screen Shot 2020-09-24 at 10 40 02 AM

src/app/Plans/components/PlanStatus.tsx
mturley: It's subtle, but it's there. The right side was a few pixels higher. I know, nitpicky 😄 
Screen Shot 2020-09-24 at 10 42 45 AM

Screen Shot 2020-09-24 at 10 42 28 AM

gildub: Oh I see, reducing my screen helps!
Well in UX/UI every pixel matter!
Thanks

vconzola:
@mturley As we discussed in our 1-1, can we right align the button and kebab column and set a min column width so that the Target providers column header is never truncated? Also, I don't know if this is PF ok, but can we reduce the size of the text above the progress bar to 14 px to increase the space between the status message and the number of VMs?

mturley:
Yeah, @vconzola @gildub I'll make a commit you can cherry-pick to address these things, it'll be easier than repeating everything we discussed

gildub:
@vconzola could you please do another review?

vconzola:
@gildub It's looking good. There's one more change I'd like to see, if possible. @mturley If we set a min width on the status column and changed the Button label to "Start" would that elongate the progress bar. It still looks kinda weird with the message and VM summary so close to each other. If you'd like I can open a separate issue for this and we can come back to it later.

gildub:
@vconzola,

I've defined the status column at 30% cell width to give it more room.
I've also put the "Start migration" on 2 lines (you might not like it, just trying options), which makes the action column smaller but keep the full meaning.
I thought about reducing action column spacing (padding and margin) but I haven't found yet how to apply that only to last cell (--pf-c-table--Cell--PaddingLeft, etc).

There is also an issue when resizing the display to smaller size as the Kebab moves under the button, which looks a bit strange. Although we can force the two (button and kebab) to stay inline, that would make that column kinda of big compare to the other ones. One solution would be to have all actions inside the kebab or to split them into two columns (but it then gives more space just for the actions, unless shrinking their spacing if possible).

It seems there is an overall issue with the Progress bar inside a table. 
This is look worse when making the display smaller because the progress bar, in essence, needs horizontal space no matter what, so it's always going to be a fight against the other columns.
Looking for a better fit, what about using a [Donut utilization chart](https://www.patternfly.org/v4/documentation/react/charts/chartdonututilization) instead?

+1 to take it from here to another PR so this one doesn't block the VMs migration plan details.

mturley:
The donut chart variant takes up way too much vertical space IMO, making the rows super tall also looks strange. I think maybe what we can do is:
1. put a CSS class on the progress bar component and give it a static minimum width, either that or put some significant margin/padding to the right of the status message and the left of the VM count message, to force that div to grow.
2. modify the CSS class I already put on the actions column to add `white-space: nowrap` which I believe will prevent it from ever wrapping onto multiple lines. We can probably also do the same thing for the status text above the progress bar.
3. We can maybe change "Start migration" to "Start" or "Migrate"?

Yes, let's open a new issue so we can merge this and unblock the next PR.

gildub:
@mturley, yeah the donut takes a bit of vertical space, not ideal either.
Looking forward to implement your recommendation.

gildub:
@vconzola, could you please ack so we can take from here. Thanks

src/app/Plans/components/PlansTable.tsx
```diff
@@ -51,18 +52,27 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
   const columns: ICell[] = [
     { title: 'Name', transforms: [sortable, wrappable] },
     { title: 'Source provider', transforms: [sortable, wrappable] },
-    { title: 'Target provider', transforms: [sortable] },
+    { title: 'Target provider', transforms: [sortable, wrappable] },
     { title: 'VMs', transforms: [sortable] },
-    { title: 'Plan status', transforms: [sortable] },
-    { title: '', columnTransforms: [classNames(alignment.textAlignRight)] },
+    { title: 'Plan status', transforms: [sortable, cellWidth(30)] },
```
mturley, line 57: Ah, I didn't realize there was a cellWidth transform. Nice find.

gildub: Thanks!

mturley:
Actually @gildub I think if you grab this commit that addresses everything Vince has pointed out so far: https://github.com/konveyor/virt-ui/commit/283f25ce5960bde7b28b61ef70ae262a6d80b65f

gildub:
@mturley, nice the no-wrap property, that's it!

mturley:
Ok @vconzola sorry for the noise, this is ready for a final review for reals this time. I think it's looking good

mturley (approved):
I got an ack from @vconzola over slack, going to go ahead and merge this one. 👍

vconzola:
@gildub @mturley This looks great and handles responsiveness very well. Nice work guys.

PR #70, opened 2020-09-23
Add useStorageClassesQuery, combine with mocks in useMappingResourceQueries, factor out helpers, load storage classes from API

### LINKED ISSUE #67: Fetch storage classes for each provider (first dependent query)
Storage classes are a separate resource keyed by provider, we can't fetch all of them at once. So we need a storage classes query that depends on the providers query and gets storage classes for a certain provider. This will be used in the expanded content of the OpenShift providers table and in the mapping builder.

mturley:
Closes #67 

`useStorageClassesQuery` takes a nullable array of providers, and if it is null the query is disabled. Once the providers become defined, the query fetches storage classes for each provider (concurrently runs a fetch for each one, bundled up into one promise with `Promise.all`), and reduces them into a lookup object indexed by provider name. It's a bit weird since in some places we call it with only one provider and in some we call it with multiple, and those two result caches can't be shared because of the different keys. I don't think it's worth optimizing away though, and this pattern guarantees correctness when providers change.

`useMappingResourceQueries` is intended as a wrapper around all queries needed for the `availableSources` (networks, datastores) and `availableTargets` (networks, storage classes) used in various mapping-related views. Currently the only real query it uses is `useStorageClassesQuery`, and it uses the existing mock data for the rest of its return values.

As part of making `useMappingResourceQueries` I realized it would be useful to be able to mount multiple query hooks and get an aggregate status for all of them, so I abstracted out that behavior into the helper functions `getAggregateQueryStatus` and `getFirstQueryError`. Those can be used to get an overall view of whether any of the queries in the list are loading or errored, or whether all of them are idle or successful. If we need to use a bunch of queries together and show one loading state, we can use these.

I refactored the storage class types from plain strings to objects with `name` properties, which had some ripple effects and made some of our logic unnecessary, so I cut some of that out.

Then, I used these new queries to properly provide storage classes to the AddEditMappingModal, MappingDetailView (mappings table) and OpenShiftProvidersTable ✨

### DIFF
--- a/src/app/Mappings/Network/NetworkMappingsPage.tsx
+++ b/src/app/Mappings/Network/NetworkMappingsPage.tsx
@@ -2,9 +2,7 @@ import * as React from 'react';
 import {
   PageSection,
   Title,
-  Bullseye,
   EmptyState,
-  Spinner,
   EmptyStateIcon,
   EmptyStateBody,
   Button,
@@ -17,6 +15,7 @@ import { INetworkMapping, MappingType } from '@app/queries/types';
 import MappingsTable from '../components/MappingsTable';
 import AddEditMappingModal from '../components/AddEditMappingModal';
 import { fetchMockStorage } from '@app/queries/mocks/helpers';
+import LoadingEmptyState from '@app/common/components/LoadingEmptyState';
 
 const isFetchingInitialNetworkMappings = false; // Fetching for the first time, not polling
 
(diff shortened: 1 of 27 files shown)


mturley:
Had to fix a few things, this should be ready now.

ibolton336 (approved):
LGTM :+1:

PR #77, opened 2020-09-23
Use express to provide env vars at run time, add concurrent express dev server to support backend-only routes

### LINKED ISSUE #68: REMOTE_API_URL needs to be a run-time configuration variable instead of a build-time one.
Right now, the .env file contents are statically replaced by webpack at build time wherever they are referenced in the source. This means you can't change the API URL on a deployed build, you'd have to rebuild it.

We need to do something similar to the MigMeta object in MTC and ship the value of REMOTE_API_URL to the client at run time via the Express server.

mturley:
Closes #68 

This PR adds a similar pattern to the migMeta file from mig-ui, but using the same values in our existing `.env` file. It also adds support for backend-only routes served directly from express, which we will need for OAuth. This is very similar to what I did in https://github.com/konveyor/mig-ui/pull/957.

* In dev mode, Express is run concurrently on port 9001 alongside webpack-dev-server. Requests to specific URLs listed in `webpack.dev.js`'s proxy config (currently only `/hello`) are sent to port 9001, while the rest of the app is still served at port 9000 as normal. stdout from both servers are combined in the same console, color-coded by source. The express server is set up with `nodemon` and will restart itself when `server.js` changes.
* There is a new file `runtime-env-vars.js` which lists the variable names from `.env` that should be served at run time and encodes them.
* The HTML to bootstrap the app is now defined in `index.html.ejs`, which has an additional bit of script to decode a base64-encoded JSON blob passed in as a template param called `_env_encoded`.
  * When running in dev mode, OR when building in prod mode with `DATA_SOURCE=mock`, webpack will output a `dist/index.html` file just as before, and inject the `_env_encoded` value at build time from values in the `.env` file. The app can then still be hosted statically (we keep our PR previews).
  * When running in prod mode, webpack will only inject the JS and CSS bundles, but will not inject the `_env_encoded` parameter, and it will output a `dist/index.html.ejs` file. Express will then render that ejs file instead of serving up a static .html file, and it will inject the values found in the `.env` file when Express started up.

We are still using webpack's Dotenv plugin to inject values at build time when the UI references `process.env.*`, since we can still tree-shake away our mock data that way by referencing `process.env.DATA_SOURCE`. However, we can now also use `RUNTIME_ENV.*` in our UI code, which is a decoded form of `_env_encoded`. We should use this for any values that need to change on a per-deployment basis, such as API URLs.

This PR also replaces the `REMOTE_API_URL` variable in `.env` with `INVENTORY_API_URL`, and adds `CLUSTER_API_URL` for future use.

To test it out:
* Run `yarn start:dev:remote` or `yarn start:dev:mock` and see that things still work as intended. In your browser's console you can type `_env` to see the values that were injected by webpack. You can also navigate to `/hello` to see the route proxied to Express.
* Run `yarn build:mock` and see that you still get a `dist/index.html` file with encoded vars in it, which you can serve on port 8080 with `yarn start:mock`. You can still look at `_env` in the console and check that `/hello` works.
* Run `yarn build:remote` (or just `yarn build`, if you have `DATA_SOURCE=remote` in your `.env`), and see that you get a `dist/index.html.ejs` file with a placeholder for the vars. Run it with `yarn start`, and at localhost:8080 you can type `_env` in the console to see the values injected by Express. **Change the contents of `.env` and restart `yarn start`, without rebuilding, and you'll see the values have changed.**
* Notice that I didn't break the PR previews :)


mturley:
cc @eriknelson

ibolton336 (approved):
LGTM!

PR #78, opened 2020-09-24
Plans details table

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/57

### DIFF
--- a/src/app/Plans/components/VMMigrationDetails.tsx
+++ b/src/app/Plans/components/VMMigrationDetails.tsx
@@ -30,6 +30,7 @@ import { IMigration } from '@app/queries/types';
 import { useSortState, usePaginationState } from '@app/common/hooks';
 import VMStatusTable from './VMStatusTable';
 import PipelineSummary from '@app/common/components/PipelineSummary';
+import { PlanStatusConditionsType } from '@app/common/constants';
 
 import { MOCK_MIGRATIONS } from '@app/queries/mocks/plans.mock';
 
@@ -84,9 +85,13 @@ const VMMigrationDetails: React.FunctionComponent = () => {
   currentPageItems.forEach((migration: IMigration) => {
     const isExpanded = isVMExpanded(migration);
 
-    let buttonText = '';
-    if (migration.other.status !== 'Ready') {
-      buttonText = 'Cancel';
+    let isButtonCancel = false;
+    if (
+      migration.other.status !== PlanStatusConditionsType.Ready &&
+      migration.other.status !== PlanStatusConditionsType.Finished &&
+      migration.other.status !== PlanStatusConditionsType.Error
+    ) {
+      isButtonCancel = true;
     }
 
     rows.push({
@@ -103,10 +108,10 @@ const VMMigrationDetails: React.FunctionComponent = () => {
           title: <PipelineSummary total={0} current={0} status={migration.other.status} />,
         },
         {
-          title: buttonText ? (
+          title: isButtonCancel ? (
             <>
               <Button variant="secondary" onClick={() => alert('TODO')} isDisabled={false}>
-                {buttonText}
+                Cancel
               </Button>
             </>
           ) : null,
--- a/src/app/Plans/components/VMMigrationDetails.tsx
+++ b/src/app/Plans/components/VMMigrationDetails.tsx
@@ -1,0 +1,217 @@
+import * as React from 'react';
+import { useRouteMatch } from 'react-router-dom';
+import {
+  Breadcrumb,
+  BreadcrumbItem,
+  Button,
+  Card,
+  CardBody,
+  Pagination,
+  PageSection,
+  Title,
+} from '@patternfly/react-core';
+import {
+  classNames,
+  Table,
+  TableHeader,
+  TableBody,
+  TableVariant,
+  sortable,
+  classNames as classNamesTransform,
+  ICell,
+  IRow,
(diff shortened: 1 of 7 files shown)


mturley:
@gildub I noticed your status on slack mentioned needing to add stuff to this page, is this PR ready for review or should it be a draft?

gildub:
@mturley, I've updated it. Thanks

src/app/Plans/components/VMMigrationDetails.tsx
```diff
@@ -0,0 +1,217 @@
...
+  },
+  status: vmStatus2,
+  other: {
+    copied: 87952,
+    total: 87952,
+    status: 'Running',
+  },
+};
+
+const MOCK_TEMP_MIGRATIONS: ITempMockMigration[] = [mockMigration1, mockMigration2];
```
mturley, line 80: Can we move these to their own mocks file in `@app/queries/mocks/migrations.mock.ts`, or maybe add them to `plans.types.ts`? Even if we're going to replace them later it would de-clutter this a bit.

src/app/Plans/components/VMMigrationDetails.tsx
```diff
@@ -0,0 +1,217 @@
...
+    { title: 'End time', transforms: [sortable] },
+    { title: 'Data copied', transforms: [sortable] },
+    { title: 'Status', transforms: [sortable] },
+    { title: '' },
+  ];
+
+  const rows: IRow[] = [];
+
+  currentPageItems.forEach((migration: ITempMockMigration) => {
+    const isExpanded = expandedVMs.includes(migration);
```
mturley, line 131: Just in case, let's use the `isItemSelected` function returned from `useSelectionState` to determine expanded state instead of `.includes()`, i'm worried we'll lose this state when queries refetch because of referential equality with stale data.

src/app/Plans/components/VMMigrationDetails.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+    let buttonText = '';
+    if (migration.other.status !== 'Ready') {
+      buttonText = 'Cancel';
+    }
+
+    rows.push({
+      meta: { migration },
+      isOpen: isExpanded,
+      // isOpen: true,
```
mturley, line 141: Can you remove the commented line here?

src/app/Plans/components/VMMigrationDetails.tsx
```diff
@@ -0,0 +1,217 @@
...
+        </Breadcrumb>
+        <Title headingLevel="h1">Migration Details by VM</Title>
+      </PageSection>
+      <PageSection>
+        <Card>
+          <CardBody>
+            <Pagination {...paginationProps} widgetId="migration-vms-table-pagination-top" />
+            <Table
+              aria-label="Migration VMs table"
+              variant={TableVariant.compact}
```
mturley, line 193: I think we can remove the `variant` prop here, this table doesn't look compact in the mocks. The VMStatusTable should stay compact though.

src/app/queries/mocks/plans.mock.ts
```diff
@@ -161,40 +160,160 @@ const plan2: IPlan = {
 MOCK_PLANS = [plan1, plan2];
 
 export const migration1: IMigration = {
```
mturley, line 162: Ah, I hadn't realized we already had `IMigration`. What's the reason for having both this and the `ITempMockMigration` from above?

mturley (changes requested):
This is looking really good, I just have a few minor comments.

src/app/queries/mocks/plans.mock.ts
gildub: @mturley, this is temporary to avoid touching back and forth the structure until we have more visibility on those structure.
But that's annoying. Let me see if I can do better now that I have the UI in place.

src/app/Plans/components/VMMigrationDetails.tsx
gildub: Yes, sorry about that, I consolidated the migration mock data.

src/app/queries/mocks/plans.mock.ts
gildub: I've removed it after consolidating the mock migrations.

src/app/Plans/components/VMMigrationDetails.tsx
gildub: Done.

src/app/Plans/components/VMMigrationDetails.tsx
mturley: No need to apologize! 👍

gildub:
@vconzola, please review.

mturley (dismissed):
LGTM, but I'll wait for @vconzola's review before merging.

gildub:
As discussed today during Scrum, all actions but the enable edit/remove options from Dropdown (Kebab) for the `PlansTable`.

I suggest to do that last one into a separate PR otherwise I'll come back to it first thing tomorrow.

gildub:
@vconzola, please check it out.

mturley:
yeah, I think that can be considered out of scope since it's on the main plans page and not the plan details page. We can open an issue for that.

mturley (approved):
Approved by Vince over Slack. Opened #98 to follow up about the kebab actions.

Nice work @gildub !

PR #82, opened 2020-09-28
Refactors Progress bar

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/58


src/app/Plans/components/PlanStatus.tsx
```diff
@@ -1,43 +1,52 @@
...
+      title = 'Failed';
+      variant = ProgressVariant.danger;
+      break;
+    case 'Finished':
+      title = 'Complete';
+      variant = ProgressVariant.success;
+      break;
+    default:
+      title = 'Running';
+      break;
```
mturley, line 35: I usually avoid `switch` statements because of some issues I ran into in the past... but, I can't really remember why... so maybe this is fine? @ibolton336 do you have any opinions about `switch` vs `if-else` for something like this?

Also I think you don't need a `default` case here, instead you could just have `let title = 'Running'` be the default value above. It seems in that same case you never set a value for `variant`, is that intentional?

mturley (commented):
This looks great! A few minor questions about the `switch` logic.

src/app/Plans/components/PlanStatus.tsx
gildub: @mturley, I wasn't happy enough with that piece, I've re-factored .

mturley: Let's just keep it, I'm realizing my take on it is personal preference. It is weird that it's the only JS syntax that doesn't use brackets around things that seem like blocks :)

gildub: @mturley, you're right, there was actually some duplicated logic between the condition check between the caller and PlanStatus, I removed the switch.

mturley: oh, ok cool!

mturley:
Sorry, looks like my PRs Ian just merged caused conflicts here

src/app/Plans/components/PlanStatus.tsx
gildub: @mturley, sorry I added one more to use more types instead of string.

gildub:
@mturley, No worries.

It's rebased and I also added one more to use more types instead of string.

src/app/common/constants.ts
```diff
@@ -12,6 +12,20 @@ export enum ProviderType {
...
+  finished = 'Finished',
+  error = 'Error',
+}
+
+export enum PlanStatusType {
+  ready = 'Ready',
+  running = 'Running',
+  finished = 'Complete',
+  error = 'Failed',
+}
```
mturley, line 27: Is the value in the actual API capitalized or no? For clarity we may want to use the same exact string here for the key and value, so either `Ready = 'Ready'` or `ready = 'ready'`

gildub:
@vconzola, please let me know what you think as we are now using PF progress bar as intended.

vconzola:
@gildub Looks great! Thanks for making the changes.  :-)

src/app/Plans/components/PlansTable.tsx
```diff
@@ -66,10 +77,33 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
 
   currentPageItems.forEach((plan: IPlan) => {
     let buttonText: string | React.ReactNode;
-    if (plan.status.conditions.every((condition) => condition.type === 'Ready')) {
+    let isStatusReady = false;
+    let title = '';
+    let variant: ProgressVariant | undefined;
+
+    if (
+      plan.status.conditions.every((condition) => condition.type === PlanStatusConditionsType.ready)
```
ibolton336, line 85: Looking good, Gilles. One thing -  I am wondering how this logic reads- I think there can be a case where there is a ready condition & some additional warning conditions in the conditions array?

ibolton336 (changes requested):

src/app/Plans/components/PlansTable.tsx
gildub: Thanks for the review; 
I replaced the conditions check to search for only one valid case.

src/app/Plans/components/PlansTable.tsx
```diff
@@ -49,6 +61,16 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
 
   const getMigration = (plan) => migrations.filter((migration) => migration.plan === plan);
 
+  const ratioVMs = (plan) => {
```
mturley, line 64: Sorry I didn't catch this before. Can we add a type to the arguments of these functions?
`const getMigration = (plan: IPlan) => { ...`
`const ratioVMs = (plan: IPlan) => { ...`

I wonder if we should add `noImplicitAny` in our tsconfig... maybe later.

src/app/Plans/components/PlansTable.tsx
```diff
@@ -66,10 +88,29 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
 
   currentPageItems.forEach((plan: IPlan) => {
     let buttonText: string | React.ReactNode;
-    if (plan.status.conditions.every((condition) => condition.type === 'Ready')) {
+    let isStatusReady = false;
+    let title = '';
+    let variant: ProgressVariant | undefined;
+
+    const hasCondition = (type) => {
```
ibolton336, line 95: I am ok committing this for now. Lets open an issue to create a fine tuned this status helper function & dig deeper into how status will work with the MTV API.

mturley: Agreed. Let's put a `: string` type annotation on that `type` param though.

src/app/Plans/components/PlansTable.tsx
gildub: @mturley, done. 

+1 for `noImplicityAny`

mturley (approved):
LGTM!

PR #83, opened 2020-09-28
Add queries for datastores and networks to complete implementation of useMappingResourceQueries

### LINKED ISSUE #72: Flesh out useMappingResourceQueries to load networks/storages for use in mappings
I'll also add form validation to the mapping modal while I'm at it.

mturley:
Closes #72.

Adds real queries for the rest of the `useMappingResourceQueries` hook (datastores, networks), and refactors interfaces/mocks for the related objects. All our views related to datastores/networks/storageClasses will now use real API data 🎉  Also cleans up some of the other mocks for consistency.

I noticed we started being a little repetitive with our code to sort the results of queries, so I factored those out into helper functions like `sortResultsByName` and `sortIndexedResultsByName` that can take any `QueryResult` whose data objects have a name property and return the result with the data sorted by name.


ibolton336 (approved):
LGTM!

PR #84, opened 2020-09-28
Fix remote-mode error caused by premature disabling of storageClasses mock

mturley:
Prevents the app from crashing in `start:dev:remote` mode. Good catch @ibolton336


ibolton336 (approved):
:+1:

PR #87, opened 2020-09-29
MappingBuilder enable/disabled add and remove buttons

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/42, https://github.com/konveyor/virt-ui/issues/96


src/app/Mappings/components/MappingBuilder/MappingBuilder.tsx
```diff
@@ -30,7 +30,16 @@ export const MappingBuilder: React.FunctionComponent<IMappingBuilderProps> = ({
   builderItems,
   setBuilderItems,
 }: IMappingBuilderProps) => {
-  const reset = () => setBuilderItems([{ source: null, target: null }]);
+  const [isAddSourceDisabled, toggleAddMappingDisable] = React.useReducer(
+    (isToggled) => !isToggled,
+    false
+  );
```
mturley, line 36: We don't need to have a separate state value for whether the button is disabled, since that can be derived from observing the `builderItems`. Keeping it separate and trying to sync the two could cause some bugs.

vconzola:
LGTM. Thanks @gildub .

vconzola:
@gildub Oops! After talking with @mturley I need to rescind my "LGTM" comment. :-(  We want to disable the Add button until a source and target have been selected. Sorry for the confusion.

src/app/Mappings/components/MappingBuilder/MappingBuilder.tsx
```diff
@@ -58,7 +58,7 @@ export const MappingBuilder: React.FunctionComponent<IMappingBuilderProps> = ({
         <Text component="p">{instructionText}</Text>
       </TextContent>
       {builderItems.map((item, itemIndex) => {
-        const key = item.source ? `${item.source.id}` : `empty-${itemIndex}`;
+        const key = item.source ? `${item.source.id}-${itemIndex}` : `empty-${itemIndex}`;
```
mturley, line 61: We should also update our `MOCK_VMWARE_DATASTORES` data so the ids are actually different in each one, which would make this unnecessary-- technically if the items in this list change order or one is removed etc the key is supposed to remain the same for the same object it represents, so React doesn't have to re-render its children. But these components are trivial to render so it's fine

gildub:
@vconzola, that's what I understood and I made the change accordingly just that I hadn't push the update. You can now see it in action.

src/app/Mappings/components/AddEditMappingModal.tsx
```diff
@@ -60,6 +60,8 @@ const AddEditMappingModal: React.FunctionComponent<IAddEditMappingModalProps> =
     mappingType
   );
 
+  const isEveryItemFilled = () => builderItems.every((item) => item.source && item.target);
```
mturley, line 63: I don't see what we gain by making this a function, since it is initialized and called all in the same function scope as `builderItems`, and doesn't take any arguments. If we put it after the `useState` below it could just be a variable:

```ts
const isEveryItemFilled = builderItems.every((item) => item.source && item.target);
```

and then similarly the prop type for passing it to MappingBuilder would become a boolean.

I think making things like this a function is only useful if they need to be used in multiple scopes with arguments from each scope, like if we had two different variants of this modal that both had their own set of `builderItem`s.

src/app/Mappings/components/MappingBuilder/MappingBuilder.tsx
```diff
@@ -42,19 +46,18 @@ export const MappingBuilder: React.FunctionComponent<IMappingBuilderProps> = ({
...
-    selectSourcePlaceholder = 'Select source...';
-    selectTargetPlaceholder = 'Select target...';
   }
   if (mappingType === MappingType.Storage) {
     instructionText = 'Map source datastores to target storage classes.';
-    selectSourcePlaceholder = 'Select source...';
-    selectTargetPlaceholder = 'Select target...';
   }
 
+  const allSourcesExhausted = () => builderItems.length === availableSources.length;
```
mturley, line 59: Same here, I don't think it needs to be a function, just a plain boolean.

src/app/Plans/components/Wizard/MappingForm.tsx
```diff
@@ -61,6 +61,9 @@ const MappingForm: React.FunctionComponent<IMappingFormProps> = ({
     { source: null, target: null },
   ]);
 
+  // TODO add to disable Next step
+  const isEveryItemFilled = () => builderItems.every((item) => item.source && item.target);
```
mturley, line 65: Ah, okay, we do use `isEveryItemFilled` in multiple places. I'd say either make both copies of it plain booleans or move it to a `helpers.ts` file somewhere and have it take `builderItems` as an argument. It's pretty trivial, so either way is fine with me, I don't think we lose much having it duplicated.

src/app/common/components/AddTooltip.tsx
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
 import { Tooltip } from '@patternfly/react-core';
+import { TooltipProps } from '@patternfly/react-core/dist/js/components/Tooltip/Tooltip';
```
mturley, line 3: You should be able to just add TooltipProps to the import from `react-core` in the line above.

mturley (commented):
LGTM, just some really minor nitpicks.

src/app/common/components/AddTooltip.tsx
gildub: I know but it doesn't work for me: 

`Module '"../../../../node_modules/@patternfly/react-core/dist/js"' has no exported member 'TooltipProps'.ts(2305)`

src/app/Mappings/components/AddEditMappingModal.tsx
gildub: Yes functions are best for reusability but sometimes just for readability.
That said it's called in two different places.

src/app/Plans/components/Wizard/MappingForm.tsx
gildub: Yes, I agree.

src/app/Mappings/components/MappingBuilder/MappingBuilder.tsx
gildub: Ok that one is called only once although I thought this was clearer. I changed it.

gildub:
@vconzola, please review.
One little change: I placed the tooltip on the bottom so it would not hide the selection if it was partly filled.

src/app/common/components/AddTooltip.tsx
mturley: Aha! Looks like that was a bug in PF, it was supposed to be exported and they fixed it in a newer release. if I update PF on my end it fixes it. Sounds like that's out of scope for this PR though so we can leave this

src/app/Mappings/components/AddEditMappingModal.tsx
```diff
@@ -167,6 +168,7 @@ const AddEditMappingModal: React.FunctionComponent<IAddEditMappingModalProps> =
                   availableTargets={mappingResourceQueries.availableTargets}
                   builderItems={builderItems}
                   setBuilderItems={setBuilderItems}
+                  isEveryItemFilled={() => builderItems.every((item) => item.source && item.target)}
```
mturley, line 171: It doesn't really matter much, but it's still unnecessary for this to be a function. You can just pass `isEveryItemFilled={builderItems.every((item) => item.source && item.target)}` and change the prop to a boolean, and usages from `isEveryItemFilled()` to `isEveryItemFilled`.

src/app/Mappings/components/AddEditMappingModal.tsx
```diff
@@ -95,6 +95,7 @@ const AddEditMappingModal: React.FunctionComponent<IAddEditMappingModalProps> =
               onClose();
             }
           }}
+          isDisabled={!builderItems.every((item) => item.source && item.target)}
```
mturley, line 98: Oh, actually @gildub if you're not reusing `isEveryItemFilled` here, you might as well just not even have it be a prop for MappingBuilder, and just have it be defined inside MappingBuilder. It's sort of an implementation detail anyway. I think that might be cleaner... sorry for the nit pick..

src/app/Mappings/components/AddEditMappingModal.tsx
gildub: Yes, wasn't pay attention there! Thanks

src/app/Mappings/components/AddEditMappingModal.tsx
gildub: @mturley, sorry I'm not following you here, `isEveryItemFilled` flag needs to be shared by `MappingForm` too.

mturley (approved):
LGTM, will wait for @vconzola to review before merging.

src/app/Mappings/components/AddEditMappingModal.tsx
mturley: Oh right, I think I got confused because I saw you wrote out `builderItems.every((item) => item.source && item.target)` twice in this file, so I figured you could just write it out everywhere it's used instead of passing it as a prop. Either way is fine though, it's not worth holding up the PR

gildub: Well I see now what you meant.

mturley:
@gildub once you add the tooltip message for the other case (all exhausted) this will be ready to go

gildub:
@mturley, @vconzola, I believe we now have all cases covered.

src/app/Mappings/components/MappingBuilder/MappingBuilder.tsx
```diff
@@ -30,6 +32,16 @@ export const MappingBuilder: React.FunctionComponent<IMappingBuilderProps> = ({
   builderItems,
   setBuilderItems,
 }: IMappingBuilderProps) => {
+  const messageSelectBoth = 'You must select a source and target before adding another mapping.';
+  const messageExhausted = `All sources ${mappingType.toLowerCase()} have been mapped.`;
```
mturley, line 36: This creates the messages "all sources network have been mapped" and "all sources storage have been mapped". We need to instead use actual conditions for the two words, so it is phrased "all source networks" and "all source datastores".

mturley (changes requested):

src/app/Mappings/components/MappingBuilder/MappingBuilder.tsx
gildub: Done

src/app/Mappings/components/MappingBuilder/MappingBuilder.tsx
```diff
@@ -33,7 +33,7 @@ export const MappingBuilder: React.FunctionComponent<IMappingBuilderProps> = ({
   setBuilderItems,
 }: IMappingBuilderProps) => {
   const messageSelectBoth = 'You must select a source and target before adding another mapping.';
-  const messageExhausted = `All sources ${mappingType.toLowerCase()} have been mapped.`;
+  const messageExhausted = `All sources ${mappingType.toLowerCase()}s have been mapped.`;
```
mturley, line 36: @gildub this is still going to produce the wrong spelling/phrasing.
1. `mappingType` is either Network or Storage, we need "networks" and *"datastores"*
2. The word "sources" should not be plural, only the type.

We need this:
```ts
  const messageExhausted = `All source ${
    mappingType === MappingType.Network ? 'networks' : 'datastores'
  } have been mapped.`;
```

gildub: Sorry, I shouldn't have been working on it late last night, not paying enough attention.

mturley: No problem at all!

Yeah, it's JS expressions all the way down :)

mturley (approved):
Looks great!

PR #91, opened 2020-09-29
Add useVMwareTreeQuery, use in plan wizard and implement tree search

### LINKED ISSUE #39: Add search state to the tree view in step 2 of the plan wizard
Currently the `onSearch` prop of the `<TreeView>` is not connected to anything. Related to #38, since this will involve filtering the same generated tree items based on the search state before passing them into the TreeView.

Also weirdly, pressing the enter key while focused in the search bar expands/collapses the tree, not sure what's going on there. We'll want to address that too somehow.

### LINKED ISSUE #80: Add API query for tree data in plan wizard VM filter step

mturley:
Closes #80 
Closes #39

This removes the mock data that was formatted as `TreeViewDataItem[]` (the structure PF's TreeView needs) and instead adds mock data formatted the way the tree data is in the API and adds a helper to convert the API structure to the PF structure at render time.

Search in a tree is a bit odd, this search will filter down to tree items which either directly match the search text or contain a descendant that matches the search text. I apply this filter in the same recursive loops used to convert the API data, but at each level of recursion it checks the entire subtree of each child, so in theory the performance will worsen exponentially with larger trees. I don't expect that to be a problem except for very very large trees, though.

I also ran into a weird PF CSS issue, I added some workaround CSS for now and opened https://github.com/patternfly/patternfly/issues/3532 to address it upstream.

This does not add checkbox selection state to the tree, I'll address that in another PR (https://github.com/konveyor/virt-ui/issues/38).


gildub (commented):
That's a bit late, but LGTM!

PR #94, opened 2020-10-01
Self signed cert error handling & reusable http service hook 

### LINKED ISSUE #63: Figure out auth/login and prod cluster deployment configuration details
Since we're getting closer to our deadline, it might be time to start thinking about this. Could definitely use your help, @eriknelson .

### LINKED ISSUE #89: Surface certificate invalid warning in UI
image

Similar to cam, we will need a way for a user to accept expired self-signed SSL certificates.

### LINKED ISSUE #90: Convert run-time .env vars to a virtMeta.json file
Based on the operator changes in https://github.com/konveyor/virt-operator/pull/11, we're going to need to parse and serve the API URLs and OAuth configuration from a JSON file instead of the `.env` file, similarly to MTC.

I think we can keep `.env` and the `Dotenv` webpack plugin for use with `DATA_SOURCE` so it can still tree-shake our mocks at build-time, and we can have it for any other future build variants. We'll just have to also add `virtMeta.json` and ship that to the UI in Express similarly to how we do the `RUNTIME_ENV.*` vars right now.

ibolton336:
Closes #89 
Closes #90 
Closes #63

- Adds a cert-error route which renders an alert when the cert invalid warning arises from a fetch request. 
- Adds a reusable hook ( httpService.ts file)  designed to handle errors & any side effects for all fetch requests across the app. 
- Adds a OAuthContext api to serve as a store for http related values like the oauth token & the url string used to display a link to users for accepting certs.


config/helpers.js
```diff
@@ -0,0 +1,36 @@
...
+    process.exit(1);
+  }
+  const localConfig = require(configPath);
+  const migMeta = {
+    clusterApi: localConfig.clusterApi,
+    oauth: {
+      clientId: localConfig.oauthClientId,
+      redirectUri: localConfig.redirectUri,
+      userScope: localConfig.userScope,
+      clientSecret: startRemoteClientSecret,
```
mturley, line 26: Looks like we can just use `localConfig.oauthClientSecret` here instead of having the `startRemoteClientSecret` variable duplicating that value

config/helpers.js
```diff
@@ -0,0 +1,36 @@
...
+    console.error('ERROR: config/config.dev.json is missing');
+    console.error(
+      'Copy config/config.dev.json.example to config/config.dev.json' +
+        ' and optionally configure your dev settings. A valid clusterUrl is ' +
+        ' required for start:remote.'
+    );
+    process.exit(1);
+  }
+  const localConfig = require(configPath);
+  const migMeta = {
```
mturley, line 20: Let's call it `virtMeta` everywhere instead of `migMeta`

mturley: I actually wonder if we even need to create a meta object here at all. All the values are just copied from the JSON file which we already have parsed as `localConfig`. Why don't we just call the config.dev.json file `virtMeta.dev.json` instead, and have it structured the way we want this object structured (oauth stuff nested), and use it directly (just `return require(configPath)` and then in other places `const virtMeta = getLocalConfig()`).

config/webpack.common.js
```diff
@@ -5,11 +5,12 @@ const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
 const Dotenv = require('dotenv-webpack');
 require('dotenv').config(); // For runtime-env-vars.js
```
mturley, line 6: I think we can totally remove the node `dotenv` which we're using for `_env_encoded`, since the new virtMeta is replacing its purpose. We should `yarn remove dotenv` and get rid of this import (`dotenv-webpack` will still use it internally but we don't need to directly depend on it anymore)

config/webpack.common.js
```diff
@@ -5,11 +5,12 @@ const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
...
 const BG_IMAGES_DIRNAME = 'bgimages';
+const helpers = require('./helpers');
 
 module.exports = (env) => {
   return {
     entry: {
-      app: path.resolve(__dirname, 'src', 'index.tsx'),
+      app: path.resolve(__dirname, '../src/index.tsx'),
```
mturley, line 13: Yeah, moving all these configs into `config/` makes sense, thanks.

config/webpack.common.js
```diff
@@ -126,23 +130,26 @@ module.exports = (env) => {
...
     output: {
       filename: '[name].bundle.js',
-      path: path.resolve(__dirname, 'dist'),
+      path: path.resolve(__dirname, '../dist'),
     },
     plugins: [
       new HtmlWebpackPlugin(
         env === 'development' || process.env.DATA_SOURCE === 'mock'
           ? {
               // In dev and mock-prod modes, populate window._env at build time
```
mturley, line 139: let's change this comment to refer to `window._virt_meta`.

config/webpack.common.js
```diff
@@ -126,23 +130,26 @@ module.exports = (env) => {
...
               filename: 'index.html',
-              template: path.resolve(__dirname, 'src', 'index.html.ejs'),
+              template: path.resolve(__dirname, '../src/index.html.ejs'),
               templateParameters: {
-                _env_encoded: require('./runtime-env-vars'),
+                _env_encoded: require('../runtime-env-vars'),
```
mturley, line 143: Let's remove `_env_encoded` entirely and delete the `runtime-env-vars` script. We should also remove everything from `.env.example` except for `DATA_SOURCE` and change the README so it describes adding those variables to virtMeta instead of .env

config/webpack.common.js
```diff
@@ -126,23 +130,26 @@ module.exports = (env) => {
...
       filename: '[name].bundle.js',
-      path: path.resolve(__dirname, 'dist'),
+      path: path.resolve(__dirname, '../dist'),
     },
     plugins: [
       new HtmlWebpackPlugin(
         env === 'development' || process.env.DATA_SOURCE === 'mock'
           ? {
               // In dev and mock-prod modes, populate window._env at build time
               filename: 'index.html',
```
mturley, line 140: To your earlier point, maybe we can remove this line since index.html is the default. Doesn't really matter though.

config/webpack.dev.js
```diff
@@ -22,7 +22,7 @@ module.exports = merge(common('development'), {
     proxy: [
       {
         // NOTE: Any future backend-only routes added to server.js need to be listed here:
-        context: ['/hello'], // TODO remove this /hello example once we have some real routes here
+        context: ['/login', '/login/callback'],
```
mturley, line 25: 💪

deploy/oAuthHelpers.js
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const { AuthorizationCode } = require('simple-oauth2');
+const fetch = require('node-fetch');
+
+let cachedOAuthMeta = null;
+
+const sanitizeMigMeta = (migMeta) => {
```
mturley, line 7: sanitizeMigMeta -> sanitizeVirtMeta
migMeta -> virtMeta
etc etc, I'll stop calling these out

deploy/server.js
```diff
@@ -0,0 +1,75 @@
...
+    return res.status(500).json('Authentication failed');
+  }
+});
+
+app.get('*', (req, res) => {
+  if (process.env['DATA_SOURCE'] === 'mock') {
+    res.sendFile(path.join(__dirname + '/dist/index.html'));
+  } else {
+    res.render('index.html.ejs', {
+      _env_encoded: require('./runtime-env-vars'),
```
mturley, line 70: In addition to removing _env_encoded here I think you also need to add _virt_meta and inject the contents here from getLocalConfig just like we do in webpack, for this to work in remote-prod mode. You can probably abstract the `Buffer.from()` part of it to another helper too like `getEncodedLocalConfig` or whatever since it's in both places.

scripts/remote-dev.js
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// Helpers
+
+function setupOAuthClient() {
+  const remoteConfig = JSON.parse(fs.readFileSync(remoteConfigFile));
+  const oauthRedirectUri = `http://localhost:${remoteConfig.devServerPort}/login/callback`;
+
+  const oauthClientName = 'mig-ui';
+  const remoteDevSecret = 'bWlncmF0aW9ucy5vcGVuc2hpZnQuaW8K';
```
mturley, line 43: We can just use `remoteConfig.oauth.clientSecret` or whatever here too instead of duplicating it.

scripts/remote-dev.js
```diff
@@ -0,0 +1,126 @@
...
+
+  try {
+    console.log('Checking to see if mig-ui oauthclient exists in cluster...');
```
mturley, line 46: Tiny thing, but this console.log should use the `${oauthClientName}` instead of hard-coding "mig-ui". I know, I know. Ignore this if you want

scripts/remote-dev.js
```diff
@@ -0,0 +1,126 @@
...
+  } catch (error) {
+    // Some error other than the client not existing occurred
+    if (!error.stderr.toString().includes('not found')) {
+      console.error('ERROR: Something went wrong while trying to get the remote-dev oauthclient:');
+      console.error(error.stdout.toString());
+      process.exit(1);
+    }
+  }
+
+  console.log('Attempting to create oauthclient for mig-ui...');
```
mturley, line 60: same here with the hard-coded "mig-ui"

scripts/remote-dev.js
```diff
@@ -0,0 +1,126 @@
...
+    execSync(`echo '${JSON.stringify(oauthClient)}' | oc create -f-`);
+  } catch (error) {
+    console.error('ERROR: Something went wrong trying to create a new OAuthClient:');
+    console.error(error.stdout.toString());
+    process.exit(1);
+  }
+}
+
+function setupCors() {
+  if (!process.env.ORIGIN3_HOST && process.env.DEPRECATED_CORS === 'true') {
```
mturley, line 85: What's the deal with this `DEPRECATED_CORS` stuff? @eriknelson do you know if that variable is needed / if this stuff is really deprecated?

mturley: Also I think we can remove `ORIGIN3_HOST` since this will always be an OpenShift 4 host right?

scripts/run-local-express.sh
```diff
@@ -1,8 +1,14 @@
 #!/bin/bash
 _dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+export MIGMETA_FILE="$_dir/../tmp/migmeta.json"
+# export VIEWS_DIR="$_dir/../public"
```
mturley, line 4: Pretty sure we don't need `VIEWS_DIR` since the `index.html.ejs` just gets dumped into `dist` with the rest of the bundle. Maybe that's a mistake, not sure. If you want we could add this back and update webpack to put it there, and express to find it there.

mturley (changes requested):
Some comments based on our discussion on slack.

Also it looks like you added a package-lock.json file, you must have used an `npm` command instead of `yarn`, let's get rid of that.

tmp/migmeta.json
```diff
@@ -0,0 +1 @@
+{"clusterApi":"https://api.cluster-ibb6039ocp4b.ibb6039ocp4b.mg.dog8code.com:6443","oauth":{"clientId":"mig-ui","redirectUri":"http://localhost:9000/login/callback","userScope":"user:full","clientSecret":"bWlncmF0aW9ucy5vcGVuc2hpZnQuaW8K"},"namespace":"openshift-migration","configNamespace":"openshift-config","inventoryApi":"https://inventory-openshift-migration.apps.cluster-ibb6039ocp4b.ibb6039ocp4b.mg.dog8code.com"}
```
mturley, line 1: Looks like we don't want this committed, right? does webpack generate it in both dev and prod modes? Should probably be in `.gitignore`

mturley (dismissed):
Some more comments for ya

mturley:
Lots of little comments, but this is awesome. Nice work

src/app/common/LoginHandlerComponent.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { useEffect } from 'react';
+import { useLocation, Redirect, useHistory } from 'react-router-dom';
+import { connect } from 'react-redux';
```
mturley, line 3: What's redux doing in here? shoo!

mturley:
When this is done it looks like it will also close #90

mturley:
#63 too!

config/helpers.js
ibolton336: done

config/helpers.js
ibolton336: done

config/webpack.common.js
ibolton336: I'll leave it since it provides a bit more clarity as to what webpack is doing.

src/app/queries/httpService.ts
```diff
@@ -0,0 +1,47 @@
...
+interface HttpResponse<T> extends Response {
+  parsedBody?: T;
+}
+
+export const isSelfSignedCertError = <T>(response: HttpResponse<T>): boolean => {
+  // HACK: Doing our best to determine whether or not the
+  // error was produced due to a self signed cert error.
+  // It's an extremely barren object.
+  const certErrorText = 'Failed to fetch';
+  return response instanceof TypeError && response.message === certErrorText;
```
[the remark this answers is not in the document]
ibolton336: I have been looking everywhere and there is no better way to single this error our. There is no code associated with it as far as I can tell. Maybe we can open an issue and circle back to see if this changes as we start doing CRUD requests & passing more headers along.

src/app/common/context/OAuthContext.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from 'react';
+import { LocalStorageKey, useLocalStorageContext } from './LocalStorageContext';
+
+interface IOAuthContext {
+  setFailedUrl: (url: string) => void;
```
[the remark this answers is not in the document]
ibolton336: Maybe i can just change the name of this context to encapsulate http/fetch/oauth related stuff

ibolton336:
@mturley  I think i hit everything for the most part. Still need to confirm query is handling the errors it needs to

mturley (approved):
LGTM!

PR #95, opened 2020-10-01
Move PlanWizard to its own page

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/54

### DIFF
--- a/src/app/Plans/PlansPage.tsx
+++ b/src/app/Plans/PlansPage.tsx
@@ -31,6 +31,7 @@ const PlansPage: React.FunctionComponent = () => {
   const providersQuery = useProvidersQuery();
 
   const vmwareProviders = providersQuery.data?.vsphere || [];
+  const vmwareProviders = []; //providersQuery.data?.vsphere || [];
   const openshiftProviders = providersQuery.data?.openshift || [];
 
   let addPlanDisabledObj: Pick<IAddTooltipProps, 'isTooltipEnabled' | 'content'> = {
@@ -73,11 +74,21 @@ const PlansPage: React.FunctionComponent = () => {
                     content={addPlanDisabledObj.content}
                   >
                     <div className={`${spacing.mtMd}`}>
-                      <Link to="/plans/create">
+                      {/* TODO: Fix with a wrapper */}
+                      {addPlanDisabledObj.isTooltipEnabled ? (
                         <Button isDisabled={addPlanDisabledObj.isTooltipEnabled} variant="primary">
                           Create migration plan
                         </Button>
-                      </Link>
+                      ) : (
+                        <Link to="/plans/create">
+                          <Button
+                            isDisabled={addPlanDisabledObj.isTooltipEnabled}
+                            variant="primary"
+                          >
+                            Create migration plan
+                          </Button>
+                        </Link>
+                      )}
                     </div>
                   </AddTooltip>
                 </EmptyState>
--- a/src/app/Plans/PlansPage.tsx
+++ b/src/app/Plans/PlansPage.tsx
@@ -10,7 +10,7 @@ import {
   Button,
   Alert,
 } from '@patternfly/react-core';
-import { Link } from 'react-router-dom';
+import { Link, useHistory } from 'react-router-dom';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import { PlusCircleIcon } from '@patternfly/react-icons';
 
@@ -28,9 +28,11 @@ const plans = MOCK_PLANS;
 const migrations = MOCK_MIGRATIONS;
 
 const PlansPage: React.FunctionComponent = () => {
+  const history = useHistory();
   const providersQuery = useProvidersQuery();
 
-  const vmwareProviders = providersQuery.data?.vsphere || [];
+  // const vmwareProviders = providersQuery.data?.vsphere || [];
+  const vmwareProviders = [];
   const openshiftProviders = providersQuery.data?.openshift || [];
 
   let addPlanDisabledObj: Pick<IAddTooltipProps, 'isTooltipEnabled' | 'content'> = {
@@ -73,21 +75,13 @@ const PlansPage: React.FunctionComponent = () => {
                     content={addPlanDisabledObj.content}
                   >
                     <div className={`${spacing.mtMd}`}>
-                      {/* TODO: Fix with a wrapper */}
-                      {addPlanDisabledObj.isTooltipEnabled ? (
-                        <Button isDisabled={addPlanDisabledObj.isTooltipEnabled} variant="primary">
-                          Create migration plan
-                        </Button>
-                      ) : (
-                        <Link to="/plans/create">
-                          <Button
-                            isDisabled={addPlanDisabledObj.isTooltipEnabled}
-                            variant="primary"
-                          >
-                            Create migration plan
-                          </Button>
-                        </Link>
-                      )}
+                      <Button
+                        onClick={() => history.push('/plans/create')}
+                        isDisabled={addPlanDisabledObj.isTooltipEnabled}
+                        variant="primary"
+                      >
+                        Create migration plan
+                      </Button>
                     </div>
                   </AddTooltip>
                 </EmptyState>
--- a/src/app/Plans/PlansPage.tsx
+++ b/src/app/Plans/PlansPage.tsx
@@ -10,6 +10,7 @@ import {
   Button,
   Alert,
 } from '@patternfly/react-core';
+import { Link } from 'react-router-dom';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import { PlusCircleIcon } from '@patternfly/react-icons';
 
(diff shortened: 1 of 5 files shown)


src/app/Plans/PlansPage.tsx
```diff
@@ -75,13 +73,11 @@ const PlansPage: React.FunctionComponent = () => {
...
                   >
                     <div className={`${spacing.mtMd}`}>
-                      <Button
-                        isDisabled={addPlanDisabledObj.isTooltipEnabled}
-                        onClick={toggleWizard}
-                        variant="primary"
-                      >
-                        Create migration plan
-                      </Button>
+                      <Link to="/planwizard">
```
mturley, line 76: Let's see if we can keep the URL paths roughly based on the breadcrumb bar, so this should be a route under `/plans`. Maybe `/plans/create`?

mturley: Also does this link properly disable with the button is disabled? Not sure how that works when it's wrapped around it like this. If not we may need to either only render the <Link> around it when the button would be enabled, or handle the link behavior as an onClick instead.

src/app/Plans/components/PlansTable.tsx
```diff
@@ -172,9 +169,9 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
     <>
       <Level>
         <LevelItem>
-          <Button variant="secondary" onClick={toggleWizard} isDisabled={false}>
-            Create
-          </Button>
+          <Link to="/planwizard">
```
mturley, line 172: See above re: maybe using `/plans/create` instead of `/planwizard`. Thoughts on that @ibolton336 ?

src/app/Plans/components/Wizard/PlanWizard.tsx
```diff
@@ -121,16 +114,23 @@ const PlanWizard: React.FunctionComponent<IPlanWizardProps> = ({
   };
 
   return (
-    <Modal isOpen width="95%" showClose={false} hasNoBodyWrapper>
+    <PageSection variant="light">
+      <Breadcrumb className={`${spacing.mbLg} ${spacing.prLg}`}>
+        <BreadcrumbItem>
+          <Link to={`/plans`}>Migration plans</Link>
+        </BreadcrumbItem>
+        <BreadcrumbItem>Plan Wizard</BreadcrumbItem>
```
mturley, line 122: The word "wizard" shouldn't really be shown in the UI as the name of the page, right @vconzola ? I imagine this breadcrumb bar should read "Migration plans > Create" ?

src/app/Plans/components/Wizard/PlanWizard.tsx
```diff
@@ -121,16 +114,23 @@ const PlanWizard: React.FunctionComponent<IPlanWizardProps> = ({
...
+      </Breadcrumb>
       <Wizard
         title="Create a Migration Plan"
         steps={steps}
-        onClose={onClose}
         onNext={onMove}
         onBack={onMove}
         onSubmit={(event) => event.preventDefault()}
+        hideClose={true}
+        cancelButtonText=""
```
mturley, line 131: This hides the Cancel button. @vconzola do we still want the cancel button in place? It could just bring you back to the plans page.

src/app/Plans/components/Wizard/PlanWizard.tsx
```diff
@@ -121,16 +114,23 @@ const PlanWizard: React.FunctionComponent<IPlanWizardProps> = ({
...
-    <Modal isOpen width="95%" showClose={false} hasNoBodyWrapper>
+    <PageSection variant="light">
+      <Breadcrumb className={`${spacing.mbLg} ${spacing.prLg}`}>
+        <BreadcrumbItem>
+          <Link to={`/plans`}>Migration plans</Link>
+        </BreadcrumbItem>
+        <BreadcrumbItem>Plan Wizard</BreadcrumbItem>
+      </Breadcrumb>
       <Wizard
         title="Create a Migration Plan"
```
mturley, line 125: Let's remove the `title` prop and instead add a title to the page section above, by the breadcrumb bar like we have on our other pages. The black title bar looks a little awkward embedded on the page.

src/app/routes.tsx
```diff
@@ -62,6 +63,12 @@ export const routes: AppRouteConfig[] = [
     path: '/plans',
     title: `${APP_TITLE} | Migration Plans`,
   },
+  {
+    component: PlanWizard,
+    exact: true,
+    path: '/planwizard',
```
mturley, line 69: See above re: the route path

mturley (changes requested):
Looking good! A couple of little things.

src/app/routes.tsx
mturley: One thing to be aware of is if we make this `/plans/create` I believe it needs to be higher in the list than `/plans/:planId` in order to prevent matching that route instead, so let's stay aware of that when this change gets rebased on/under #78.

mturley:
@vconzola what do you think about putting the wizard in a Card? if we move the title up into the top page section above the breadcrumb bar, we could give the lower section the grey background like our other pages have, with the wizard in a card on top of that, instead of having the sections both have white backgrounds and blend together.

src/app/Plans/components/Wizard/PlanWizard.tsx
gildub: Looks better, I've changed and aligned it to go along with above comments.

@vconzola to confirm it or use something else.

src/app/Plans/components/Wizard/PlanWizard.tsx
gildub: I've restored it to help @vconzola review it.

src/app/Plans/PlansPage.tsx
gildub: That's a good point. I'll done a quick unary fix before adding a wrapper for long run.

src/app/routes.tsx
gildub: I've added an TODO comment too to help.

src/app/Plans/PlansPage.tsx
```diff
@@ -75,13 +73,11 @@ const PlansPage: React.FunctionComponent = () => {
...
                     <div className={`${spacing.mtMd}`}>
-                      <Button
-                        isDisabled={addPlanDisabledObj.isTooltipEnabled}
-                        onClick={toggleWizard}
-                        variant="primary"
-                      >
-                        Create migration plan
-                      </Button>
+                      <Link to="/plans/create">
+                        <Button isDisabled={addPlanDisabledObj.isTooltipEnabled} variant="primary">
```
mturley, line 79: @gildub I tested rendering this button with `isDisabled={true}` and the Link still takes you to the wizard page. Also, `<Link>` renders an `<a>` and having a `<button>` inside an `<a>` is not semantically allowed in HTML (even though it renders, we're probably getting a warning about it).

Instead we should probably just use the button and navigate in its `onClick` handler. You can use the `useHistory` hook from `react-router-dom` and do `useHistory().push('/plans/create')`. https://reactrouter.com/web/api/Hooks/usehistory

gildub: Oh I see re having `<button>` embedded in a `<a>`. I was actually in unknown territory for me when I did that this morning. 
I guess I pick up a wrong approach. Good to know, thanks.

Let me use the `useHistory`.

src/app/Plans/PlansPage.tsx
```diff
@@ -10,7 +10,7 @@ import {
   Button,
   Alert,
 } from '@patternfly/react-core';
-import { Link } from 'react-router-dom';
+import { Link, useHistory } from 'react-router-dom';
```
mturley, line 13: This Link import is unused now, let's remove it

src/app/Plans/PlansPage.tsx
```diff
@@ -28,9 +28,11 @@ const plans = MOCK_PLANS;
 const migrations = MOCK_MIGRATIONS;
 
 const PlansPage: React.FunctionComponent = () => {
+  const history = useHistory();
   const providersQuery = useProvidersQuery();
 
-  const vmwareProviders = providersQuery.data?.vsphere || [];
+  // const vmwareProviders = providersQuery.data?.vsphere || [];
```
mturley, line 34: Can you revert this change and go back to using the real data for `vmwareProviders`?

src/app/queries/mocks/plans.mock.ts
```diff
@@ -231,7 +231,7 @@ const plan4: IPlan = {
   },
 };
 
-export const MOCK_PLANS: IPlan[] = [plan1, plan2, plan3, plan4];
+export const MOCK_PLANS: IPlan[] = []; //[plan1, plan2, plan3, plan4];
```
mturley, line 234: Same here, I assume this was to test the empty state, can we revert it?

mturley (changes requested):
Last thing!

src/app/Plans/PlansPage.tsx
gildub: Of course!

src/app/Plans/PlansPage.tsx
gildub: Darn me!

src/app/queries/mocks/plans.mock.ts
gildub: Same!

mturley (approved):
Excellent!

PR #97, opened 2020-10-01
Toggle menu after selection only in mobileView

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/92


mturley (approved):
Nice enhancement! I like it.

PR #99, opened 2020-10-02
Disable Kebab unless conditions are met

gildub:
Closes https://github.com/konveyor/virt-ui/issues/98

Also added a commit to make "Cancel" button of `VMMigrationDetails` to be enabled only when VM migration is running.


src/app/Plans/components/PlanActionsDropdown.tsx
```diff
@@ -11,6 +25,7 @@ const PlansActionsDropdown: React.FunctionComponent = () => {
       isPlain
       dropdownItems={[
         <DropdownItem
+          isDisabled={hasNotCondition(PlanStatusConditionsType.Ready)}
```
mturley, line 28: Are we sure that it will lose the ready condition once it's in-progress? If it's errored, does it consider itself ready for a retry? not sure how this data looks

gildub: According to @jortel the conditions has one entry "Ready" if it's ready otherwise it shouldn't be present.

gildub: Good question, @jortel could help us out here please?

gildub: @mturley , sorry, I was wrong. There will be an explicit condition "Execute".

gildub:
@mturley and @vconzola, please review.

mturley (approved):
LGTM 👍

Issue #81, opened 2020-09-27
Add API query for VMs filtered by tree selections


mturley:
I'm not sure what this filter logic will look like, or whether we'll need to change any backend stuff to support it.

mturley:
Note to self: the tree query results contain VMs as leaf nodes, we can just perform lookups based on those ids.

Issue #60, opened 2020-09-21
Add useFormState validation to all other forms


mturley:
In #53 we only added validation using useFormState to the Add Provider form. We should put it in all the other forms we have so far (mappings, host network selection, plan wizard steps, ???)

Following is exhaustive list of forms, checked one are using useFormState:
- [x] `AddProviderModals` - Providers 
- [x] `PlanWizard` - Plan migration wizard
- [ ] `SelectNetworkModal` - Host Network  
- [ ] `AddEditMappingModal` - for Network and Storage mappings
- [ ] `MappingForm` - for Network and Storage mappings from Wizard Plan

mturley:
Also, the validation criteria for the Add Provider form were pretty arbitrary and strict, we need to figure out the actual character limits etc to enforce there.

gildub:
The form validation left to be done are:
[ ] `SelectNetworkModal` - Host Network 
[ ] `AddEditMappingModal` - for Network and Storage mappings

mturley:
Related note: the form state in the MappingForm (storage and network mapping steps of the plan wizard) is incomplete, but that will be finished as part of #103

Issue #50, opened 2020-09-16
Fix URL label in Provider creation form 


gildub:
The label after "URL" needs to be "Service account token".

![cnv-provider-label](https://user-images.githubusercontent.com/1901741/93326887-d3d45400-f819-11ea-9fc1-4ca5d211b3ed.png)

PR #100, opened 2020-10-02
Add useFormState to plan wizard, add tree conversion and selection logic, fix some layout, TS and a11y stuff

### LINKED ISSUE #38: Add checkbox selection state to tree view in step 2 of the plan wizard.
This has strange recursion implications because of the need to represent a subtree's selection state based on the selection of its descendants. We can reference the implementation in [this example on patternfly.org](https://www.patternfly.org/v4/documentation/react/components/treeview#with-checkboxes), possibly modified to work with our `useSelectionState` hook, but we may need to modify that hook to support selecting multiple items (a subtree) at once synchronously.

In order to do this, we should also define our TreeViewDataItems declaratively as a function of API data and selection state, instead of mutating the tree items directly when selection state changes. This may require traversing the tree structure more than once.. I started brainstorming that here: https://github.com/konveyor/virt-ui/pull/21#pullrequestreview-484300687

### LINKED ISSUE #71: Add useFormState to plan wizard, figure out how lifting that state will work
I was thinking we'd need to use Context to lift the form state for wizard steps, but that complicates TS types (the consumer can't tell what form you're passing in since you could be using it for any provider, so we'd need generics and explicit form field interfaces there). I think instead we can just pass the form state objects directly into the wizard steps.

### LINKED ISSUE #88: PF Select ignores aria-label propagation from SimpleSelect, generating `serious` error.
It seems like Select triggered from SimpleSelect is simply ignoring aria-label which is propagated via ...props.
I tried to force by adding aria-label explicitly to `Select` component with the same result

```
serious: ARIA input fields must have an accessible name https://dequeuniversity.com/rules/axe/3.5/aria-input-field-name?application=axeAPI index.js:216
Element:  
<ul id="target-provider" class="pf-c-select__menu" role="listbox">
index.js:114
HTML: <ul class="pf-c-select__menu" role="listbox" id="target-provider"> index.js:122
Fix any of the following:
  aria-label attribute does not exist or is empty
  aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
  Element has no title attribute or the title attribute is empty react_devtools_backend.js:2273:24
```
​
To reproduce:
Use any component calling `SimpleSelect` and make sure aria-label prop is define in the SimpleSelect call. 
For example use `AddEditMappingModal` and then in debug mode (dev tools) click on any select field, the error should appear in console.

mturley:
Resolves #71 
Resolves #38
Resolves #88
It takes care of most of #60 but I'll leave that open for the hosts network selection modal.

This is a big one, it got away from me a bit.
* Adds a `usePlanWizardFormState` hook internal to the `PlanWizard` file, not to be reused (since that will create another copy of all the state), but lifted out of the component so that I could grab its inferred types using `ReturnType<>`. This hook returns an object containing form objects (`useFormState` calls) for each wizard step.
* Passes the corresponding form object down to each wizard step and uses each one's `isValid` state to block the Next button.
* Incrementally enables jumping between wizard steps as you visit them for the first time.
* Fixes the Cancel button of the wizard which was a noop.
* Enhances useFormState so it can handle `yup.ArraySchema` conditionally when array values are passed.
* Uses real API data for the Filter VMs tree, with helper functions to recursively convert from `VMwareTree` objects to `TreeViewDataItem`s.
* Adds checkbox selection state to the Filter VMs tree, with some more recursive magic (selecting an item will select all its descendants).
* Implements the Review step of the wizard, showing selections from the form state.
* Adds an `isDirty` property to `useFormState` at the field and form levels, to detect if the values differ from the initial values.
* Adds a prompt preventing the user from leaving the wizard if any wizard form state is dirty.
* Fixes the aria-label issue with Select components
* Cleans up some TS stuff, layout / style stuff and some of our react-axe warnings.

Note:
* This does not implement #81, we still need to use the tree view selections to filter the VMs table and use real VMs there.
* Creating a new mapping in the plan wizard doesn't currently save it into form state, so that feature is blocked for now. Because of #103, I didn't want to have to build that logic twice, so we'll get that state wired up as part of that issue. To proceed through the wizard for now, you have to select an existing mapping.


src/app/common/components/SimpleSelect.tsx
```diff
@@ -19,6 +19,7 @@ export interface ISimpleSelectProps
     SelectProps,
     'onChange' | 'isExpanded' | 'onToggle' | 'onSelect' | 'selections' | 'value'
   > {
+  'aria-label': string; // Make required to satisfy react-axe
```
gildub, line 22: React-axe is enforcing it but shouldn't `aria-label` be mandatory when using user facing fields?

gildub:
@mturley, LGTM! Just a question/comment.

It's nice to see the form validation in full action, very neat!

gildub:
@mturley, in the description I believe you meant "It takes care of most of #60..." instead of #50.

mturley:
Ah, yep I did, thanks @gildub

src/app/common/components/SimpleSelect.tsx
mturley: Yeah, I only commented to be clear that we're adding it for that reason since it's not clear why PatternFly excludes it. I'm not sure what the deal is with that, I think maybe it's because you can also use `aria-labelledby` technically. I can just remove the comment.

mturley:
Addressed @gildub's comment, rebased.

src/app/Plans/components/Wizard/GeneralForm.tsx
```diff
@@ -1,77 +1,82 @@
...
+        helperTextInvalid={form.fields.sourceProvider.error}
+        validated={form.fields.sourceProvider.isValid ? 'default' : 'error'}
       >
         <SimpleSelect
-          id="sourceProvider"
+          id="source-provider"
+          aria-label="Source provider"
           options={sourceProvidersOptions}
-          value={[sourceProvidersOptions.find((option) => option.value === sourceProvider)]}
+          value={[
```
ibolton336, line 71: What do we do if there is no available provider to select?
image

ibolton336 (changes requested):

src/app/Plans/components/Wizard/GeneralForm.tsx
mturley: Hmm, that's a bug I was seeing before that I thought I fixed in this PR. Looking now.

mturley: Pushed changes to prevent opening the wizard if there are insufficient providers.

ibolton336 (approved):

ibolton336:
LGTM! Nice job :100:

PR #101, opened 2020-10-03
Pipeline summary

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/59

### DIFF
--- a/src/app/Plans/components/VMMigrationDetails.tsx
+++ b/src/app/Plans/components/VMMigrationDetails.tsx
@@ -80,7 +80,6 @@ const VMMigrationDetails: React.FunctionComponent = () => {
     {
       title: 'Status',
       transforms: [sortable],
-      columnTransforms: [classNamesTransform(alignment.textAlignCenter)],
     },
     { title: '', columnTransforms: [classNamesTransform(alignment.textAlignRight)] },
   ];
--- a/src/app/Plans/components/VMMigrationDetails.tsx
+++ b/src/app/Plans/components/VMMigrationDetails.tsx
@@ -6,9 +6,9 @@ import {
   Button,
   Card,
   CardBody,
+  Flex,
   Pagination,
   PageSection,
-  Text,
   Title,
 } from '@patternfly/react-core';
 import {
@@ -77,8 +77,12 @@ const VMMigrationDetails: React.FunctionComponent = () => {
     { title: 'Start time', transforms: [sortable] },
     { title: 'End time', transforms: [sortable] },
     { title: 'Data copied', transforms: [sortable] },
-    { title: 'Status', transforms: [sortable] },
-    { title: '' },
+    {
+      title: 'Status',
+      transforms: [sortable],
+      columnTransforms: [classNamesTransform(alignment.textAlignCenter)],
+    },
+    { title: '', columnTransforms: [classNamesTransform(alignment.textAlignRight)] },
   ];
 
   const rows: IRow[] = [];
@@ -106,7 +110,11 @@ const VMMigrationDetails: React.FunctionComponent = () => {
           migration.other.total / 1024
         )} GB`,
         {
-          title: <PipelineSummary status={migration.status2} />,
+          title: (
+            <Flex direction={{ default: 'column' }} alignItems={{ default: 'alignItemsCenter' }}>
+              <PipelineSummary status={migration.status2} />
+            </Flex>
+          ),
         },
 
         {
--- a/src/app/Plans/components/VMMigrationDetails.tsx
+++ b/src/app/Plans/components/VMMigrationDetails.tsx
@@ -105,7 +105,7 @@ const VMMigrationDetails: React.FunctionComponent = () => {
           migration.other.total / 1024
         )} GB`,
         {
-          title: <PipelineSummary total={0} current={0} status={migration.other.status} />,
+          title: <PipelineSummary status={migration.status2} />,
         },
 
         {
--- a/src/app/Plans/components/VMMigrationDetails.tsx
+++ b/src/app/Plans/components/VMMigrationDetails.tsx
@@ -21,6 +21,7 @@ import {
   IRow,
   wrappable,
   expandable,
+  cellWidth,
 } from '@patternfly/react-table';
 import { Link } from 'react-router-dom';
 import { useSelectionState } from '@konveyor/lib-ui';
@@ -79,7 +80,7 @@ const VMMigrationDetails: React.FunctionComponent = () => {
     { title: 'Data copied', transforms: [sortable] },
     {
       title: 'Status',
-      transforms: [sortable],
+      transforms: [sortable, cellWidth(10)],
     },
     { title: '', columnTransforms: [classNamesTransform(alignment.textAlignRight)] },
   ];
@@ -110,7 +111,7 @@ const VMMigrationDetails: React.FunctionComponent = () => {
         )} GB`,
         {
           title: (
-            <Flex direction={{ default: 'column' }} alignItems={{ default: 'alignItemsCenter' }}>
+            <Flex direction={{ default: 'column' }}>
               <PipelineSummary status={migration.status2} />
             </Flex>
           ),
--- a/src/app/Plans/components/VMMigrationDetails.tsx
+++ b/src/app/Plans/components/VMMigrationDetails.tsx
@@ -8,6 +8,7 @@ import {
   CardBody,
   Pagination,
   PageSection,
+  Text,
   Title,
 } from '@patternfly/react-core';
 import {
--- a/src/app/Plans/components/VMMigrationDetails.tsx
+++ b/src/app/Plans/components/VMMigrationDetails.tsx
@@ -6,7 +6,6 @@ import {
   Button,
   Card,
   CardBody,
-  Flex,
   Pagination,
   PageSection,
   Title,
@@ -110,11 +109,7 @@ const VMMigrationDetails: React.FunctionComponent = () => {
           migration.other.total / 1024
         )} GB`,
         {
-          title: (
-            <Flex direction={{ default: 'column' }}>
-              <PipelineSummary status={migration.status2} />
-            </Flex>
-          ),
+          title: <PipelineSummary status={migration.status2} />,
         },
 
         {
--- a/src/app/common/components/PipelineSummary.css
+++ b/src/app/common/components/PipelineSummary.css
@@ -1,3 +1,3 @@
-.pipelineSummary {
+.pipelineDash {
   background: linear-gradient(to bottom, white calc(50% - 1px), rgb(41, 40, 40) calc(50% - 1px) calc(50% + 1px), white calc(50% + 1px));
 }
--- a/src/app/common/components/PipelineSummary.css
+++ b/src/app/common/components/PipelineSummary.css
@@ -1,3 +1,12 @@
-.pipelineDash {
-  background: linear-gradient(to bottom, white calc(50% - 1px), rgb(41, 40, 40) calc(50% - 1px) calc(50% + 1px), white calc(50% + 1px));
+.dash {
+  height: 2px;
+  width: 10px;
+}
+
+.dashReached {
+  background-color: rgb(56, 139, 56);
+}
+
+.dashNotReached {
+  background-color: rgb(210, 210, 210);
 }
--- a/src/app/common/components/PipelineSummary.css
+++ b/src/app/common/components/PipelineSummary.css
@@ -1,6 +1,7 @@
 .dash {
   height: 2px;
   width: 10px;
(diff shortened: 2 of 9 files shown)


src/app/Plans/components/VMMigrationDetails.tsx
```diff
@@ -105,7 +106,7 @@ const VMMigrationDetails: React.FunctionComponent = () => {
           migration.other.total / 1024
         )} GB`,
         {
-          title: <PipelineSummary total={0} current={0} status={migration.other.status} />,
+          title: <PipelineSummary status={migration.status2} />,
```
ibolton336, line 109: @gildub Nice progress so far. Is this a typo? "status2"?

src/app/common/components/PipelineSummary.tsx
```diff
@@ -1,19 +1,63 @@
...
 }
 
+const Dash = MinusIcon;
+
+const pipe = (Face, times) => {
+  return times === 0 ? null : (
+    <>
+      <Face />
+      {times > 1 ? <Dash /> : null}
+      {pipe(Face, times - 1)}
```
ibolton336, line 23: Nice!

gildub: Thanks

src/app/Plans/components/VMMigrationDetails.tsx
gildub: That's because we didn't have clear visibility about the entire migration data. So in the meantime it's mocked.
Basically status2 is just a placeholder until that's resolved.

gildub:
@vconzola, this is ready for initial review.

vconzola:
@gildub Looking good so far. With only 4 steps max, it's probably not necessary for us to show "Step x of y". Do we know yet, what the step names are? It would be more useful to show those above the pipeline than the step numbers if there's space. It kills me that we lose so much space in the rows for the button.

gildub:
@vconzola, the 4 maximum know steps, could be less depending on user choices steps, are:
* pre-hook
* disk transfer
* import
* post-hook

Just to confirm, you're suggesting to display the current active step (one of above) when migration is running, is that correct?
If yes that would definitely fit.
Although this raises the question of what do we display when the migration has not started yet or is finished?

mturley (commented):
My one piece of minor feedback here would be that we should either have the dashes between the icons be the same grey as the steps not yet reached, or be black only if they lead to a non-grey icon

gildub:
@mturley, I thought about that too but couldn't see at the time how to change the linear-gradient for the background.
I could just have two separate CSS classes to handle the case which I have revisited using a simple `Div` along with `Flex`.
Aso instead of black/dark color for reached I used a similar green as the full icon.

gildub:
@vconzola, I've pushed the Cancel button to the right to free space. What do you think?

mturley:
Fair enough @gildub I think it looks good and I'm nitpicking too much, we can revisit the colors later if we want. If Vince is happy I'm happy.

vconzola:
Hi @gildub Does moving the Cancel button allow us to put more text above the pipeline? Is there room for the text below? If necessary you can reduce text size to 14 point.

Not started
Running pre-hook
Transferring data
Importing data  <- What is this step doing? What are we importing?
Running post-hook
Complete

Regarding color, I agree with @mturley  that the dashes between the icons should all be the same color gray, if you can figure out how to do it.

mturley:
These colors are looking really nice!
Screen Shot 2020-10-08 at 10 49 55 AM

gildub:
@vconzola, There should be enough space (and if needed reducing text size) for using any of those steps descriptions.

Regarding the dashes color, previous discussion leads us to use light gray for not reached steps and green for reached ones.

src/app/common/components/PipelineSummary.css
```diff
@@ -0,0 +1,12 @@
+.dash {
+  height: 2px;
+  width: 10px;
+}
+
+.dashReached {
+  background-color: rgb(56, 139, 56);
```
mturley, line 7: Actually we should probably use PatternFly CSS variables for these colors so we don't have to hard-code the matching rgb values: https://www.patternfly.org/v4/guidelines/colors

I think maybe for `dashReached` it would be `background-color: var(--pf-global--success-color--100)` and for `dashNotReached` it would be `background-color: var(--pf-global--disabled-color--200)`

gildub: I was wondering about that too. Great I'll use PF colors. Thanks

gildub: That way I don't have to do an exact color pick! ;)

src/app/common/components/PipelineSummary.tsx
```diff
@@ -1,19 +1,93 @@
...
+
 const PipelineSummary: React.FunctionComponent<IPipelineSummaryProps> = ({
-  total,
-  current,
   status,
 }: IPipelineSummaryProps) => {
-  return <Text>{status}</Text>;
+  let sum: React.ReactNode;
+
+  const chain = (Face, times, color) => {
```
mturley, line 36: It doesn't matter much, but technically a function which takes params and returns JSX should be formatted as a component. So this could be `<Chain>` with Face, times and color as props. Either way, the args/props should have TS types.

I see that it's recursive.. I'm actually not sure if components are allowed to render recursively? I think so.

gildub: @mturley, it should be definitely be `Chain`.

Well not sure about that rendering part either. Is Chain rendered every time? I'm understanding it as building a JSX.Element not rendering it, but maybe that's what it means.

gildub: Actually Chain is a `React.ReactNode`.

gildub: I've changed Chain to be a `JSX.Element`.

gildub:
@vconzola, @mturley, please review last update.

src/app/common/components/PipelineSummary.tsx
gildub: @mturley, Well spotted, thank you!

mturley: I think `React.ReactNode` was fine, I actually wasn't aware of the JSX.Element type. Not sure the difference there.

src/app/common/components/PipelineSummary.tsx
```diff
@@ -1,19 +1,101 @@
...
 const PipelineSummary: React.FunctionComponent<IPipelineSummaryProps> = ({
-  total,
-  current,
   status,
 }: IPipelineSummaryProps) => {
-  return <Text>{status}</Text>;
+  let title: string;
+  let summary: JSX.Element;
+
+  const Chain = (Face, times, color) => {
```
mturley, line 38: I was thinking instead of a plain function this should be a `React.FunctionComponent<IChainProps>` and Face, times, color could be moved to an `IChainProps` interface and given types. Then below instead of `        `{Chain(Face, times - 1, color)}` you'd use `<Chain Face={Face} times={times - 1} color={color} />`. Same for usages below for the `full` and `empty` variables.

The distinction is a bit pedantic, but these definitely follow the component model (a function of props that returns JSX) so I think they fit better in that syntax.

src/app/common/components/PipelineSummary.tsx
```diff
@@ -1,19 +1,101 @@
...
+import { MigrationVMStepsType } from '@app/common/constants';
 
 interface IPipelineSummaryProps {
-  total: number;
-  current: number;
-  status: string;
+  status: IVMStatus;
 }
 
+const Dash = (isReached: boolean): JSX.Element => {
```
mturley, line 24: Same for Dash, I think it belongs as a `React.FunctionComponent`.

src/app/common/components/PipelineSummary.tsx
mturley: > I'm understanding it as building a JSX.Element not rendering it, but maybe that's what it means.

As far as I understand, I think that is what it means. It's doing essentially the same thing any of our other small components are doing, the only weird part is how it is recursive, but I think that will still work fine (?)

mturley:
@ibolton336 I'm curious if you have an opinion about the plain-function vs component question here for `Dash` and `Chain`.

src/app/common/components/PipelineSummary.tsx
gildub: Sure but what about the rendering?

vconzola:
@gildub At my monitor's resolution the pipeline column is breaking in the middle. I know @mturley was able to prevent this in another part of the UI by setting a min width on a specific table column. Can you do the same here? 
Screen Shot 2020-10-08 at 12 09 08 PM

mturley:
@gildub I think you can prevent this by using `flexWrap={{ default: 'nowrap' }}` on the Flex parent

mturley:
and we may also want to put `white-space: nowrap;` CSS on the text above it

gildub:
@vconzola, resizing display works now. Sorry I didn't (re)check. Thanks

src/app/common/components/PipelineSummary.tsx
mturley: I'm not sure what you mean? You should just be able to embed the components `<LikeThis>` instead of `{likeThis()}` and it should render fine

gildub: @mturley, oh I meant if there is extra rendering depending on the type (JSX vs React Elements).

mturley: That's what I guess I'm not understanding, I don't see the real difference between `JSX.Element` and `React.ReactNode`. Apparently they are mostly interchangeable (`ReactNode` is [apparently a little more inclusive of things like arrays](https://github.com/typescript-cheatsheets/react/blob/cc2ec67e0b36b7e838df8f2a54b4cd84df6c78cd/README.md#useful-react-prop-type-examples)). So it seems like these should all just be components, and therefore return `React.ReactNode`.

What extra rendering are you referring to?

vconzola:
@gildub Looks great! Thanks.

fdupont-redhat:
Looks really awesome. I love this design :smiley:

src/app/common/components/PipelineSummary.tsx
gildub: React.FC gives a functional programming approach, isn't?

src/app/common/components/PipelineSummary.tsx
gildub: Well the extra rendering I meant is because I don't clearly see the demarcation line between the two either.
So is it ReactNode rendering a JSX? in which case a JSX element wouldn't save that step, besides the whole result of the current component being rendered after.

gildub:
@fdupont-redhat, we need to give credits to the original designer to @vconzola.
And credits too to @mturley as my de-facto mentor.

gildub:
Actually I think the only thing that would make it almost perfect is a `ResourcesEmptyIcon` per @vconzola initial mockup.
It doesn' exist in PF. So I've created a RFE, just in case:
https://github.com/patternfly/patternfly-react/issues/4979

mturley:
Nice, thanks for opening that issue @gildub.

src/app/common/components/PipelineSummary.tsx
```diff
@@ -1,19 +1,124 @@
...
+const Dash: React.FunctionComponent<IDashProps> = ({ isReached }: IDashProps) => {
+  return (
+    <FlexItem alignSelf={{ default: 'alignSelfCenter' }}>
+      {isReached ? <div className="dash dashReached" /> : <div className="dash dashNotReached" />}
+    </FlexItem>
+  );
+};
+
+interface IChainProps {
+  /* eslint-disable @typescript-eslint/no-explicit-any */
```
mturley, line 32: Sorry for one last little thing.. can you make this `eslint-disable-next-line` instead of `eslint-disable`? As-is it disables that rule for the rest of the file or until an `eslint-enable` comment.

mturley (commented):
Looks fantastic! One tiny thing about the eslint comment and then I think it's good to go

gildub:
@mturley, ready for final review.

mturley (approved):
LGTM 👍

vconzola:
@mturley I agree with @gildub about the empty icon. If we can we use the icon he created before it's accepted into PF, I say let's do it. Ideally, we'd have a way to incorporate both the "almost empty" and "almost full" icons along with the empty and full ones for each step, but that could be an enhancement later.

mturley:
@vconzola we opened an issue on PF, we don't have an icon created yet, but we could make one if it's a priority. I'll open an issue to follow up on that. https://github.com/konveyor/virt-ui/issues/115

vconzola:
@mturley My bad. I misread @gildub comment. Definitely, a nice to have, not a priority.

Issue #85, opened 2020-09-28
Figure out how to set/get migration network details for a VMware host in the API


mturley:
The hosts table has columns for Network, Bandwidth and MTU, and there is an operation to set the selected network for a host. @jortel do you know how/where we would get/set this data in the API? Currently I see how to get the networks on each host, but no notion of a "selected migration network" and no Bandwidth or MTU values.

Relevant mockups:
![Screen Shot 2020-09-28 at 5 06 24 PM](https://user-images.githubusercontent.com/811963/94486247-206e4680-01ad-11eb-9437-055bb9052cad.png)

![Screen Shot 2020-09-28 at 5 06 29 PM](https://user-images.githubusercontent.com/811963/94486257-23693700-01ad-11eb-976f-13d8f4aa119c.png)

mturley:
Closing this issue in favor of #138.

PR #102, opened 2020-10-05
Select Host mapping network more details

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/45

### DIFF
--- a/src/app/Providers/components/VMwareProviderHostsTable/SelectNetworkModal.tsx
+++ b/src/app/Providers/components/VMwareProviderHostsTable/SelectNetworkModal.tsx
@@ -10,16 +10,39 @@ interface ISelectNetworkModalProps {
   selectedHosts: IHost[];
   onClose: () => void;
 }
+// TODO Replace once https://github.com/konveyor/virt-ui/issues/85 has been addressed.
+interface ITempHostNetwork {
+  name: string;
+  address: string;
+  bandwidth: string;
+  mtu: number;
+  isDefault?: boolean;
+}
+
+const MOCK_NETWORKS: ITempHostNetwork[] = [
+  {
+    name: 'storage_network',
+    address: '192.168.0.1/24',
+    bandwidth: '1 GB/s',
+    mtu: 1499,
+    isDefault: true,
+  },
+  { name: 'compute_network', address: '192.168.0.2/24', bandwidth: '2 GB/s', mtu: 1400 },
+  { name: 'other_network', address: '192.168.0.3/24', bandwidth: '3 GB/s', mtu: 1500 },
+];
+// End TODO
 
 const SelectNetworkModal: React.FunctionComponent<ISelectNetworkModalProps> = ({
   selectedHosts,
   onClose,
 }: ISelectNetworkModalProps) => {
-  const networkOptions = selectedHosts.map((host) => ({
-    toString: () => `${formatHostNetwork(host.network)}`,
-    value: host.network.name,
-    props: { description: `${host.bandwidth}; MTU:${host.mtu}` },
+  // TODO Replace once https://github.com/konveyor/virt-ui/issues/85 has been addressed.
+  const networkOptions = MOCK_NETWORKS.map((network) => ({
+    toString: () => formatHostNetwork(network),
+    value: network.name,
+    props: { description: `${network.bandwidth}; MTU:${network.mtu}` },
   }));
+  // End TODO
 
   // TODO add a library like Formik, react-final-form, react-hook-form and use it for validation?
   //   or maybe roll our own simple validation? maybe use Yup?
@@ -72,7 +95,6 @@ const SelectNetworkModal: React.FunctionComponent<ISelectNetworkModalProps> = ({
               setSelectedNetwork((selection as OptionWithValue<string>).value)
             }
             placeholderText="Select a network..."
-            // description=<{Text>blah</Text>}
           />
         </FormGroup>
         <div>
--- a/src/app/Providers/components/VMwareProviderHostsTable/SelectNetworkModal.tsx
+++ b/src/app/Providers/components/VMwareProviderHostsTable/SelectNetworkModal.tsx
@@ -16,8 +16,9 @@ const SelectNetworkModal: React.FunctionComponent<ISelectNetworkModalProps> = ({
   onClose,
 }: ISelectNetworkModalProps) => {
   const networkOptions = selectedHosts.map((host) => ({
-    toString: () => `${formatHostNetwork(host.network)} - ${host.bandwidth} - ${host.mtu}`,
+    toString: () => `${formatHostNetwork(host.network)}`,
     value: host.network.name,
+    props: { description: `${host.bandwidth}; MTU:${host.mtu}` },
   }));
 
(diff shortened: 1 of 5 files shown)


vconzola (changes requested):
@gildub I see that the details are only inline for now. Let me know if you run into technical issues with displaying them as shown in the mockups.

src/app/Providers/components/VMwareProviderHostsTable/SelectNetworkModal.tsx
```diff
@@ -11,21 +11,13 @@ interface ISelectNetworkModalProps {
...
   selectedHosts,
   onClose,
 }: ISelectNetworkModalProps) => {
-  console.log('TODO: generate options for selected hosts: ', selectedHosts);
-  const networkOptions = MOCK_NETWORKS.map((network) => ({
-    toString: () => formatHostNetwork(network),
-    value: network.name,
+  const networkOptions = selectedHosts.map((host) => ({
+    toString: () => `${formatHostNetwork(host.network)} - ${host.bandwidth} - ${host.mtu}`,
+    value: host.network.name,
```
mturley, line 20: @gildub to display the bandwidth/MTU as an option description like the mockups show, you can add `props: { description: '...' }` to these option objects. Any props available on the PatternFly `<SelectOption>` can be passed this way, it's an optional property of our `OptionWithValue` interface in SimpleSelect.

gildub: @vconzola, yes I did have an issue having those descriptions on a separate line. @mturley thanks for the pointers.

gildub:
@vconzola, please review.

vconzola:
@gildub Looks good.  Thanks.

src/app/Providers/components/VMwareProviderHostsTable/VMwareProviderHostsTable.tsx
```diff
@@ -25,11 +25,9 @@ const VMwareProviderHostsTable: React.FunctionComponent<IVMwareProviderHostsTabl
 
   const getSortValues = (host: IHost) => {
     const { name, network, bandwidth, mtu } = host;
-    // First cell is the generated checkbox from using onSelect.
-    return ['', name, 'network', 'bandwidth', 'mtu'];
-    //TODO update when api data is available
-    // return ['', name, formatHostNetwork(network), bandwidth, mtu];
+    return ['', name, formatHostNetwork(network), bandwidth, mtu];
```
ibolton336, line 28: Looking good so far.. a few issues are showing up for me in remote mode. Are you using VScode? The ts compiler is complaining about a few things on this page 

image

ibolton336 (dismissed):

src/app/Providers/components/VMwareProviderHostsTable/helpers.ts
```diff
@@ -2,5 +2,6 @@ import { IHostNetwork } from '@app/queries/types';
 
 export const formatHostNetwork = (network: IHostNetwork): string => {
   const { name, address, isDefault } = network;
-  return `${name} - ${address}${isDefault ? ' (default)' : ''}`;
+  const line = `${name} - ${address}${isDefault ? ' (default)' : ''}`;
```
ibolton336, line 5: We should make "line" a more descriptive variable name so it is self documenting.

gildub: Absolutely!

src/app/Providers/components/VMwareProviderHostsTable/VMwareProviderHostsTable.tsx
gildub: There was an existing issue, I've removed unused attribute of `UsePaginationState`.
I can't see anything else and `yarn ci` doesn't report any error on that component/page.
I do use VSCode and I've seen it lagging sometimes forcing me to reload editor.

mturley: Hm, I can't reproduce those errors you were seeing @ibolton336

ibolton336: @mturley I think my editor was using a newer version of ts. 4.0.9. Just changed it to use the workspace version.

mturley:
Looking at the source, I think the issues with API data in remote mode may predate this PR but should be fixed.

I'm seeing this error when visiting the hosts page:
Screen Shot 2020-10-07 at 6 52 16 PM

The problem here is that the real host object coming from the API doesn't match our `IHost` interface. The real data looks like this:
Screen Shot 2020-10-07 at 6 57 53 PM

So there is no `network` property on a host, which is why `formatHostNetwork` is crashing. We may need to check with @jortel about how the inventory API can tell us which migration network is selected for a host, and we need to correct our interface and mock data to match.

If we want to merge this PR ahead of fixing this issue, we may want to put the mock data back in place in SelectNetworkModal for now and comment out the formatHostNetwork lines in the table component again. That way we can merge this without breaking `yarn start:dev:remote`.

mturley:
We have https://github.com/konveyor/virt-ui/issues/85 open to address the underlying issue I'm describing.

src/app/Providers/components/VMwareProviderHostsTable/VMwareProviderHostsTable.tsx
mturley: @ibolton Re: handling the case where a host has an undefined network, maybe that's the right solution but I'm not sure. Depending on how we get the real network data from the API, wherever it actually is it might always be defined? I guess we do need to think about a case where there is no network selected for a host.

ibolton336: @mturley Ah yeah the undefined check will probably at the component render function level so probably no need to check for undefined here. +1 for setting the network cell back to a 'network' string in remote mode for now until #85 is fixed.

gildub:
@mturley, @ibolton336, altering IHost and IHostNetwork is impacting parent components so I restored `MOCK_NETWORKS` augmented with the field we need so the UI part of this PR remains coherent.

gildub:
@mturley, @ibolton336 , @jortel, 
Per our last scrum discussion it turns out that the migration network is to be determined by identifying an IP address.
Putting this one on hold until better visibility

mturley:
@gildub I'm still getting the `Cannot read property 'name' of undefined` error when I open the VMware providers page. It looks like you're still calling `formatHostNetwork` with the `host.network` value in there, we probably need to use one of the MOCK_NETWORKS there too

gildub:
@mturley, no worries, I'll get back to it first thing on Monday.

gildub:
@mturley, of course having such data gap between mock and remote data is far from ideal. I think the best thing to do for now is just allow undefined network data until we close the gap.

mturley:
@gildub the undefined data is fine as long as it doesn't crash the page when running in remote data mode.

mturley (approved):
LGTM, thanks for the fix @gildub !

PR #106, opened 2020-10-06
Fix handling of CORS errors in authorizedFetch

mturley:
In my commits for #94 I broke the redirect for CORS errors, this fixes it and tweaks the look and feel of that page a bit.


ibolton336 (approved):
LGTM! Thanks for cleaning this up.

PR #107, opened 2020-10-07
useFormState for SelectNetworkModal and AddEditMappingModal

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/60

### DIFF
--- a/src/app/Mappings/components/AddEditMappingModal.tsx
+++ b/src/app/Mappings/components/AddEditMappingModal.tsx
@@ -6,13 +6,13 @@ import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import SimpleSelect, { OptionWithValue } from '@app/common/components/SimpleSelect';
 import { MappingBuilder, IMappingBuilderItem } from './MappingBuilder';
 import { getMappingFromBuilderItems } from './MappingBuilder/helpers';
-import { Mapping, MappingType, IOpenShiftProvider, IVMwareProvider } from '@app/queries/types';
+import { MappingType, IOpenShiftProvider, IVMwareProvider } from '@app/queries/types';
 import { useProvidersQuery, useMappingResourceQueries } from '@app/queries';
 import { updateMockStorage } from '@app/queries/mocks/helpers';
 import './AddEditMappingModal.css';
 import { usePausedPollingEffect } from '@app/common/context';
 import LoadingEmptyState from '@app/common/components/LoadingEmptyState';
-import { useFormField, useFormState } from '@app/common/hooks/useFormState';
+import { useFormField, useFormState, getFormGroupProps } from '@app/common/hooks/useFormState';
 import ValidatedTextInput from '@app/common/components/ValidatedTextInput';
 
 interface IAddEditMappingModalProps {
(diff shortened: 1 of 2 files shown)


mturley:
@gildub do you plan to also add useFormState to the AddEditMappingModal in this PR? If not we should remove the "Resolves #60"

gildub:
@mturley, yes it's the plan, I've updated the description. Thanks

src/app/Mappings/components/AddEditMappingModal.tsx
```diff
@@ -1,36 +1,58 @@
...
+    newMappingName: useFormField('', yup.string()),
+  }),
+  storageMapping: useFormState({
+    mapping: useFormField<Mapping | null>(null, yup.mixed<Mapping>().required()),
+    isSaveNewMapping: useFormField(false, yup.boolean().required()),
+    newMappingName: useFormField('', yup.string()),
+  }),
+});
+
+export type MappingFormState = ReturnType<typeof useMappingFormState>;
```
mturley, line 48: We don't necessarily need to export this type or even have this form lifted into a `useMappingFormState`, that was just useful to have `PlanWizardFormState` as a prop type for passing forms down into the wizard steps.

Also I think we only need one `useFormState` in this component, we can remove the `networkMapping` and `storageMapping` useFormState calls here because the mapping field will be part of that singular form (the one you call `mapping` right now), and its fields are redundant (isSaveNewMapping is only relevant in the wizard, `newMappingName` is redundant with `name`, and we don't need separate state for network vs storage since the modal will just be using one at a time).

So I think the whole thing can be replaced with:

```ts
const form = useFormState({
  name: ...
  sourceProvider,
  targetProvider,
  // and then in my PR i'll add `mapping` here
});
```

and it can probably go right inside the component below instead of having `useMappingFormState`, just like you did in SelectNetworkModal (although that doesn't really matter either way)

src/app/Mappings/components/AddEditMappingModal.tsx
```diff
@@ -113,27 +131,37 @@ const AddEditMappingModal: React.FunctionComponent<IAddEditMappingModalProps> =
...
+                </Form>
               </GridItem>
               <GridItem />
               <GridItem sm={12} md={5}>
-                <FormGroup label="Source provider" isRequired fieldId="source-provider">
+                <FormGroup
+                  label="Source provider"
+                  isRequired
+                  fieldId="source-provider"
+                  helperTextInvalid={forms.mapping.fields.sourceProvider.error}
```
mturley, line 149: We can use the `getFormGroupProps` helper from the useFormState file for these validation-related props.
```ts
{...getFormGroupProps(form.fields.sourceProvider)}
```

mturley: Also we should use it in the FormGroup around the target provider field below as well.

src/app/Providers/components/VMwareProviderHostsTable/SelectNetworkModal.tsx
```diff
@@ -51,32 +63,38 @@ const SelectNetworkModal: React.FunctionComponent<ISelectNetworkModalProps> = ({
...
+          fieldId="admin-password"
+        />
         <FormGroup
           label="Network"
           isRequired
           fieldId="network"
-          helperTextInvalid="TODO"
-          validated="default" // TODO add state/validation/errors to this and other FormGroups
+          helperTextInvalid={form.fields.selectedNetwork.error}
+          validated={form.fields.selectedNetwork.isValid ? 'default' : 'error'}
```
mturley, line 83: If you have `getFormGroupProps` spread here you don't need the `helperTextInvalid` and `validated` props, they are provided by `getFormGroupProps`.

mturley (changes requested):
Looking good, a couple of minor things for Monday

src/app/Mappings/components/AddEditMappingModal.tsx
gildub: Ok, actually it's interesting how we can use file level variables and export them to lift up state.

Yeah you're right this ended up over-complicated because I started addressing MappingBuilder which is not happening in this PR anymore.

mturley (approved):
LGTM! thanks @gildub

PR #113, opened 2020-10-08
Plan wizard: Select VMs step: traverse tree selections to generate VM selection table contents

### LINKED ISSUE #81: Add API query for VMs filtered by tree selections

mturley:
Resolves #81 

It turns out we didn't need a separate API query for the VMs, since all the data we need for this table is already in the tree structure.

The tree traversal helpers are a bit confusing... let me know if I can make any of that code more readable. It's not the most efficient thing, so I leveraged `React.useMemo` to memoize it and only recompute if the tree query data changes.


ibolton336 (approved):
LGTM!

PR #114, opened 2020-10-09
Add Dockerfile for konveyor/virt-ui

fdupont-redhat:
This pull request adds a Dockerfile to build the konveyor/virt-ui image. It is based on ubi8/nodejs-12 provided and supported by Red Hat. Yarn is installed using npm, as the method based on RPM requires being root and doesn't provide more security.


Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+COPY --from=builder /virt-ui/config /opt/app-root/src/config
+COPY --from=builder /virt-ui/deploy /opt/app-root/src/deploy
+COPY --from=builder /virt-ui/dist /opt/app-root/src/dist
+COPY --from=builder /virt-ui/dist/index.html.ejs /opt/app-root/src/views/index.html.ejs
+COPY --from=builder /virt-ui/node_modules /opt/app-root/src/node_modules
+COPY --from=builder /virt-ui/package.json /opt/app-root/src
+
+ENV VIRTMETA_FILE="/etc/virt-ui/virtMeta.json"
+
+ENTRYPOINT ["npm", "run", "-d", "start"]
```
gildub, line 36: Shouldn't it be `yarn` instead of running `npm`?

gildub (commented):
LGTM

Dockerfile
fdupont-redhat: IIUC, `yarn` will wrap `npm` that also wraps `node`. But happy to change if `yarn` has a config to run the service.
I've taken this from [UBI8 NodeJS 12 homepage](https://catalog.redhat.com/software/containers/ubi8/nodejs-12/5d3fff015a13461f5fb8635a).

fdupont-redhat:
I have tried to make the image smaller by starting from `ubi8/ubi` instead of `ubi8/nodejs-12` and it only saved 40 MB, i.e. 3.4%, while make the Dockerfile more complex. So, I'm keeping `ubi8/nodejs-12` as the base image.

Dockerfile
mturley: I `think` as long as `yarn install` is used instead of `npm install` to install the dependencies, that's what's important (since `yarn.lock` needs to be used to lock the versions). I'm not sure it matters which CLI is used to run the scripts. If this maps to the CLI command `npm run -d start`, then you should be able to use `ENTRYPOINT ["yarn" "start"]` I imagine? (`yarn start` is the correct command)

fdupont-redhat: In the runner image, `yarn` is not present. It's only present in the `builder` image.

mturley (approved):

PR #117, opened 2020-10-09
Use create provider hook to create provider & relevant secret

ibolton336:
This PR adds post support using the k8s client pulled in from lib-ui along with react-query's useMutation hook. 
- fixes storage of currentUser to include full token information. Issue opened for handling token expiry #137 
- adds mockableMutation for ensuring post requests don't break mock mode. 
- helper function to check for duplicate requests
- logic for handling rollback in the event that either a secret or provider add fails while the other succeeds.


ibolton336:
TODO:
- Force cert acceptance on sign in to avoid SSL errors when making requests to the cluster
- never ending loading screen on host page api hit on new providers with 0 associated hosts

src/app/queries/fetchHelpers.ts
```diff
@@ -43,3 +76,11 @@ export const useAuthorizedFetch = <T>(url: string): QueryFunction<T> => {
   const fetchContext = useFetchContext();
   return () => authorizedFetch(url, fetchContext);
 };
+
+export const useAuthorizedMutate = <T, TData>(
```
mturley, line 80: This still works as a hook because it just returns a function, even though that function is imperative 👍

mturley: We might want to call it `useAuthorizedPost` though to distinguish it from the react-query mutate stuff

src/app/queries/helpers.ts
```diff
@@ -27,6 +37,34 @@ export const useMockableQuery = <TResult = unknown, TError = unknown>(
     queryFn: process.env.DATA_SOURCE !== 'mock' ? config.queryFn : () => mockPromise(mockData),
   });
 
+// export const useMockableMutation = <
```
mturley, line 40: I'll revisit this next. Looks like you were close.

src/app/queries/helpers.ts
```diff
@@ -27,6 +37,34 @@ export const useMockableQuery = <TResult = unknown, TError = unknown>(
...
+//   // mutationFn:()=> console.log('help')
+//     // process.env.DATA_SOURCE !== 'mock'
+//     //   ? config.onMutate
+//     //   : () => window.alert('Sorry, this feature is not available in mock mode.'),
+// }
+// );
+// export const getClusterApiUrl = (relativePath: string): string =>
+//   `${VIRT_META.clusterApi}/apis/virt.konveyor.io/v1alpha1/namespaces/openshift-migration${relativePath}`;
+
 export const getApiUrl = (relativePath: string): string =>
```
mturley, line 68: Let's rename this to `getInventoryApiUrl` to distinguish it from `getClusterApiUrl`

src/app/queries/providers.ts
```diff
@@ -19,6 +52,66 @@ export const useProvidersQuery = (): QueryResult<IProvidersByType> => {
 
   return sortIndexedResultsByName<Provider, IProvidersByType>(result);
 };
+export const useCreateProvider = () => {
+  const client = useClientInstance();
+  const useProviderPost = async (
```
mturley, line 57: Same here, this is async / imperative, it shouldn't have `use` in the name

src/app/queries/providers.ts
```diff
@@ -19,6 +52,66 @@ export const useProvidersQuery = (): QueryResult<IProvidersByType> => {
...
+        queryCache.invalidateQueries('providers');
+      },
+      onError: () => {
+        console.log('did we fail');
+      },
+    }
+  );
+
+  const createProvider = async (formValues) => {
+    try {
```
mturley, line 100: I think maybe the `try/catch` here isn't necessary, I'm not sure but I think any exceptions would be caught inside the post function above, and then `onError` would be called instead of an exception making it to this point

src/app/queries/providers.ts
```diff
@@ -19,6 +52,66 @@ export const useProvidersQuery = (): QueryResult<IProvidersByType> => {
...
+      await mutate(formValues);
+    } catch (error) {
+      // Uh oh, something went wrong
+    }
+  };
+  return {
+    createProvider,
```
mturley, line 107: That means I think we could literally just return `createProvider: mutate` here, instead of having a `const createProvider` above.

tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "outDir": "dist",
     "module": "esnext",
     "target": "es5",
-    "lib": ["es6", "dom", "es2019"],
+    "lib": ["es6", "dom", "es2019", "ES2020"],
```
mturley, line 8: just curious, which es2020 syntax are you using? we should make sure this doesn't impact our supported browsers (webpack probably transpiles it away but not sure)

mturley (commented):
Just some notes while wrapping my head around this. A lot of them are notes to myself-- let's maybe communicate around who is claiming what if we address these, i'll take some of them.

tsconfig.json
mturley: (I realized it was for `Promise.allSettled`, which was still giving me errors, so I added `"ES2020.Promise"` which solved it)

mturley:
Self-addressed my feedback and added useMockableMutation in https://github.com/ibolton336/virt-ui/pull/2

mturley (approved):
LGTM! Yikes, the rollback stuff does seem like a pain in the ass. Hopefully we can factor that out into something generic for all the resources somehow.

PR #121, opened 2020-10-12
Cloud MA link

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/110


mturley (dismissed):

mturley:
@gildub I think I screwed up when I created the Button in CloudAnalyticsInfoAlert, it's not working when you click on it. I thought `variant="link"` would make it render an `<a>` and the href prop would work, but I guess not. We should either use a plain HTML `<a href="">` or use an `onClick` instead of `href` to set `window.location` or something, whichever looks right with spacing etc.

gildub:
@mturley, No worries I'll look into it.

mturley (approved):
Perfect 👍

Issue #76, opened 2020-09-23
Add a FilterToolbar to the plans table


mturley:
Leaving this out of #66, let's add it as a follow-up. To the left of the Create button there should be a FilterToolbar (driven by useFilterState) to allow the user to filter by a few columns.

![Screen Shot 2020-09-23 at 2 31 21 PM](https://user-images.githubusercontent.com/811963/94054291-7a37d080-fda9-11ea-852b-6f2676607d12.png)

@vconzola which columns should we able to filter on here? I'm guessing all of them except number of VMs?

vconzola:
@mturley Correct, filter by all columns except number of VMs.

PR #125, opened 2020-10-13
Add FilterToolbar to migration plan table

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/76

### DIFF
--- a/src/app/Plans/components/PlansTable.tsx
+++ b/src/app/Plans/components/PlansTable.tsx
@@ -32,6 +32,8 @@ import { IPlan, IMigration } from '@app/queries/types';
 import './PlansTable.css';
 import { PlanStatusType, PlanStatusConditionsType } from '@app/common/constants';
 import CreatePlanButton from './CreatePlanButton';
+import { FilterToolbar, FilterType, FilterCategory } from '@app/common/components/FilterToolbar';
+import { useFilterState } from '@app/common/hooks/useFilterState';
 
 interface IPlansTableProps {
   plans: IPlan[];
@@ -42,6 +44,19 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
   plans,
   migrations,
 }: IPlansTableProps) => {
+  const filterCategories: FilterCategory[] = [
+    {
+      key: 'name',
+      title: 'Name',
+      type: FilterType.search,
+      placeholderText: 'Filter by name...',
+      getItemValue: (item) => {
+        return item.metadata.name;
+      },
+    },
+  ];
+
+  const { filterValues, setFilterValues, filteredItems } = useFilterState(plans, filterCategories);
   const getSortValues = (plan: IPlan) => {
     return [
       plan.metadata.name,
@@ -53,7 +68,7 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
     ];
   };
 
-  const { sortBy, onSort, sortedItems } = useSortState(plans, getSortValues);
+  const { sortBy, onSort, sortedItems } = useSortState(filteredItems, getSortValues);
   const { currentPageItems, setPageNumber, paginationProps } = usePaginationState(sortedItems, 10);
   React.useEffect(() => setPageNumber(1), [sortBy, setPageNumber]);
 
@@ -169,6 +184,13 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
   return (
     <>
       <Level>
+        <LevelItem>
+          <FilterToolbar
+            filterCategories={filterCategories}
+            filterValues={filterValues}
+            setFilterValues={setFilterValues}
+          />
+        </LevelItem>
         <LevelItem>
           <CreatePlanButton variant="secondary" label="Create" />
         </LevelItem>
--- a/src/app/Plans/components/PlansTable.tsx
+++ b/src/app/Plans/components/PlansTable.tsx
@@ -54,6 +54,24 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
         return item.metadata.name;
       },
     },
+    {
+      key: 'sourceProvider',
+      title: 'Source provider',
+      type: FilterType.search,
+      placeholderText: 'Filter by name...',
+      getItemValue: (item) => {
+        return item.spec.provider.sourceProvider.name;
+      },
+    },
+    {
+      key: 'targetProvider',
+      title: 'Target provider',
+      type: FilterType.search,
+      placeholderText: 'Filter by name...',
+      getItemValue: (item) => {
+        return item.spec.provider.destinationProvider.name;
+      },
+    },
   ];
 
   const { filterValues, setFilterValues, filteredItems } = useFilterState(plans, filterCategories);
--- a/src/app/Plans/components/PlansTable.tsx
+++ b/src/app/Plans/components/PlansTable.tsx
@@ -80,12 +80,11 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
       getItemValue: (item) => {
         const res = item.status.conditions.find(
           (condition) =>
-            condition.type === PlanStatusConditionsType.Ready ||
-            condition.type === PlanStatusConditionsType.Execute ||
-            condition.type === PlanStatusConditionsType.Finished ||
-            condition.type === PlanStatusConditionsType.Error
+            condition.type === PlanStatusType.Ready ||
+            condition.type === PlanStatusType.Execute ||
+            condition.type === PlanStatusType.Finished ||
+            condition.type === PlanStatusType.Error
         );
-
         return res ? PlanStatusConditionsType[res.type] : '';
       },
     },
--- a/src/app/Plans/components/PlansTable.tsx
+++ b/src/app/Plans/components/PlansTable.tsx
@@ -72,6 +72,23 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
         return item.spec.provider.destinationProvider.name;
       },
     },
+    {
+      key: 'status',
+      title: 'Status',
+      type: FilterType.search,
+      placeholderText: 'Filter state...',
+      getItemValue: (item) => {
+        const res = item.status.conditions.find(
+          (condition) =>
+            condition.type === PlanStatusConditionsType.Ready ||
+            condition.type === PlanStatusConditionsType.Execute ||
+            condition.type === PlanStatusConditionsType.Finished ||
+            condition.type === PlanStatusConditionsType.Error
+        );
+
+        return res ? PlanStatusConditionsType[res.type] : '';
+      },
+    },
   ];
 
   const { filterValues, setFilterValues, filteredItems } = useFilterState(plans, filterCategories);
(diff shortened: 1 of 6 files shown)


gildub:
@vconzola, besides plan migration `name` are there any other fields needing to have filter option?

vconzola:
@gildub Yes, user should be able to filter by Source provider, Target provider, and Plan status.

src/app/Plans/components/PlansTable.tsx
```diff
@@ -42,6 +44,19 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
   plans,
   migrations,
 }: IPlansTableProps) => {
+  const filterCategories: FilterCategory[] = [
```
mturley, line 47: Per https://github.com/konveyor/virt-ui/issues/76#issuecomment-697908751 we'll need to add filter categories for all columns except number of VMs. so I think it would be Search filters for source provider / target provider (although maybe Select instead since we can query the list of available providers), and a Select filter for the plan status (logic might be tricky there... I can't remember if I made the Select filter able to have custom filter logic and not just a plain string match. if that's hard we could wait for another PR).

@vconzola should the filters for source/target provider in the plans table be search filters or select/dropdown filters?

src/app/Plans/components/PlansTable.tsx
```diff
@@ -169,6 +184,13 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
...
       <Level>
+        <LevelItem>
+          <FilterToolbar
+            filterCategories={filterCategories}
+            filterValues={filterValues}
+            setFilterValues={setFilterValues}
+          />
+        </LevelItem>
         <LevelItem>
           <CreatePlanButton variant="secondary" label="Create" />
```
mturley, line 195: The placement of the Create button in the dead center of the page is a little weird, I think we should put it in that first LevelItem right next to the FilterToolbar. We probably want it vertically-aligned at the top instead of the middle (so it stays in the same place visually when the filter chips appear) and with a medium margin from the toolbar, but I'm not sure what will be needed exactly until we move it.

src/app/common/hooks/useFilterState.ts
```diff
@@ -22,14 +22,8 @@ export const useFilterState = <T>(
       if (filterCategory?.getItemValue) {
         itemValue = filterCategory.getItemValue(item);
       }
-      return values.every((filterValue) => {
-        if (!itemValue) return false;
-        const lowerCaseItemValue = String(itemValue).toLowerCase();
-        const lowerCaseFilterValue = String(filterValue).toLowerCase();
-        return lowerCaseItemValue.indexOf(lowerCaseFilterValue) !== -1;
-      });
+      return values.every((filterValue) => (itemValue ? itemValue.includes(filterValue) : false));
```
mturley, line 25: Why this change? it was case-insensitive before due to the toLowerCase calls.

mturley (changes requested):

src/app/common/hooks/useFilterState.ts
mturley: Ah I see, is this to support the Select filter logic? If so we might want to have two different branches here depending on the filter type, so the Search ones can still be case-insensitive

src/app/Plans/components/PlansTable.tsx
vconzola: @mturley @gildub The Create button should be 16 px to the right of the search field as documented here: https://www.patternfly.org/v4/components/toolbar/react

src/app/Plans/components/PlansTable.tsx
vconzola: @mturley Not sure what you mean by search filters or select/dropdown filters, but they should be this: https://www.patternfly.org/v4/guidelines/filters#attribute-value-textbox-filters

mturley: @vconzola what we're calling a "search filter" is the first mockup shown at that link, where the value is typed in a text box. What we're calling a "select filter" is the second mockup there, where the value is chosen from a list.

I wasn't sure which would be better since provider names are plain text, but the list of possible values is known.

vconzola: @mturley Use what you're calling a search filter.

mturley: Got it, thanks

gildub: @mturley, sorry I totally zapped the comment on #76.

Adding them.

src/app/common/hooks/useFilterState.ts
gildub: Ok I didn't know the requirement the filter has to be case insensitive. I'll restore it.

mturley: I suppose it's not a hard requirement, I just thought it'd be nice. If it's difficult don't worry about it.

gildub: Not hard, just oversimplification shortcut I shouldn't taken.

src/app/Plans/components/PlansTable.tsx
gildub: @vconzola, I've re-positioned button. Please validate.

src/app/common/components/FilterToolbar/FilterToolbar.tsx
```diff
@@ -24,18 +25,18 @@ export interface OptionPropsWithKey extends SelectOptionProps {
...
+export interface IBasicFilterCategory<T> {
   key: string;
   title: string;
   type: FilterType; // If we want to support arbitrary filter types, this could be a React node that consumes context instead of an enum
-  getItemValue?: (item: any) => any;
+  getItemValue?: (item: T) => string | boolean;
 }
 
-export interface ISelectFilterCategory extends IBasicFilterCategory {
+export interface ISelectFilterCategory extends IBasicFilterCategory<IPlan> {
```
mturley, line 35: This should be `ISelectFilterCategory<T> extends IBasicFilterCategory<T>`, the `IPlan` can be passed in when it's used.

src/app/common/components/FilterToolbar/FilterToolbar.tsx
```diff
@@ -24,18 +25,18 @@ export interface OptionPropsWithKey extends SelectOptionProps {
...
   selectOptions: OptionPropsWithKey[];
 }
 
-export interface ISearchFilterCategory extends IBasicFilterCategory {
+export interface ISearchFilterCategory extends IBasicFilterCategory<IPlan> {
```
mturley, line 39: Same here

src/app/common/components/FilterToolbar/FilterToolbar.tsx
gildub: I needs a real type here because it's passing as parameter to the `IBasicFilterCategory` interface

src/app/common/components/FilterToolbar/FilterToolbar.tsx
gildub: same

src/app/common/components/FilterToolbar/FilterToolbar.tsx
mturley: Right, ok. The FilterToolbar file shouldn't be specific to Plans though, and `ISelectFilterCategory<T> extends IBasicFilterCategory<T>` will define this interface as a generic one. You can then use `ISelectFilterCategory<IPlan>` in the actual table component and that will define what `T` means here, so it will make its way to the `IBasicFilterCategory`.

mturley: To get this to work you will actually also need to add generics to `FilterCategory`, turn it into `export type FilterCategory<T> = ISearchFilterCategory<T> | ISelectFilterCategory<T>;` and then use `FilterCategory<IPlan>[]` as the type of the categories array in the table. That should get you the right types passed down.

gildub: Works fine!

src/app/common/components/FilterToolbar/FilterToolbar.tsx
gildub: same same, thanks

src/app/common/components/FilterToolbar/FilterControl.tsx
```diff
@@ -9,9 +9,10 @@ import {
 } from './FilterToolbar';
 import SelectFilterControl from './SelectFilterControl';
 import SearchFilterControl from './SearchFilterControl';
+import { IPlan } from '@app/queries/types';
```
mturley, line 12: We also shouldn't have `IPlan` in here, since FilterControl is also supposed to be generic. I know you had to leave for the day, I'll just push a commit to fix it and merge.

src/app/common/components/FilterToolbar/FilterToolbar.tsx
```diff
@@ -24,29 +25,29 @@ export interface OptionPropsWithKey extends SelectOptionProps {
...
-export type FilterCategory = ISearchFilterCategory | ISelectFilterCategory;
+export type FilterCategory<T> = ISearchFilterCategory<T> | ISelectFilterCategory<T>;
 
 export interface IFilterValues {
   [categoryKey: string]: FilterValue;
 }
 
 export interface IFilterToolbarProps {
-  filterCategories: FilterCategory[];
+  filterCategories: FilterCategory<IPlan>[];
```
mturley, line 50: Same here, we shouldn't have `IPlan` anywhere in `common/components/*`

src/app/common/hooks/useFilterState.ts
```diff
@@ -1,16 +1,17 @@
...
   filterValues: IFilterValues;
   setFilterValues: (values: IFilterValues) => void;
   filteredItems: T[];
 }
 
-export const useFilterState = <T>(
-  items: T[],
-  filterCategories: FilterCategory[]
-): IFilterStateHook<T> => {
+export const useFilterState = (
```
mturley, line 11: This was fine with `<T>`, we just have to thread the `T` down through all the interfaces with generics at every level. `IPlan` should only be in the table.

mturley (commented):
Everything looks good except for the final feedback about generics, but I don't want to hold this up. I'll push a commit to fix it in a moment, and we can look at it together after we merge.

mturley:
Ok @gildub the trick was to make everything that deals with the filter categories generic, including the components themselves. I figured out how to do that via https://wanago.io/2020/03/09/functional-react-components-with-generic-props-in-typescript/, basically the pattern is weird because we can't use `React.FunctionComponent` with a generic type interface. So we use some types it uses internally (`PropsWithChildren` and `JSX.Element`... I know I said basically never use it but this is an exception because it's what React expects a function component to return)

This is the basic pattern for a generic component that I arrived at, it's just like any other generic function:
```tsx
// Definition:
const MyComponent = <T>(props: React.PropsWithChildren<IMyComponentProps<T>>): JSX.Element | null => ...
// Usage:
<MyComponent<ISomething> some="props">
```

I think this is a really cool pattern we can use in a bunch of places for lib-ui! It's a shame the type annotation is so awkward though.

mturley (approved):
LGTM!

gildub:
@mturley, that's really nice. Well done.

src/app/common/components/FilterToolbar/FilterControl.tsx
gildub: Yeah exactly, that was not ideal but working!

Issue #126, opened 2020-10-13
Follow up when VM `concerns` array `severity` types are finalized


mturley:
Right now the `concerns` array on an `IVMwareVM` has objects with `severity` values assumed to be one of `'Warning' | 'Critical' | 'Advisory' | 'Info'`. We should follow up and make sure the possible values there are correct when @jortel has them finalized, and double check the logic in the `VMConcernsIcon` and `VMConcernsDescription` components.

mturley:
Also, when each concern has a `description` and `remedy` we should render those in `VMConcernsDescription` in addition to the `concern.name`.

mturley:
Related: https://github.com/konveyor/virt-ui/issues/238

gildub:
I asked Jeff for an update as of today.

gildub:
The design document has been updated, the only possible values are Advisory, Warn and Critical.
Therefore we need to remove Info and replace Warning with Warn.
![concerns](https://user-images.githubusercontent.com/1901741/100112680-ce296900-2e6f-11eb-9945-b2ed48d0f3fe.png)

gildub:
Actually we don't have to change Warning, Jeff is fixing the inventory so it uses Warning.

gildub:
Jeff confirms it has been merged, so we're aligned with inventory controller.

mturley:
Thanks for following up on this @gildub !

PR #129, opened 2020-10-14
Restructure mapping steps of wizard, use form state for mapping builder items, fix wizard corner cases

### LINKED ISSUE #103: Restructure mapping steps of wizard based on new mockups
Mockups TBD, the radio buttons from the mapping steps in the wizard are being replaced by a unified Select dropdown. As part of this issue we should make sure the "create new mapping on the fly" data is saved to the form state, which I left out of #100 because we'd have to rework it for this anyway.

mturley:
Resolves #103 

Another big one... ~~I'm pretty sure this finalizes all the wizard logic~~* except for actually creating the CRs.
\*welp, looks like we also have #126, #131 and #134.

* Adds VMs query and loads full VM objects on the select VMs step
  * Uses them to properly render the "Migration assessment" column and expanded content from the vm's `concerns` array (see #126)
  * Stores the full VM objects in form state instead of the tree nodes so their networks and disks can be referenced
* Adds a "N selected" indicator at the bottom of the Select VMs step to help indicate that you can have selections across multiple pagination pages (makes "select all" behavior more clear)
* Uses the networks and disks on the selected VMs to determine the required source datastores/networks for mapping steps
* Changes the layout / look-and-feel of the mapping steps of the wizard based on discussions with @vconzola 
  * Instead of radio buttons, we now have a single Select for choosing to either make a new mapping or start with an existing one
  * On making a selection, the mapping builder appears with items prefilled from the existing mapping if any, and any missing items based on required sources of the selected VMs are added (with highlights to show what was added)
  * Since the mapping builder will always have all required sources prefilled, the source selectors have been disabled / replaced by static source names, and all add/remove functionality has been removed. The only action you can take is to select targets
  * Offers to save the mapping for reuse only if you chose to create a new mapping, or the mapping you selected had missing sources added
* Switches the wizard to use the same mock mappings stored in localStorage that are currently used by the mappings page/modal, so that it will match up with real resources. Should be removed when we have real mapping CRUD.
* Fixes bugs in a few places where selection state was being compared with API data via referential equality. We shouldn't be doing this, since polling can replace the API objects with immutable copies that will no longer be considered equal, but represent the same objects. Instead compares things by selfLink or name.
* Disables polling for the whole wizard (was causing some issues with available options possibly changing after state was set)
* Moves all field state from local component state up to the `useFormState` level, so it is persisted when moving between wizard steps. The only exceptions are purely presentational state (dropdown open/close, search text, etc) which is ok to lose when leaving a step.
* Adds a shared yup schema for mapping builder items, with a custom `.test()` method to check that there are no empty selections in the items. Uses this as the schema of the `builderItems` form fields so that form validation follows mapping builder logic (can't proceed to the next step until all targets are selected). Used in both the wizard and the AddEditMappingModal.
* Adds useEffects that will reset dependent forms when certain selections are changed, to prevent stale / hidden form state.
  * When we change providers, we need to reset the Filter VMs tree selections, since those available options are loaded from the providers.
  * When we change the Filter VMs tree selections, we need to reset the selected VMs, since those available options come from the tree selections
  * When we change the selected VMs, we need to reset the mapping steps since those available options come from the selected VMs.
  * Because each of these effects is triggered by the last, each one has a ripple effect on the forms coming after it, e.g. changing providers will result in all subsequent forms being reset.
* Changes the logic of the `canJumpTo` properties on wizard steps: instead of always allowing jumping to steps that have been visited once, allows jumping to a step only if all forms leading up to that step are valid (so that the reset effects from above will block jumping ahead to steps without prerequisite selections).
  * Weird side-effect: when the form you're editing becomes valid, in addition to the Next button becoming enabled, the next step becomes jumpable in the nav immediately. I actually think this is nice, and it correctly reflects what you should be able to move to, but it's different from how the wizard step nav behaved before.
* Adds our first conditional validation: the name of a new mapping is only required when the "save mapping to use again" checkbox is checked. This is done by lifting the `useFormField` for that checkbox out and referencing its value when defining the schema for the name field. This required some changes to `useFormState` since we were caching the schema too aggressively.
* Since we're storing the mapping builder items directly in form state instead of mapping objects, we need to generate real mapping objects for the review step and for submitting the wizard. Adds a helper for this, used at the PlanWizard level so those objects are available for submission (will be null until required values in the forms are present)


mturley:
@vconzola I think this is ready for your review. To test the "highlighted missing sources" behavior, you can create a network mapping for providers `VCenter1` -> `OCPv_1` that includes `vmware-network-1` as the only source, and then in the plan wizard, select those same providers, select all datacenters and all VMs, and choose your mapping from the dropdown. You should see `vmware-network-2` added with the highlighted target, and it should turn white when you select the target.

mturley:
Something to remember: when we add edit capability, we'll need to change the form `reset()` calls so they explicitly set those forms to empty values instead of initial values (since the initial values for editing will come from the existing plan). Unless we prefill the existing plan another way of course.

gildub:
@mturley, Unless I missed something with the new approach, when I test with `dev:mock`, in the mapping steps, it seems the list of source (network or storage) items is wrong.
For example the storage mapping shows this when starting:
map

gildub:
For me, the previous network and storage mapping made more sense.
Anyway now we have two different ways between the wizard mapping and the network/storage mapping unless the plan is to use the new approach as well for the network/storage.
At least with the dev:mock test data, the static source names list seems incomplete, it shows one or two sources where the data as 4.

src/app/Plans/components/Wizard/VMConcernsDescription.tsx
```diff
@@ -0,0 +1,54 @@
+import { IVMwareVM } from '@app/queries/types';
+import { TextContent, Text, List, ListItem } from '@patternfly/react-core';
+import * as React from 'react';
+
+interface IVMConcernsDescriptionProps {
+  vm: IVMwareVM;
+}
+
+const criticalText = (
```
gildub, line 9: This might a question of style but since `criticalText` and siblings are not just used directly since there are all used onces. I understand it might be for clarity but then I was uncomfortable using Fragment everywhere so I wonder if it makes more sense to transform it to a React Node element.

gildub (commented):
LGTM besides my previous comment about the source list.

Quite many things happening across the board here.
I'll review again after https://github.com/konveyor/virt-ui/pull/124 has been merged

src/app/Plans/components/Wizard/VMConcernsDescription.tsx
mturley: Yeah, I only made them variables because the code was getting hard to read when they were inline. I'm not sure I know what you mean in terms of making them React Nodes, you mean components? The component version would still need to return a fragment, or something else like a span (the only reason I'm not using strings is so that I can render the `<strong>` in some of them)

mturley:
> For me, the previous network and storage mapping made more sense.
> Anyway now we have two different ways between the wizard mapping and the network/storage mapping unless the plan is to use the new approach as well for the network/storage.

Yeah, they are a little different now, but only because the version in the wizard has more context to work from. Since it knows what VMs you're migrating, it knows exactly what sources will be required, so there's no reason to make the user add and select them.

> At least with the dev:mock test data, the static source names list seems incomplete, it shows one or two sources where the data as 4.

Can you be more specific? It's intentional that it won't show all of the sources in the provider, it only shows the sources required to migrate the selected VMs.

mturley:
Ok thanks @vconzola. I can probably squeeze them in here, there's a lot of code to be reviewed anyway

src/app/Plans/components/Wizard/VMConcernsDescription.tsx
mturley: @gildub I added https://github.com/konveyor/virt-ui/pull/129/commits/f7f6eda428a6f5d7954e9423ef8611fb18446690 to clean this up a bit and consolidate the logic. Is that a bit better?

mturley:
Rebased and resolved conflicts with https://github.com/konveyor/virt-ui/pull/124

src/app/Plans/components/Wizard/VMConcernsDescription.tsx
gildub: @mturley, sorry about the confusion, you know between JSXElement,, ReactElement and ReactNode!

That is slicker!

mturley: Heh yeah, I know. The TS types for React aren't perfect... I wish the React core team would maintain them, but they come from the DefinitelyTyped community :) Facebook still wants to try and push Flow for typing...

My understanding is, basically don't ever use JSX.Element or ReactElement, since they are specific to various kinds of components / React internals. I think the best practice is to pretty much always just use ReactNode when referring to a variable that contains JSX, since it's the most inclusive type among them (JSX.Element and ReactElement are both assignable to ReactNode), so it can store JSX no matter where it came from. Most of the PF components that work with values like this use ReactNode.

mturley: I think even `null` is a valid `ReactNode` because you can return it from a component or render it as a child.

vconzola:
@mturley This looks good to me. If I ever find the old mockups I did with different instructional text we can change it later.

mturley:
Sounds good, I'll open an issue for that so we don't forget

mturley:
https://github.com/konveyor/virt-ui/issues/135

mturley:
@ibolton336 @gildub as soon as you guys are happy with this let's get it merged to unblock https://github.com/konveyor/virt-ui/issues/134

gildub (approved):
LGTM!

PR #133, opened 2020-10-14
Use virtMeta.json with a capital M everywhere

mturley:
This was inconsistent in the fallback default used by `server.js` and the Dockerfile value. It also should be changed to `virtMeta.json` with a capital M in the controller.


gildub (approved):
LGTM!

PR #136, opened 2020-10-15
Clean up types and async error handling, add mockable mutation helper, display error message on AddProviderModal

mturley:
When mutations are triggered in mock mode, it'll wait 1 second and then throw an error to be handled just like real API errors.


mturley:
oops, meant to open this against @ibolton336 's branch.

PR #139, opened 2020-10-19
Add target namespace selection to General step of plan wizard

### LINKED ISSUE #134: Add a field for selecting the target namespace for a plan
This came up in an email thread (subject: "How do we set target project?"), and it looks like we'll need to add a field to the General step of the plan wizard for specifying the target namespace. We'll need to add an inventory API query to get the namespace choices for the selected target provider, and disable or hide the field until the target provider is selected.

mturley:
Resolves #134 

The `/namespaces` API endpoint under a provider used by the query doesn't exist yet, so we shouldn't merge this until we can test against @jortel's updated cluster once that is added.

Ready for review, but marking as draft to prevent merging for now.


gildub (approved):
LGTM

mturley:
API changes are in place and I verified this works, so marking it ready and merging.

PR #140, opened 2020-10-20
Adds filter to all SelectVMs columns

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/131


vconzola:
@gildub Filter dropdown options look good except that "analysis" should be lowercase in "Migration analysis". Thanks.

gildub:
@mturley, please run a final review.

mturley (approved):
LGTM! the only thing I notice is that when the filters don't match any rows, the empty table looks a bit weird and we should render an EmptyState instead. Opened https://github.com/konveyor/virt-ui/issues/147 to follow up on that so we don't block this PR.

PR #142, opened 2020-10-20
Adds filter to VMMigrationDetails

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/141

Allows to filter VM name, start and end times.
Status could be added if needed, meanwhile we better have useMockableQuery in place first to be aligned with final data.

### DIFF
--- a/src/app/Plans/components/VMMigrationDetails.tsx
+++ b/src/app/Plans/components/VMMigrationDetails.tsx
@@ -9,6 +9,8 @@ import {
   Pagination,
   PageSection,
   Title,
+  Level,
+  LevelItem,
 } from '@patternfly/react-core';
 import {
   Table,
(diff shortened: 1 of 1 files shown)


gildub:
@vconzola, please confirm this is as expected.

vconzola:
@gildub In the filter dropdown, "Finished time" should be "End time" to be consistent with the column heading. Also, at some point we'll want to be able to filter (and sort) by status, but that can be handles as a separate PR, post beta.

gildub:
@vconzola, could you please review?

I've also created https://github.com/konveyor/virt-ui/issues/146 to follow up with the status column.

vconzola (commented):
LGTM. Thanks.

mturley (approved):
LGTM! One followup issue: the filters for start and end time should ideally be a date/time range picker instead of a search filter, but we don't quite have support for that in PF or FilterToolbar yet. There is a beta component https://www.patternfly.org/v4/components/date-picker we can use in the future, but it doesn't support times and ranges yet, and then once it does we'll also need to add a new type of FilterCategory for time ranges. Opened https://github.com/konveyor/virt-ui/issues/148.

PR #143, opened 2020-10-20
Make most relevant status to surface

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/64

### DIFF
--- a/src/app/Plans/components/PlanActionsDropdown.tsx
+++ b/src/app/Plans/components/PlanActionsDropdown.tsx
@@ -2,7 +2,8 @@ import * as React from 'react';
 import { Dropdown, KebabToggle, DropdownItem, DropdownPosition } from '@patternfly/react-core';
 
 import { IStatusCondition } from '@app/queries/types';
-import { PlanStatusConditionsType } from '@app/common/constants';
+import { StatusConditionsType } from '@app/common/constants';
+import { hasCondition } from '@app/common/helpers';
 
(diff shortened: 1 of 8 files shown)


vconzola:
@gildub It looks like there are 5 possible status conditions that will be displayed, but I can't follow the code well enough to understand which icon will be shown for each condition. Can you elaborate in a comment? Thanks.

gildub:
@vconzola, there are currently 3 available icons associated with Provider statuses: They are [`Ok, `Error` or `Warning`](https://github.com/konveyor/lib-ui/blob/master/src/components/StatusIcon/StatusIcon.tsx#L15).

We filter categories by severity as described [here](https://github.com/konveyor/virt-ui/issues/64#issuecomment-712928975). 

So when a given provider has at least one condition holding following category:
- `Critical` and `Error` then a Error icon will be displayed
- `Ready` then Ok icon to be used 
- Otherwise a warning icon to be displayed.

We could add more icons if needed along with a mapping between state and desired icon.

vconzola:
@gildub Sounds good. Thanks for the explanation.

mturley (commented):
@gildub LGTM! The only thing missing is that we should also use the new `mostSeriousCondition` helper for `getSortValues` on the OpenShiftProvidersTable, right?

gildub:
@mturley, absolutely, I've added it to the `OpenShiftProvidersTable`

mturley (approved):
Nice!

PR #144, opened 2020-10-20
Handle token expiry 

### LINKED ISSUE #137: handle token expiry with the k8s client
Implement a check for token expiry when making authorized requests with the k8s client.

ibolton336:
Closes #137 
Pulls in the setTokenExpiryHandler function from the k8s client library & defines it in the network context. Anytime we make an api call which fails with a 401 code using the k8s client, this function will be called internally by the k8s client. This will refresh the expired token by clearing out the currentUser from local storage, forcing a new oauth token to be requested and validated.

### DIFF
--- a/src/app/queries/providers.ts
+++ b/src/app/queries/providers.ts
@@ -1,6 +1,6 @@
 import { queryCache, QueryResult, useMutation, MutationResultPair } from 'react-query';
 
-import { usePollingContext } from '@app/common/context';
+import { usePollingContext, useNetworkContext } from '@app/common/context';
 import { POLLING_INTERVAL } from './constants';
 import {
   useMockableQuery,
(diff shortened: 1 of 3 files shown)


arpcv (approved):

gildub (approved):
LGTM

src/app/queries/providers.ts
```diff
@@ -47,6 +48,10 @@ export const useCreateProviderMutation = (
   AddProviderFormValues,
   unknown // TODO replace `unknown` for TSnapshot? not even sure what this is for
 > => {
+  //move this to its own abstracted post/mutation function
+  const { checkExpiry } = useFetchContext();
```
mturley, line 52: I'll rebase #145 on this and move it into the new error catch in useMockableMutation.

mturley (approved):
LGTM!

PR #145, opened 2020-10-21
Add useCreateMappingMutation and useMappingsQuery hooks, refactor mapping object types to match API, clean up mapping pages

### LINKED ISSUE #73: Add a query/useMutation hook for creating mappings
Depends on #72

mturley:
Resolves #73 

Creating and viewing mappings is now working end-to-end with real data!

* Refactors the mapping-related TS types to match the API structure (this involved a lot of cleanup)
* Adds a `useCreateMappingMutation` hook modeled after the `useCreateProviderMutation` implementation
* Calls the new mutation in the `AddEditMappingModal`
* Adds a new `MutationStatus` component for rendering the loading/error state of any mutation (might abstract this later to work with regular queries too, or create an equivalent. the error message structures are more based on whether it's a response from the inventory API or the kube client, not whether it's a query or mutation).
* Adds a static "Target namesspace pod network" option to the top of the target network selections, which results in a `{ type: 'pod' }` mapping target with no name/namespace.
* Adds a `useMappingsQuery` hook to list the mappings using the kube-client. Mostly based on our other queries, but integrating the kube client here required some new TS types (e.g. `IKubeList<T>`).
* Fixes the way we were importing the `queryCache` from react-query, so it invalidates properly. Creating a mapping will immediately refresh the list including your new mapping!
* Changes our `queryKey` template strings to arrays. This is just so it's simpler to invalidate queries with looser matching in the future, but it's also cleaner and I just wanted it all to be consistent with my new keys.
* Because it was annoying me and it's... sort of related, refactors the StorageMappingsPage and NetworkMappingsPage components to use one generic MappingsPage component (they were 99% identical already).

This doesn't add the create mapping calls to the plan wizard (for when the user selects to save a mapping for reuse), I figure we can handle that at the same time as creating plans. I did double check that my changes to the MappingBuilder here didn't break the usage in the wizard, though.


mturley:
@ibolton336 I noticed after merging https://github.com/konveyor/virt-ui/pull/144 that the `checkExpiry` function was expecting an error with a `response.status` object, which is only there if it's an axios error (the Fetch API error handling passes the Response object instead, which just has `status` at the top level). So as part of my rebase here I updated it to handle both.

gildub (dismissed):
LGTM

mturley:
Rebased and should be ready for final review.

ibolton336 (approved):
LGTM!

PR #149, opened 2020-10-22
Use TableEmptyState when filtering returns no results

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/147

### DIFF
--- a/src/app/Plans/components/PlansTable.tsx
+++ b/src/app/Plans/components/PlansTable.tsx
@@ -29,12 +29,14 @@ import { StatusIcon, StatusType } from '@konveyor/lib-ui';
 import PlanActionsDropdown from './PlanActionsDropdown';
 import { useSortState, usePaginationState } from '@app/common/hooks';
 import { IPlan, IMigration } from '@app/queries/types';
-import './PlansTable.css';
 import { PlanStatusType, StatusConditionsType } from '@app/common/constants';
 import CreatePlanButton from './CreatePlanButton';
 import { FilterToolbar, FilterType, FilterCategory } from '@app/common/components/FilterToolbar';
 import { useFilterState } from '@app/common/hooks/useFilterState';
 import { hasCondition } from '@app/common/helpers';
+import TableEmptyState from '@app/common/components/TableEmptyState';
+
+import './PlansTable.css';
 
 interface IPlansTableProps {
   plans: IPlan[];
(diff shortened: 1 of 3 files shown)


src/app/Plans/components/PlansTable.tsx
```diff
@@ -239,17 +240,24 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
...
+          sortBy={sortBy}
+          onSort={onSort}
+        >
+          <TableHeader />
+          <TableBody />
+        </Table>
+      ) : (
+        <TableEmptyState
+          titleText="No migration plans found"
+          bodyText="No results match your filter. Go back and make a different selection."
```
mturley, line 258: Nit-pick, but the "Go back" instruction only makes sense in the wizard, I think we can leave this as just "No results match your filter."

src/app/Plans/components/VMMigrationDetails.tsx
```diff
@@ -197,19 +198,26 @@ const VMMigrationDetails: React.FunctionComponent = () => {
...
+                  toggleVMExpanded(rowData.meta.migration);
+                }}
+              >
+                <TableHeader />
+                <TableBody />
+              </Table>
+            ) : (
+              <TableEmptyState
+                titleText="No migration details found"
+                bodyText="No results match your filter. Go back and make a different selection."
```
mturley, line 218: same here

src/app/Plans/components/Wizard/SelectVMsForm.tsx
```diff
@@ -275,20 +275,28 @@ const SelectVMsForm: React.FunctionComponent<ISelectVMsFormProps> = ({
...
+            toggleVMsExpanded(rowData.meta.vm);
+          }}
+        >
+          <TableHeader />
+          <TableBody />
+        </Table>
+      ) : (
+        <TableEmptyState
+          titleText="No VMs found"
+          bodyText="No results match your filter. Go back and make a different selection."
```
mturley, line 296: This one is good though.

mturley (commented):
LGTM, just a wording change

src/app/Plans/components/Wizard/SelectVMsForm.tsx
mturley: Oh wait actually, we shouldn't say "go back" here either, because the filters are on the same step in this case. I only had that phrasing for the case where you have to actually go back to the Filter VMs wizard step.

mturley: @gildub ^ (sorry if you were making a commit before I realized my mistake)

mturley (approved):

PR #151, opened 2020-10-23
Add useCreatePlanMutation and usePlansQuery hooks, refactor plan object types to match API, clean up plan / plan details pages

### LINKED ISSUE #150: Add query/useMutation hook for creating plans
Realized I never opened this, PR coming shortly.

mturley:
Resolves #150

* Adds `usePlansQuery` and `useCreatePlanMutation` hooks for listing / creating plans in the cluster.
* Refactors the plan object TypeScript interfaces to match the real CRDs, especially around status data
* Refactors a bunch of our components to fit the new plan structure, involved creating a few new lookup helpers
* Uses the query on the plans and VM migration details pages
  * The VM migration details page was previously built around displaying the Migration CR, but since all that data is also available on the Plan CR itself, I removed everything related to the Migration CR from the UI for now and drove all of this from the Plan. I'll need to reintroduce the types for Migration when I add the mutation to start a plan, but the structure is different anyway and I didn't want to leave unused outdated types in here.
* Adds provider and VM queries to the migration details page to look up context for the plan and render VM names (plan only contains VM ids)
* Adds helpers to generate plan objects from form values before creating them
* Uses the plan mutation to create plans on wizard submission 🎉 
* Uses the mapping mutation to create mappings from the plan wizard if "save mapping to use again" was selected on either of those steps
* Enhances the MutationStatus component to handle multiple sets of mutation results at once (shows one spinner for multiple concurrent mutations, but renders their errors separately if present)
* Adds a useEffect to the Review step that will reset the mutation statuses when leaving that step, so you can go back and it will clear your previous errors
* Adds logic to the VM migration details page that will detect a not-started plan (empty array of statuses) and instead create a bare status array with only the VM ids, so it can still display a list of VM names. Disables the expandable feature of the table in this mode.

Migration plans are real ✨


ibolton336 (approved):
:+1: Awesome job!

PR #154, opened 2020-10-25
Add useCreateMigrationMutation, wire up Start button 🎉, enhance MutationStatus with dismiss button and more options, fix progress/pipeline corner cases

mturley:
We can start real migrations now!

* Adds `useCreateMigrationMutation`, returns a create function which takes a plan and creates a bare Mutation CR referencing that plan.
  * Mutation CRs need a name, but it doesn't make sense to ask the user for one, so this generates a name by combining the plan name with the current timestamp.
* Wires up the mutation on the plans page! The mutation hook is lifted to the page component so it can display error status outside the card the PlansTable is contained in.
* Enhances the `MutationStatus` component so its spinner is optional (no spinner shifting down the plans table) and so it supports non-inline alerts, custom classNames for spacing, and a built-in close/dismiss button which will call the mutation result's `reset` function. This works on all the other mutation error alerts for providers, mappings, plans.
* Also disables the Cancel button on a running plan, since we won't have that implemented yet for beta.

I'm going to go ahead and try to use this to record a demo of a real migration from the UI 😄


gildub (approved):
Re-review: still looks good!

mturley:
@ibolton336 are you cool with merging this? I synced with @gildub and he's planning to look into some of the progress rendering corner cases

ibolton336:
Looks good!

PR #158, opened 2020-10-26
Remove all string length min/max validations in forms for now

### LINKED ISSUE #156: Text fields are limited to 20 characters
When creating a provider with the wizard, it complains that the username is longer than 20 characters and keeps the _Add_ button greyed out. 

image

Given that the default VMware user is `administrator@vsphere.local` is longer than 20 characters, it doesn't sound as a reasonable limit. Even more as it won't be displayed in any table where it could break the alignment. A lot of users will have connected their vCenter to Active Directory and will use domain based names, such as `john.doe@my.too.deep.domain.tree`, hence more than 20 characters.

Would it be possible to extend the limit to 256 characters?

mturley:
Resolves #156 .

We need to add more specific validation back to these fields, but we first need to determine what those requirements really are. https://github.com/konveyor/virt-ui/issues/157

### DIFF
--- a/src/app/Providers/components/AddProviderModal/AddProviderModal.tsx
+++ b/src/app/Providers/components/AddProviderModal/AddProviderModal.tsx
@@ -34,16 +34,16 @@ const useAddProviderFormState = () => {
   return {
     [ProviderType.vsphere]: useFormState({
       providerType: providerTypeField,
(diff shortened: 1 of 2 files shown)


fdupont-redhat (approved):
LGTM

PR #160, opened 2020-10-27
MigrationVMStepsType step to display human readable value

gildub:
This addresses the case where the step were displaying an internal value instead of human friendly string.

### DIFF
--- a/src/app/Plans/components/VMStatusTable.tsx
+++ b/src/app/Plans/components/VMStatusTable.tsx
@@ -37,7 +37,7 @@ const VMStatusTable: React.FunctionComponent<IVMStatusTableProps> = ({
               <Step status={status} index={index} />
             </FlexItem>
             <FlexItem>
-              <Text>{MigrationVMStepsType[step.name]}</Text>
+              <Text>{MigrationVMStepsType[step.name] || step.name}</Text>
             </FlexItem>
           </Flex>
         ),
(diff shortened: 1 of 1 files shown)


vconzola (commented):
LGTM.

mturley:
Oh, is the verbose description being added to the API data? We could just display that directly instead of mapping types to strings in the UI code, that'd be much better.

gildub:
@mturley, I suppose we could just keep this one in the meantime it won't hurt.

src/app/Plans/components/VMStatusTable.tsx
```diff
@@ -36,7 +37,7 @@ const VMStatusTable: React.FunctionComponent<IVMStatusTableProps> = ({
               <Step status={status} index={index} />
             </FlexItem>
             <FlexItem>
-              <Text>{step.name}</Text>
+              <Text>{MigrationVMStepsType[step.name]}</Text>
```
mturley, line 40: Re: handling the case where it is undefined, you could just use `MigrationVMStepsType[step.name] || step.name` to fall back on the name value itself

gildub: Yes, that's what I just did.

mturley (dismissed):
LGTM 👍

mturley (approved):

PR #164, opened 2020-10-27
Add proper form field validation based on real API expectations

### LINKED ISSUE #128: Enforce that user cannot create a plan or mapping with the name of an existing plan/mapping
We'll need to call the relevant queries so that the list of existing names is in scope when defining the form field schema. we can use the`.test()` method on the yup schema for the name field to add additional custom validation.

### LINKED ISSUE #157: Determine real requirements for form validation and add to form schemas
We had some placeholder string min/max limits that were incorrect and only there for testing purposes, but I'm removing those to fix #156.

We need to figure out the actual requirements our text fields need to meet for submitting CRs and implement those validation rules for our forms properly. Character limits, disallowed characters, etc.

mturley:
Resolves #157 
Resolves #128 

Implements guidance from @fdupont-redhat on how to validate our text fields. Also prevents users from creating providers/mappings/plans with a name that is already in use.

Still waiting on a couple of unknowns in https://github.com/konveyor/virt-ui/issues/157#issuecomment-717542095.

### DIFF
--- a/src/app/Mappings/components/AddEditMappingModal.tsx
+++ b/src/app/Mappings/components/AddEditMappingModal.tsx
@@ -33,6 +33,7 @@ import LoadingEmptyState from '@app/common/components/LoadingEmptyState';
 
 import './AddEditMappingModal.css';
 import MutationStatus from '@app/common/components/MutationStatus';
+import { dnsLabelNameSchema } from '@app/common/constants';
 
 interface IAddEditMappingModalProps {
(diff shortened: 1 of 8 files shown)


gildub (approved):
LGTM

mturley:
Rebased and resolved conflicts. @gildub can you re-approve?

gildub (approved):
Still looks good!

PR #165, opened 2020-10-27
set up a custom wrapper to handle axios responses from the k8s client…

ibolton336:
This PR handles errors from axios which are formatted differently than fetch errors. 

Any time we are fetching from the k8s client, we can pass in ```        queryFn: useAuthorizedK8sFetch(planResource),```  as the query function for useMockableQuery. This will handle the SSL issues & token expiry issues for requests against the clusterApi url.

### DIFF
--- a/src/app/queries/fetchHelpers.ts
+++ b/src/app/queries/fetchHelpers.ts
@@ -97,7 +97,7 @@ export const useAuthorizedFetch = <T>(url: string): QueryFunction<T> => {
   return () => authorizedFetch(url, fetchContext);
 };
 
-export const useAuthorizedK8sFetch = <T>(resourceType: any): QueryFunction<T> => {
+export const useAuthorizedK8sFetch = <T>(resourceType: KubeResource): QueryFunction<T> => {
   const fetchContext = useFetchContext();
   const client = useClientInstance();
   return () => authorizedK8sFetch(resourceType, fetchContext, client);
--- a/src/app/queries/fetchHelpers.ts
+++ b/src/app/queries/fetchHelpers.ts
@@ -3,6 +3,8 @@ import { INetworkContext, useNetworkContext } from '@app/common/context/NetworkC
 import { QueryFunction, MutateFunction } from 'react-query/types/core/types';
 import { useHistory } from 'react-router-dom';
 import { History, LocationState } from 'history';
+import { useClientInstance } from '@app/client/helpers';
+import { ClusterClient, KubeResource } from '@konveyor/lib-ui';
 
 interface IFetchContext {
   history: History<LocationState>;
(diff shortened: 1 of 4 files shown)


mturley:
@ibolton336 looking good. we should also use this for the `useMappingsQuery`

src/app/queries/fetchHelpers.ts
```diff
@@ -49,6 +51,41 @@ export const authorizedFetch = async <T>(
...
+    } else {
+      throw response;
+    }
+  } catch (error) {
+    checkExpiry(error, history);
+
+    const isAxiosSelfSignedCertError = (err) => {
+      // HACK: Doing our best to determine whether or not the
+      // error was produced due to a self signed cert error.
+      // It's an extremely barren object.
```
gildub, line 75: That's a detail, don't change it unless you agree with me.
It's difficult to figure out what's the link between a self signed cert error here and "Network Error".
Also not sure how the HACK header is relevant.

gildub (approved):
Just a small detail about comment but otherwise LGTM.

src/app/queries/fetchHelpers.ts
mturley: The problem here is a limitation of the way the cert errors are reported in this `err` object. They are literally just a TypeError with the message "Network Error", so it is a hacky solution that just assumes any error that looks like that is a cert error.

However, I actually don't think that message is consistent across Firefox and Chrome, @ibolton336 we might need different logic here. Can we check if it is an instance of TypeError like we do for the other helper or is axios obfuscating that?

src/app/queries/fetchHelpers.ts
```diff
@@ -60,6 +97,12 @@ export const useAuthorizedFetch = <T>(url: string): QueryFunction<T> => {
   return () => authorizedFetch(url, fetchContext);
 };
 
+export const useAuthorizedK8sFetch = <T>(resourceType: any): QueryFunction<T> => {
```
mturley, line 100: Also, instead of `any` here we should use `KubeResource`

src/app/queries/fetchHelpers.ts
ibolton336: Right @mturley This all stems from the self signed cert temporarily solution until we start using a different solution (server side requests & cert storage, etc. ).  @gildub If you have a better solution please let me know otherwise this is the best approach I can think of for handling the "Network error" message thrown from axios. We arent getting a TypeError from axios - it is completely different than what we receive from fetch. Axios is obfuscating it.

mturley: Double checking, I was wrong about the different message in Firefox, at least on mac it's working as written in both FF and Chrome. Yeah, I'm thinking more and more we should be proxying all the API calls through Express, this issue would go away.

mturley (approved):
LGTM. @gildub I think the issues you point out are valid but will go away when we address https://github.com/konveyor/virt-ui/issues/152, so I'm going to merge for now.

mturley:
@fdupont-redhat this should resolve the issue you were seeing with the plans table just "loading" forever and not handling the SSL error.

PR #167, opened 2020-10-29
Disable hosts page until we can support Host CR operations

mturley:
Context is in https://github.com/konveyor/virt-ui/issues/138. We will reintroduce this page when that issue is addressed.


gildub (approved):
LGTM

PR #171, opened 2020-10-30
Add delete functionality for providers, mappings and plans, simplify abstraction around kube-client, clean up types, fix bug in wizard

### LINKED ISSUE #170: Add support for removing/deleting existing providers, mappings and plans

mturley:
Resolves #170

* Replaces the `useAuthorizedK8sFetch` with a more general `useAuthorizedK8sRequest` which takes a `requestFn` that can be anything returning a promise, and handles relevant errors in one place.
* Uses this helper to create `useAuthorizedK8sClient` which is a drop-in replacement for `useClientInstance` that will use the `authorized` hook in each method to handle SSL errors and token expiry.
  * This client wrapper uses actual TS generics and introduces the `IKubeResponse<T>` type, which allows code handling the results of a client method call to be strongly typed. This involved some refactoring.
  * These new types are good candidates to be moved into the actual client itself, which would make some of the explicit type annotations I added unnecessary. https://github.com/konveyor/lib-ui/issues/33
* Uses this new client abstraction for all current k8s mutations and queries we have.
* Adds a mechanism to the `PollingContext` for temporarily speeding up the polling interval for a short time, which is useful to call right after a mutation which has a delayed response (creating providers and plans), so you don't have to wait the full 5 seconds to see the results.
  * I didn't put this in `useMockableMutation` because most of the time it is not necessary, so it's just an optional feature you can use in a particular mutation hook if it will help.
  * This should only be used on create and update mutations -- on a delete mutation, you should instead proactively remove the object from the queryCache like I do with providers, and if you do that, increasing the polling speed will make the object reappear before it is truly removed.
* Adds `useDeleteProviderMutation` which deletes both the provider and the associated secret.
  * Because this deletes a CR that will not immediately delete its associated objects in the inventory API, its onSuccess callback does not invalidate the useProvidersQuery and make it refetch. Instead, it filters out the matching object from the queryCache, so it appears to be deleted immediately. As long as it is really deleted in the inventory before the next refetch, it is a good experience. If for some reason a delete took longer than 5 seconds, technically this would result in the provider disappearing, reappearing, then disappearing again, but that seems unlikely.
* Adds `useDeleteMappingMutation` and `useDeletePlanMutation`.
* Wires up these mutations to the Remove actions in the kebab menus of the respective tables.
* Adds a common component `ConfirmDeleteModal` that is used in each of these tables (the dropdown action opens the modal, and submitting the modal calls the delete mutation).
* Consolidates the `VMwareProviderActionsDropdown` and `OpenShiftProviderActionsDropdown` components into one generic `ProviderActionsDropdown` because they were identical.
* Also fixes a bug caused by #164 where the "new mapping name" field (hidden until you check the box to save mappings for later in the plan wizard) was essentially required no matter what because the `matches()` regex validation was considering an empty string invalid.

Next up: edit.


gildub (approved):
LGTM

vconzola:
@mturley So this doesn't actually do anything yet, right? It just adds the hooks but doesn't actually do the deletes? 'Cause I don't see anything happen when when I try it in demo/preview mode. Also, these will all have confirmation dialogs, right?

vconzola:
@mturley Also, a note on terminology... if the action is Add (providers), the corresponding action should be Remove. If the action is Create (mappings and plans), the corresponding action should be Delete. We can take care of this when we do our detailed UI walkthrough next week.

mturley:
@vconzola it does actually work when you're not in preview mode, but I neglected to add confirm dialogs. I'll do that. Actually that gives me a convenient place for the message that the action is not available in preview mode, it didn't have anywhere to go otherwise.

mturley:
@vconzola ah ok. I'll fix the terminology for Remove/Delete in this PR while I'm in that code since I also need to add the confirmation.

ibolton336 (approved):
Looking good!

PR #173, opened 2020-10-30
Encode all fields of secrets in base64

fdupont-redhat:
In a secret, it's not possible to mix data formats between clear text and base64. This pull request encodes the `user` and `thumbprint` fields to be consistent with the other encoded fields.

Another option would be to use a `stringData` section instead of `data` and let all fields in clear text. The field would be encoded dynamically during secret creation.

Without that the secret cannot be created and the UI throws the following error:

20201030-ProviderSecretError

### DIFF
--- a/src/app/client/helpers.ts
+++ b/src/app/client/helpers.ts
@@ -72,13 +72,15 @@ export function convertFormValuesToSecret(
   } else {
     // default to vmware
     const vmwareValues = values as VMwareProviderFormValues;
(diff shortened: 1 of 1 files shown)


mturley (approved):
LGTM

mturley:
This is bizarre, I don't see this PR anywhere in https://travis-ci.com/github/konveyor/virt-ui/pull_requests but it did run for #174. Trying close/reopen one more time.

mturley:
There it goes.

PR #174, opened 2020-10-30
Move step logic to helpers along with step refactor

gildub:
Addresses issues with steps of VM migration details not to be displayed properly and not aligned with PipelineSummary.
Refactors Step component to be decoupled from VM status.


mturley:
Some rebase conflicts here @gildub

gildub:
@mturley, updated.

mturley (approved):
This looks great!

@gildub I don't want to hold up this PR, so we can follow up with another PR. But I think maybe we should use this same `.map()` and `getStepType` approach in the `PipelineSummary` too, instead of using `Chain`. Just so the logic is all consistent no matter what. Maybe that's not necessary?

PR #181, opened 2020-11-02
Increase width of status column in VM migration details table

mturley:
Based on feedback from @fdupont-redhat 
cc @vconzola


vconzola:
LGTM.

mturley:
Trivial change, visual approval from Vince, I think I'm safe to self-merge 😄

PR #182, opened 2020-11-03
Refactor Step and PipelineSummary to rely on common VM status logic

gildub:
VMStatusTable detailed step is `failed` when `pipeline.step` contains error, changed check.
Adds error structure to plans mock accordingly.
But more importantly refactor both Step and PipelineSummary to rely on `getStepType` in order for both to follow same path.

### DIFF
--- a/src/app/Plans/components/Step.tsx
+++ b/src/app/Plans/components/Step.tsx
@@ -26,7 +26,7 @@ const Step: React.FunctionComponent<IStepProps> = ({ type, error }: IStepProps) 
         className={spacing.mlSm}
         height="1em"
         width="1em"
-        color={successColor.value}
+        color={error ? dangerColor.value : successColor.value}
       />
     );
   }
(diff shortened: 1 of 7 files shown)


mturley (commented):
@gildub in remote dev mode, on mturley-test-1 and test-gildub1 I'm still seeing the failed steps showing green/successful. I think it's because they have `completed` properties. We should render error colors if there is an error property no matter whether the step is "completed".

mturley:
Also we should render the error `reasons` I think, maybe in the "State" column (current showing ValidationFailed, maybe it could be "ValidationFailed: Failed to initialize the source provider...". If there is more than one reason I guess we could just join them with commas?

gildub:
@mturley regarding `reasons` to be displayed makes sense. Have already validated with @vconzola?

mturley:
I suppose not. @vconzola do you think longer error messages on a pipeline step (could be 1 or 2 sentences) should be displayed in that State column, or should they be presented some other way?

vconzola:
@mturley @gildub Yes, I think the state column is the right place for the error message. Display what will fit with an ellipses at the end, then show the rest of the message in a popover.

mturley:
Ok. @gildub for the ellipses part see this example "Controlling text": https://www.patternfly.org/v4/components/table#controlling-text (`cellTransforms: [truncate]`)

mturley (approved):
This looks good! I'm still seeing some weirdness in the failed plans in Jeff's cluster, but I think that's because of issues on his end, it seems like you're properly rendering what is shown in the data.

PR #184, opened 2020-11-04
Address axe errors and warnings

gildub:
In order to cover as much as possible axe-core issues, see https://github.com/konveyor/virt-ui/issues/69:

* Replaces deprecated axe-core with @axe-core/react
* Bumps @patternfly/react-core and @patternfly/react-table


mturley:
@gildub just to be safe, any time we upgrade patternfly packages can we upgrade all of them? I'd like to especially make sure `react-core` and `react-styles` are always from the same release. Technically the peer dependency warnings should catch this in theory but I'm not sure if those are always perfect.

gildub:
@mturley, yeah that's right they should always go together, my bad I was adding them gradually trying to pinpoint some of axes issues. Anyway I've bumped all of them.

mturley:
No worries at all!

mturley (approved):
Thanks!

PR #189, opened 2020-11-05
Add support for editing plans, and retain selections in later steps when changing values in earlier steps

### LINKED ISSUE #104: Add support for editing existing migration plans
Note: In the new design for the mapping steps of the wizard (#103), the "Select an existing mapping" dropdown is entirely hidden when in editing mode, you just get a bare `MappingBuilder` pre-filled with the copy of the mapping saved in the plan.

mturley:
Resolves #104.

* Adds a route `/plans/:planName/edit` which renders the PlanWizard in addition to `/plans/create`.
* Adds a route match to PlanWizard to determine if it is rendered by the `/edit` route, and if so, grabs the `planBeingEdited`.
* Adds a hook `useEditingPrefillEffect` which will wait for queries to finish and then pre-fill the forms with values based on the `planBeingEdited`.
  * Because this happens asynchronously, the wizard was acting as if the user changed the form values that are being prefilled, so the effects which reset forms on dependent value changes will only run after prefilling is complete.
  * Because the prefilling hook requires some queries not otherwise required by the plan wizard, it keeps track of its aggregate query status and returns that to the wizard so it can continue to render a loading spinner until the prefilling is done.
  * The `window.setTimeout(() => { setIsDonePrefilling(true); }, 0);` is strange; without it, React will batch all the field changes and the `isDonePrefilling` change into one render, which means [this effect](https://github.com/konveyor/virt-ui/pull/189/files#diff-55b411ad333b40962073e7eb43e4f9db265aaada61377172a4e4b5e98384ec07R160) will run anyway even though we want it to *not* run during prefilling and then run if values change after prefilling. The `setTimeout` just makes the `setIsDonePrefilling` asynchronous: the function will return and allow React to batch all the new field values into a render before it handles the new `isDonePrefilling` value in the next render.
* Cleans up some text stuff in the wizard
* Adds a new `isPrefilled` field to the mapping forms and the filterVMs form, which is not visible to the user but is set to true initially if the wizard is opened for editing.
  * If a form is reset (because the user changed providers), `isPrefilled` will go back to false, so that form will behave as if it wasn't in editing mode (e.g. the dropdown for existing mappings comes back). Essentially, changing the providers on the general step means the entire plan needs to be replaced, so it mostly behaves the same as in create mode after that.
* When the wizard is in editing mode (and a step still has `isPrefilled` set to true), some things are rendered differently:
  * "Create" text is replaced with "Edit" / "Save" where appropriate
  * On the mapping steps, the dropdown to select an existing mapping is hidden.
  * On the filter VMs step, initially and when switching the tree type tabs, instead of clearing the selections, they are reset based on the planBeingEdited (the first tree item matching each selected VM is selected). That way, the filters start out in a state which will include the already-selected VMs when moving on to the next step.
* Prevents the user from changing the plan name when editing (limitation of patching CRs).
* Adds a mutation for patching the plan CR, and wires it up when submitting the wizard.
* Factors out a lot of initialization logic into helpers since it is reused for prefilling (e.g. finding missing mapping sources)

Some changes have been made that also affect creating new plans:
* When "Filter VMs" tree selections change, the subsequent forms (Select VMs and Mappings) are no longer cleared / reset. Similarly, when the "Select VMs" selections change, the Mappings forms are no longer cleared / reset. Instead, these steps can now retain their original values while adapting to the new context.
  * If the tree selections result in a list of VMs that does not include the already-selected VMs, those VMs will be included anyway at the top of the table. That way you can still see your entire selection regardless of the tree filters. This also means that you can select something in the tree, click Next, select some VMs, go back, select something different in the tree, click Next, and select some more VMs (use multiple passes of filtering when building your list of VMs). This is a corner case for sure, but it prevents the annoying result of having your list of VMs lost if you decided to change the filters at all.
  * If the selected VMs change, the set of required mapping sources could change with it. The mapping forms will now check to see if their current selections are missing any new required sources on mount, and add them. If any previously-required sources are no longer required and the user hadn't selected a target for them, they are removed. However, no mapping items with selected targets will be removed.
  * With these changes, the only thing that will clear selections in the wizard is changing providers. Everything else is malleable, you can hop backward and forward around the wizard changing whatever you like, and the other steps will adapt logically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "webpack-merge": "^4.1.4"
   },
   "dependencies": {
-    "@konveyor/lib-ui": "^1.9.0",
+    "@konveyor/lib-ui": "^1.10.5",
     "@patternfly/react-core": "^4.75.2",
     "@patternfly/react-icons": "^4.7.16",
     "@patternfly/react-styles": "^4.7.12",
--- a/src/app/Mappings/components/MappingBuilder/helpers.ts
+++ b/src/app/Mappings/components/MappingBuilder/helpers.ts
@@ -11,11 +11,14 @@ import {
   ICommonMapping,
   INetworkMapping,
   IStorageMapping,
+  IVMwareVM,
 } from '@app/queries/types';
 import { IMappingBuilderItem } from './MappingBuilder';
 import { getMappingSourceById, getMappingTargetByRef } from '../helpers';
 import { CLUSTER_API_VERSION, VIRT_META } from '@app/common/constants';
 import { nameAndNamespace } from '@app/queries/helpers';
+import { filterSourcesBySelectedVMs } from '@app/Plans/components/Wizard/helpers';
+import { IMappingResourcesResult } from '@app/queries';
 
 export const getBuilderItemsFromMappingItems = (
   items: MappingItem[] | null,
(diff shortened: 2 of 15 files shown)


ibolton336 (approved):
LGTM!

PR #195, opened 2020-11-06
Add support for editing mappings

### LINKED ISSUE #74: Add support for editing existing mappings
Depends on #73. Helper functions exist already for populating the MappingBuilder from API objects, but we'll need to change our mapping mutation hook to support updates.

mturley:
Resolves #74 .

Reuses a lot of the same approach from #189, just without the complications of dependent wizard steps. See notes there for some of the rationale that is generically applicable (e.g. the `setTimeout`, use of `isDonePrefilling`, etc)


gildub (approved):
LGTM

PR #196, opened 2020-11-09
Remove ES2020 syntax (Promise.allSettled) and replace with polyfill for browser support

mturley:
Currently we use `Promise.allSettled()` which is a new feature supported only in Chrome 76+ (and similar new versions of other browsers). Ilanit was testing in Chrome 74 and was unable to create a new provider because that mutation uses `allSettled`.

Initially TypeScript wouldn't let us use it, and we added `ES2020` to the `lib` section of `tsconfig` so it would be allowed. I thought that would mean it is transpiled away for us, but `ts-loader` handles the actual transpiling, not `tsc`, so effectively we were just telling our toolchain to ignore it.

This PR adds the `Q` library as a polyfill (same as we are using in mig-ui), and removes the `ES2020` configuration so that TypeScript will once again complain if we try to use these not-well-supported language features.


gildub (approved):
LGTM!

PR #197, opened 2020-11-09
VMware providers MA payload using href within Button

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/109

### DIFF
--- a/src/app/Providers/components/ProvidersTable/VMware/VMwareProvidersTable.tsx
+++ b/src/app/Providers/components/ProvidersTable/VMware/VMwareProvidersTable.tsx
@@ -63,10 +63,9 @@ const VMwareProvidersTable: React.FunctionComponent<IVMwareProvidersTableProps> 
     isEqual: (a, b) => a.name === b.name,
   });
 
-  const setInventoryDownloadURL = () =>
-    VIRT_META.inventoryPayloadApi +
-    '/api/v1/extract?providers=' +
-    selectedItems.map((provider) => provider.name).join();
+  const inventoryDownloadURL = `${
+    VIRT_META.inventoryPayloadApi
+  }/api/v1/extract?providers=${selectedItems.map((provider) => provider.name).join()}`;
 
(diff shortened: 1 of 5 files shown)


src/app/Providers/components/ProvidersTable/VMware/VMwareProvidersTable.tsx
```diff
@@ -63,6 +63,11 @@ const VMwareProvidersTable: React.FunctionComponent<IVMwareProvidersTableProps>
     isEqual: (a, b) => a.name === b.name,
   });
 
+  const setInventoryDownloadURL = () =>
```
mturley, line 66: I think we can just make this a string variable instead of a function that is only called once. Also, we should use the string template syntax and avoid concatenating strings with +.

```ts
const inventoryDownloadUrl = `${
  VIRT_META.inventoryPayloadApi
}/api/v1/extract?providers=${selectedItems.map((provider) => provider.name).join()}`;
```

mturley:
I think this implementation makes the most sense (at least for now).

src/app/Providers/components/ProvidersTable/VMware/VMwareProvidersTable.tsx
gildub: I've been going back and forth between string template and concatenating strings on that one! I'll use the former then.
Re string variable, yeah re-render is done at every user action.

mturley: Yeah, I try to never use + for string concat, but the reasons are a bit vague. Something about type coercion, maybe performance? https://softwareengineering.stackexchange.com/questions/90203/why-is-so-bad-for-concatenation

gildub: Thanks for the link.

mturley (approved):
LGTM!

PR #198, opened 2020-11-09
VMware providers MA payload download using React Query

gildub:
Initial approach to use React Query to handle inventory payload request.

https://github.com/konveyor/virt-ui/issues/109


src/app/queries/fetchHelpers.ts
```diff
@@ -32,7 +33,7 @@ export const authorizedFetch = async <T>(
         ...extraHeaders,
       },
     });
-    if (response.ok && response.json) {
+    if (!isBlob && response.ok && response.json) {
```
mturley, line 36: In this case if `isBlob` is true we'll always `throw`, right? don't we want to `return` in either case? Maybe something like this?

```
const responseBodyFn = !isBlob ? response.json : response.blob;
if (response.ok && responseBodyFn) {
  return responseBodyFn();
} else {
  throw response;
}
```

src/app/Providers/components/ProvidersTable/VMware/InventoryDownload.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { VIRT_META } from '@app/common/constants';
+import { IVMwareProvider } from '@app/queries/types';
+
+interface IInventoryDowloadProps {
+  providers: IVMwareProvider[];
+}
+
+const InventoryDownload: React.FunctionComponent<IInventoryDowloadProps> = ({ providers }) => {
+  const useDownload = () => {
+    return useQuery({
```
mturley, line 14: As you mentioned on Slack, we probably don't need react-query at all here, right? Since we're not catching it, all we need to do is have an async download function that will `const payloadBlob = await authorizedFetch(...);` and then write that to the file I think

src/app/Providers/components/ProvidersTable/VMware/VMwareProvidersTable.tsx
```diff
@@ -141,15 +143,14 @@ const VMwareProvidersTable: React.FunctionComponent<IVMwareProvidersTableProps>
     });
   });
 
+  const [isDownload, setDownload] = React.useReducer((isDownload) => !isDownload, false);
+
   return (
     <>
       <Level>
+        {isDownload ? <InventoryDownload providers={selectedItems} /> : null}
```
mturley, line 151: Yeah, and then we wouldn't need `<InventoryDownload>` at all, the download function could just be in the `onClick` of the button

src/app/queries/fetchHelpers.ts
gildub: Yeah, I discovered that afterwards, I was adding another condition but your approach is better. Thanks
I've changed it.

src/app/Providers/components/ProvidersTable/VMware/InventoryDownload.tsx
gildub: Right and that's why https://github.com/konveyor/virt-ui/pull/197 seems more straightforward.
Meanwhile to make is work with this approach that seems to be the way.

mturley: @gildub hmm, now that I think of it we would be able to get error handling out of react-query.

mturley:
Closing in favor of #197 , but we can reopen if necessary and I'll leave the branch in place.

PR #199, opened 2020-11-09
When editing a plan, keep only required mapping sources when VMs change

mturley:
Based on a discussion with @vconzola 

Previously, the mapping steps of the wizard kept all mapping items with selected targets no matter what changed. So you could have this scenario:

* User selects VMs A and B, which require mapping sources 1 and 2, and clicks Next
* User selects targets for both sources
* User goes back, deselects VM B, now only source 1 is required.
* When they proceed again, both sources still appear even though the second one is now irrelevant.

Now, this will not happen (source 2 will disappear, just as it would if the user hadn't selected a target for it). This behavior is true only when creating a new mapping from scratch or editing the one prefilled from a plan (if you select an existing mapping when making a new plan, the entire mapping will still be shown even if some of its sources are irrelevant, to prevent confusion).


vconzola (commented):
LGTM.

gildub (approved):
LGTM

PR #201, opened 2020-11-09
Show namespace on target networks to keep them unique

### LINKED ISSUE #200: Target networks need to be namespace-qualified
It's possible to have two target networks show up in the mapping builder with the same name, since the names only have to be unique within their namespace and we can have networks in multiple namespaces.

We should display the namespace along with the network (either for all of them, or just for the ambiguous ones). Thoughts, @vconzola ?

mturley:
Resolves #200. See discussion there for details.

Screen Shot 2020-11-09 at 2 28 34 PM

### DIFF
--- a/src/app/Mappings/components/MappingBuilder/MappingBuilder.css
+++ b/src/app/Mappings/components/MappingBuilder/MappingBuilder.css
@@ -18,3 +18,7 @@
 .mapping-item-select .pf-c-select__toggle .pf-c-select__toggle-clear {
   display: none;
 }
+
+.mapping-item-select .pf-c-select__menu {
+  width: 100%;
+}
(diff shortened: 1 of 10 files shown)


mturley:
Marked ready as per discussion in #200, rebased and resolved conflict.

vconzola (dismissed):
@mturley Looks great as long as the namespace / network string fits in the dropdown. How does PF handle strings that overrun the width of a dropdown, or do you need to handle that yourself?

mturley:
@vconzola the dropdown item just widens, but that's a good point, it looks like if it is so wide it goes past the edge of the modal / page, it gets cut off. So I need to figure out how to add truncate / ellipses / tooltip in a dropdown.

mturley:
@vconzola I added a component we can reuse for truncated text all over the UI, since I know how much you hate wrapping text 😄 here's how it looks in these dropdowns:

Screen Shot 2020-11-11 at 3 36 27 PM

Can you re-review?

vconzola (commented):
@mturley I don't see any networks with long names in the demo preview UI, but I trust you. :-) Thanks for this.

mturley:
If you go to create a network mapping, it's in the target network choices no matter which providers you choose.

vconzola:
ahh...that one was hidden below the fold - had to scroll to see it. looks good.

gildub (approved):
LGTM

PR #203, opened 2020-11-09
Decode VM names containing URI entities (e.g. '%2f' -> '/')

### LINKED ISSUE #202: VM names display slashes (/) as "%2f".
It seems like we need to run all the VM names through `decodeURIComponent`.

mturley:
Resolves #202 .

Example: in Jeff's cluster, there is a VM named `fdupont/test`, which was being displayed as `fdupont%2ftest` and will now display correctly.


gildub (commented):
Why not call directly decodeURIComponent(vm.name)?

mturley:
Closing in favor of https://github.com/konveyor/virt-controller/issues/75.

mturley:
> Why not call directly decodeURIComponent(vm.name)?

The only reason was just because we needed it in several places and I was trying to save some keystrokes / make sure we never use the VM name without it. Maybe was unnecessary.

PR #204, opened 2020-11-09
Add support for editing providers, prevent editing or removing host provider or providers with running migrations

### LINKED ISSUE #169: Add support for editing existing providers

### LINKED ISSUE #206: Change field label from "Hostname" to "Hostname or IP address" when adding a VMware provider

mturley:
Resolves #169.
Resolves #206.

* Adds an edit mode to the provider modal
  * Hides the secret/credential fields behind a "Replace credentials" checkbox
  * Since it doesn't need a query to resolve before it can prefill the form values, it just uses the prefill values as initial values in `useFormState` instead of having a special useEffect hook for prefilling like for mappings and plans
* Creates `usePatchProviderMutation` which will patch the provider and, if "replace credentials" was checked, also patch the associated secret
* Wires up the mutation to the modal
* In order to pass the method for opening the modal in edit mode down to the ProviderActionsDropdown (and also the plansQuery which is used to determine if there is a running migration, and is also needed on the ProvidersPage), I used a React Context provider to avoid several levels of prop drilling (ProvidersPage -> ProvidersTable -> [VMware|OpenShift]ProvidersTable -> ProviderActionsDropdown). This could technically also be used for the mapping and plan cases, but those have fewer levels to pass props down so it wouldn't help as much.
* Adds logic for blocking edit and remove operations when an associated migration is running, or if the provider does not have a `url` property (indicating that it is the host provider).
* Changes "Hostname" field label to "Hostname or IP address" (https://github.com/konveyor/virt-ui/issues/206)
* Fixes mock data (mock OpenShift providers didn't have a `type` field, and all mock providers had invalid names).


mturley:
Closing in favor of #197 (we can reopen if necessary and I'll leave the branch in place)

mturley:
Dammit, wrong issue.

mturley:
Need to hold off on merging this actually, something is wrong with the form validation.

mturley:
Ok, all set now.

vconzola:
@mturley Looks good except the button label to accept the changes should be Save instead of Edit. (Verifying this with Abi, but that's what I've always used.)

mturley:
Fixed. Thanks Vince. Not sure why I didn't use Save to begin with, I did that for mappings and plans. oops.

gildub (approved):
Quite an impressive change!
LGTM.

PR #214, opened 2020-11-13
Proxy backend APIs

gildub:
https://github.com/konveyor/virt-ui/issues/152

Depends upon https://github.com/konveyor/virt-ui/pull/209 to support inventory-payload latest.

### DIFF
--- a/deploy/server.js
+++ b/deploy/server.js
@@ -112,8 +112,8 @@ if (process.env['NODE_ENV'] === 'development') {
   };
 
   inventoryPayloadApiProxyOptions = {
-    target: virtMeta.inventoryPayloadApi,
     ...inventoryPayloadApiProxyOptions,
+    target: virtMeta.inventoryPayloadApi,
     logLevel: 'debug',
   };
 }
--- a/deploy/server.js
+++ b/deploy/server.js
@@ -72,46 +72,49 @@ if (process.env['DATA_SOURCE'] !== 'mock') {
 }
 
 let clusterApiProxyOptions = {
-  target: virtMeta.clusterApi,
+  target: 'https://api.openshift-apiserver.svc.cluster.local',
   changeOrigin: true,
   pathRewrite: {
     '^/cluster-api/': '/',
   },
+  secure: false,
 };
 
 let inventoryApiProxyOptions = {
-  target: virtMeta.inventoryApi,
+  target: 'http://inventory.openshift-migration.svc.cluster.local',
   changeOrigin: true,
   pathRewrite: {
     '^/inventory-api/': '/',
   },
+  secure: false,
 };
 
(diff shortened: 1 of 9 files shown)


deploy/server.js
```diff
@@ -69,7 +71,59 @@ if (process.env['DATA_SOURCE'] !== 'mock') {
   });
 }
 
-app.get('*', (req, res) => {
+let clusterApiProxyOptions = {
+  target: virtMeta.clusterApi,
```
mturley, line 75: For these targets, we should have conditions where in prod we use the internal endpoints described in https://github.com/konveyor/virt-ui/issues/152#issuecomment-726640172 instead of the values from `virtMeta`.

e.g.

```ts
let clusterApiProxyOptions = {
  target:
    process.env['NODE_ENV'] !== 'development'
      ? 'https://api.openshift-apiserver.svc.cluster.local/'
      : virtMeta.clusterApi,
  [...]
};
```

deploy/server.js
```diff
@@ -69,7 +71,59 @@ if (process.env['DATA_SOURCE'] !== 'mock') {
...
+let inventoryPayloadApiProxyOptions = {
+  target: virtMeta.inventoryPayloadApi,
+  changeOrigin: true,
+  pathRewrite: {
+    '^/inventory-payload-api/': '/',
+  },
+};
+
+if (process.env['NODE_ENV'] === 'development') {
+  clusterApiProxyOptions = {
```
mturley, line 99: Or I suppose you could just replace the `target` properties here since you already have a condition for dev mode.

mturley (changes requested):
@gildub aside from the comments below, the only thing you were missing was to change the UI code so it uses those new `/cluster-api`, `/inventory-api` and `/inventory-payload-api` URLs instead of the ones in VIRT_META!

I made this change locally and the proxy works flawlessly!

I opened a PR against your branch to make this change: https://github.com/gildub/virt-ui/pull/1

mturley:
To save time I'll go ahead and address the other feedback in https://github.com/gildub/virt-ui/pull/1.

mturley:
Ok, https://github.com/gildub/virt-ui/pull/1 is ready and addresses all of the above.

mturley:
We should probably also remove the `CertErrorPage` and all the logic for redirecting to it. That can be a follow-up PR though.

gildub:
@mturley, great, thanks.

deploy/server.js
gildub: Now that I realize that's the endpoint for the server it all fall in place.

mturley:
Whoops! Thanks for the fix.

mturley (approved):
LGTM!

PR #228, opened 2020-11-13
Remove CertErrorPage and deps

gildub:
https://github.com/konveyor/virt-ui/issues/227


gildub:
@mturley, I was thinking of going further and remove some of the code in `fetchHelpers.ts` but since we're still using SSL it's better to just leave as it is for now. What do you think?

mturley:
Well we're still using the self-signed SSL certs on the Node side, but if one of those requests came back with a cert error it wouldn't help to redirect the user to accept it in their browser (Node would still have a problem). I'm fine with removing it, but maybe we can do that in a separate PR to make it easier to revert if I'm wrong :)

src/app/queries/fetchHelpers.ts
```diff
@@ -95,7 +94,6 @@ export const authorizedK8sRequest = async <T>(
     if (isAxiosSelfSignedCertError(error)) {
       const url = `${VIRT_META.clusterApi}/.well-known/oauth-authorization-server`;
       setSelfSignedCertUrl(url);
```
mturley, line 96: Yeah, calling `setSelfSignedCertUrl` does nothing if we don't have the CertErrorPage. We should remove it from the FetchContext as well I think.

gildub:
@mturley, Ok I've kept removal of the use of setSelfsignedCertUrl to a separate PR.

mturley (approved):

PR #229, opened 2020-11-16
Ignore self-signed certs in remote dev mode

mturley:
In remote dev mode, we still need to ignore self-signed certs in order to allow the redirect to the login page.


gildub (approved):
LGTM

PR #230, opened 2020-11-16
HPM to use virtMeta and SSL for production

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/152

### DIFF
--- a/deploy/server.js
+++ b/deploy/server.js
@@ -77,7 +77,7 @@ let clusterApiProxyOptions = {
   pathRewrite: {
     '^/cluster-api/': '/',
   },
-  logLevel: process.env.DEBUG ? 'debug' : '',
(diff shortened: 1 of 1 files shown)


mturley (approved):
LGTM

mturley (changes requested):
Actually @gildub since this note is in the issue being resolved here, can we address this comment? https://github.com/konveyor/virt-ui/issues/152#issuecomment-728140834

Basically in the `let clusterApiProxyOptions` (not in the dev mode condition) we should set `logLevel: process.env['DEBUG'] ? 'debug' : 'info'`, same for all of them. And then we should probably add `export DEBUG=1` in the `start-dev.sh` script.

mturley (dismissed):

mturley (approved):

PR #232, opened 2020-11-16
Always use virtMeta.json for proxy URLs

fdupont-redhat:
The URLs are provided by virt-operator via virtMeta.json file, so the proxified URLs shouldn't be hard coded in the UI. This pull request makes the proxy always use virtMeta values.

It also enabled debug mode for the proxy if `DEBUG` environment variable is set to `*`, i.e. enables debug for proxy at the same time as Node.

It also sets the `secure` attribute to `true` for clusterApi, as the CA certificate is configured at Node level. Not sure it will works.

### DIFF
--- a/deploy/server.js
+++ b/deploy/server.js
@@ -72,16 +72,16 @@ if (process.env['DATA_SOURCE'] !== 'mock') {
 }
 
 let clusterApiProxyOptions = {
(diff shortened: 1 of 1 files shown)


fdupont-redhat:
Closing: (bad) duplicate of https://github.com/konveyor/virt-ui/pull/230.

PR #233, opened 2020-11-17
Fix mock dev mode: only set up proxy if not in mock mode

mturley:
@gildub we broke mock mode when we changed the `target` property of the proxy config to use virtMeta in server.js. When you're in mock mode, those virtMeta paths are undefined, so HPM complains:

```
Error: [HPM] Missing "target" option. Example: {target: "http://www.example.org"}
```

This only affects `yarn start:dev` and not the surge.sh preview links, because those don't run Express.

This PR puts all of the proxy config behind a condition so it only runs when not in mock mode.

### DIFF
--- a/deploy/server.js
+++ b/deploy/server.js
@@ -71,57 +71,59 @@ if (process.env['DATA_SOURCE'] !== 'mock') {
   });
 }
 
-let clusterApiProxyOptions = {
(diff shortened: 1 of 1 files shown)


gildub (approved):
LGTM

PR #234, opened 2020-11-17
Filter VMs step of wizard: only match immediate children of each selected tree node, and render partially-selected checkboxes where appropriate

### LINKED ISSUE #216: Tree view logic for Filter VMs needs to support indeterminate parent if not all children are selected.
Currently if you select a datacenter, then deselect one of its child clusters, the datacenter is still selected and all its contents still match the filter. Instead, the datacenter should go into indeterminate / partially-checked mode, and it should only include the VMs from the still-selected children. This should also be the case in the folder view, but we need to make sure that any VMs in the partially-selected parent folder that are not in the deselected child folders are still included.

mturley:
Resolves #216

This fixes the logic bug where deselecting children of a tree node would not exclude their VMs as long as the parent was still selected (it would just grab all VMs descended from all selected nodes). To test it, using the data in Jeff's environment you can try using the "by folders" tree and selecting the Workloads folder, then deselecting all its children so it is the only selected node. On master you will get 12 VMs, including those in the subfolders. On this branch you'll get only 6 VMs (only the ones directly in the Workloads folder).

This PR includes a temporary solution for rendering the indeterminate checkboxes using refs. We should switch to passing `null` for the `checked` property instead once it is supported in PF (https://github.com/patternfly/patternfly-react/pull/5150). Opened a tech debt issue to follow up on that: https://github.com/konveyor/virt-ui/issues/235


mturley:
Hmm. I may be wrong about the PF bug. Marking as draft until I determine whether I can also include that here.

gildub (dismissed):
Works as described.
Just need a rebase to be able to merge.

mturley:
Rebased @gildub , thanks

mturley:
There were no conflicts in the rebase even though it invalidated the review, so I'm going to merge based on @gildub 's earlier approval

PR #236, opened 2020-11-18
Remove unused setSelfsignedCertUrl

gildub:
Follows upon https://github.com/konveyor/virt-ui/pull/228.


mturley:
@gildub we can also entirely remove `selfSignedCertUrl` and `setSelfSignedCertUrl` everywhere they appear in the code. They originate in `NetworkContext.tsx` and they are also in `IFetchContext` and `useFetchContext` in `fetchHelpers.ts`.

gildub:
@mturley, yeah let's just remove it, if we need we can always tap into this.

mturley (approved):

Issue #210, opened 2020-11-12
Remove custom CSS for tree indentation once we bump PF version


mturley:
The tree indentation issue that we work around in https://github.com/konveyor/virt-ui/blob/master/src/app/Plans/components/Wizard/FilterVMsForm.css has been fixed upstream (https://github.com/patternfly/patternfly/issues/3532). Once that version of the CSS is included in a `@patternfly/react-core` release and we upgrade, we should remove our custom CSS.

mturley:
The core changes were released, but there are still significant indentation issues in the React version of the TreeView. Opened another upstream issue to address it: https://github.com/patternfly/patternfly-react/issues/5165

PR #241, opened 2020-11-19
Cleanup after PF bump, address tree selection corner case

### LINKED ISSUE #235: Replace indeterminate checkbox refs with null checked value once supported in PF
When this PF PR is merged: https://github.com/patternfly/patternfly-react/pull/5150
We should replace the `ref`s introduced in https://github.com/konveyor/virt-ui/pull/234 with `checked: null` when a checkbox should be partially checked.

mturley:
* Resolves #235 
* Tried to address #210, but there are still upstream issues to deal with. 
  * Changes our temporary tree view CSS so things still look correct (the tree CSS is still broken, just in a different way 🙃)
* Fixes a corner case I missed in #234.
  * If a node is not selected but all of its children are, it should still appear partially selected.


gildub (approved):
LGTM

PR #242, opened 2020-11-19
Vm concern handle of migration analytics out-of-sync revision case

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/238

### DIFF
--- a/src/app/Plans/components/Wizard/VMConcernsIcon.tsx
+++ b/src/app/Plans/components/Wizard/VMConcernsIcon.tsx
@@ -25,8 +25,10 @@ const VMConcernsIcon: React.FunctionComponent<IVMConcernsIconProps> = ({
   if (worstConcern.severity === 'Warning') {
     return <StatusIcon status={StatusType.Warning} label="Warning" />;
   }
-  if (worstConcern.severity === 'Advisory' || worstConcern.severity === 'Info') {
-    // TODO we should add an Info status type to StatusIcon
+  if (worstConcern.severity === 'Info') {
+    return <StatusIcon status={StatusType.Info} label="Info" />;
+  }
+  if (worstConcern.severity === 'Advisory') {
     return (
       <>
         <InfoCircleIcon />
--- a/src/app/Plans/components/Wizard/VMConcernsIcon.tsx
+++ b/src/app/Plans/components/Wizard/VMConcernsIcon.tsx
@@ -2,8 +2,6 @@ import * as React from 'react';
 import { IVMwareVM } from '@app/queries/types';
 import { StatusIcon, StatusType } from '@konveyor/lib-ui';
 import { InfoCircleIcon } from '@patternfly/react-icons';
-import { Flex, FlexItem, Spinner } from '@patternfly/react-core';
-import './VMConcernsIcon.css';
 
 import { getMostSevereVMConcern } from './helpers';
 interface IVMConcernsIconProps {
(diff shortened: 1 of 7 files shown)


mturley:
@gildub , based on a discussion with @vconzola I think we actually want this to use the spinner icon (loading status) with the text "Analyzing" instead of unknown. Also, we should add a condition for this in the `VMConcernsDescription` that displays 
"Completing migration analysis. This might take a few minutes." and hides the list of concerns (since it will be out of date in this state).

src/app/Plans/components/Wizard/VMConcernsIcon.tsx
```diff
@@ -2,15 +2,28 @@ import * as React from 'react';
...
 interface IVMConcernsIconProps {
   vm: IVMwareVM;
 }
 
 const VMConcernsIcon: React.FunctionComponent<IVMConcernsIconProps> = ({
   vm,
 }: IVMConcernsIconProps) => {
+  if (vm.revisionAnalyzed < vm.revision) {
+    return (
+      <Flex spaceItems={{ default: 'spaceItemsSm' }}>
```
mturley, line 18: Can't we just use a `<StatusIcon status={StatusType.Loading}>` here?

src/app/Plans/components/Wizard/VMConcernsIcon.tsx
```diff
@@ -2,15 +2,28 @@ import * as React from 'react';
...
+        <FlexItem>
+          <Spinner aria-valuetext="Analysing" size="sm" className="spinner" />
+        </FlexItem>
+        <FlexItem>Analysing</FlexItem>
```
mturley, line 22: @vconzola should we use "Analysing" or "Analyzing" with a z? I think with a z is the American spelling and everywhere else in the world spells it with an s.

src/app/Providers/components/CloudAnalyticsInfoAlert.tsx
```diff
@@ -9,7 +9,11 @@ const CloudAnalyticsInfoAlert: React.FunctionComponent = () => {
     LocalStorageKey.isProvidersPageMAAlertHidden
   );
 
-  const link = <a href={CLOUD_MA_LINK.href}>{CLOUD_MA_LINK.label}</a>;
```
mturley, line 12: Looks like this change carried over from your other PR by accident, need to rebase.

src/app/queries/types/vms.types.ts
```diff
@@ -7,11 +7,12 @@ export interface IVMwareVMDisk {
 
 export interface IVMwareVMConcern {
   name: string;
-  severity: 'Warning' | 'Critical' | 'Advisory' | 'Info'; // TODO not sure if these are final
+  severity: 'Warning' | 'Critical' | 'Advisory' | 'Info' | 'Unknown';
```
mturley, line 10: I don't think `'Unknown'` is a valid severity value in the API, but I'm not sure. We're not handling it above, either way, should probably remove it

vconzola:
Use the American English spelling (just like we use Canceled instead of Cancelled). Localization can change/fix it if necessary.

src/app/Plans/components/Wizard/VMConcernsIcon.tsx
gildub: Of course we can!

mturley (approved):

PR #243, opened 2020-11-19
Open separate tab for cloud analytics link

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/226

### DIFF
--- a/src/app/Providers/components/CloudAnalyticsInfoAlert.tsx
+++ b/src/app/Providers/components/CloudAnalyticsInfoAlert.tsx
@@ -9,7 +9,11 @@ const CloudAnalyticsInfoAlert: React.FunctionComponent = () => {
     LocalStorageKey.isProvidersPageMAAlertHidden
   );
 
-  const link = <a href={CLOUD_MA_LINK.href}>{CLOUD_MA_LINK.label}</a>;
+  const link = (
+    <a href={CLOUD_MA_LINK.href} target="_blank" rel="noreferrer">
+      {CLOUD_MA_LINK.label}
+    </a>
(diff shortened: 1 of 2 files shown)


gildub:
@mturley, for the navigation part it turns out that NavItem doesn't has a way to pass target attribute.
So one solution is to use <a href> and tweak css props to make it look like a NavItem.
What do you think?

mturley:
@gildub I think we can just remove the `to` prop from that `NavItem` and render a normal `<a href="..." target="...">` as a child of it. Our other NavItems render `<NavLink>` children (from react-router-dom) which end up rendering `<a>` elements as well. I think the `to` prop is just a shorthand for this.

gildub:
@mturley, Using an anchor element within NavItem doesn't play very well. I used onClick instead.

mturley:
Ok, that works.

src/app/Providers/components/CloudAnalyticsInfoAlert.tsx
```diff
@@ -9,7 +9,11 @@ const CloudAnalyticsInfoAlert: React.FunctionComponent = () => {
     LocalStorageKey.isProvidersPageMAAlertHidden
   );
 
-  const link = <a href={CLOUD_MA_LINK.href}>{CLOUD_MA_LINK.label}</a>;
+  const link = (
+    <a href={CLOUD_MA_LINK.href} target="_blank" rel="noreferrer">
```
mturley, line 13: Good call on `rel="noreferrer"`, I always forget that.

mturley (approved):

PR #244, opened 2020-11-19
Restore VMware provider hosts page, add host network selection

### LINKED ISSUE #138: Handle CRUD for Host CR (getting and setting IP address / network for a vmware host) and reintroduce Hosts table in providers page
Related to #85

mturley:
Resolves #138

* Restores our disabled Hosts page linked under the VMware providers table
* Fixes existing TS types for inventory hosts, includes new networking details
* Adds new TS type for the Host CR (`IHostConfig`)
* Adds mock data for these new types
* Updates existing SelectNetworkModal to hide the credential fields if you choose to reuse provider credentials (true by default)
* Adds a `useHostConfigsQuery` to load Host CRs, correlate them with host objects and display selected networks if any
* Adds a `useConfigureHostsMutation` to handle form submission and create/patch hosts/secrets as necessary
  * Looks for existing config CRs for each selected host
  * If you're passing a new secret, looks for any existing secrets present on the existing configs, and:
    * If at least one secret exists, the first one is patched and then used for all hosts
    * If not, a new secret is created and used for all hosts
  * If you're not passing a new secret, passes `{}` for the secret on all hosts (removes any existing associated secrets)
  * For each selected host, either patches or creates a host CR depending on whether there is already a matching one
  * If any of the above fails, the mutation will fail and report the first caught error (no rollback)
* Wires up the query and mutation to the table and modal

### DIFF
--- a/src/app/Providers/components/VMwareProviderHostsTable/SelectNetworkModal.tsx
+++ b/src/app/Providers/components/VMwareProviderHostsTable/SelectNetworkModal.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import * as yup from 'yup';
-import { Modal, Button, Form, FormGroup } from '@patternfly/react-core';
+import { Modal, Button, Form, FormGroup, Checkbox } from '@patternfly/react-core';
+import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import {
   useFormState,
   useFormField,
@@ -23,14 +24,23 @@ const SelectNetworkModal: React.FunctionComponent<ISelectNetworkModalProps> = ({
   selectedHosts,
   onClose,
 }: ISelectNetworkModalProps) => {
+  const isReusingCredentials = useFormField(true, yup.boolean().required());
+  const usernameSchema = yup.string().max(320).label('Host admin username');
+  const passwordSchema = yup.string().max(256).label('Host admin password');
   const form = useFormState({
-    // TODO add a checkbox for reusing the vsphere creds and hiding these fields
-    adminUserId: useFormField('', yup.string().max(320).label('User Id').required()),
-    adminPassword: useFormField('', yup.string().max(256).label('Password').required()),
     selectedNetworkAdapter: useFormField<IHostNetworkAdapter | null>(
       null,
       yup.mixed<IHostNetworkAdapter>().label('Host network').required()
     ),
+    adminUsername: useFormField(
+      '',
+      isReusingCredentials.value ? usernameSchema : usernameSchema.required()
+    ),
+    adminPassword: useFormField(
+      '',
+      isReusingCredentials.value ? passwordSchema : passwordSchema.required()
+    ),
+    isReusingCredentials,
   });
 
   const commonNetworkAdapters: IHostNetworkAdapter[] = selectedHosts[0].networkAdapters.filter(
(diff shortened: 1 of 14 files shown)


mturley:
@vconzola the code to submit the modal isn't done yet but this is ready for your visual review.

vconzola:
@mturley Did you tell me that the UI has no way of knowing what's the default network? If so, is there any way Jeff can get that information? Because the interaction is kinda weird without it. The user is sort of picking blindly from a list of networks without knowing what's currently selected.

mturley:
@vconzola I think that's the case, or at least I thought we had talked about adding that post-beta. The problem is that the inventory API has no way of knowing, because the default is decided by the controller when the migration starts (iirc). I'll ask Jeff.

mturley:
@gildub I think this is ready to go, but I'm actually not able to test it right now, for some reason my remote dev environment has started giving 500 errors when trying to proxy to Jeff's environment (directly using the env works fine). Not sure what the deal is there.

gildub:
@mturley, I confirm something's wrong with Jeff environment. I'm facing same behavior ATM using master. Will ask for a reset and come back to it later today.

mturley:
Thanks @gildub. Once you can test this, feel free to ping me if you need any quick changes. I know I'm on PTO but I really want to get this in.

One thing I'm not sure about is whether passing `{}` for the secret in a Host config CR will actually have the effect I want (removes any existing secret, results in no secret property at all on the CR, not an empty object). Might have to pass null or something instead. Jeff wasn't sure

gildub:
@mturley, Jeff has aligned his lab per TLS recent changes in the backend.

gildub (approved):
LGTM

gildub (changes requested):
Actually after another test I realized there is an issue with multiple hosts selection which give an empty network list.

gildub:
@vconzola, displaying the name and the IP address of each network interface under the column "Network for migration data transfer" feels weird (at least to me). Why not just call the column "Physical network interface name and IP address" or equivalent? The main reason here is because those IP address are host address and not network ranges.

src/app/Providers/components/VMwareProviderHostsTable/SelectNetworkModal.tsx
```diff
@@ -10,63 +10,85 @@ import {
...
-    alert('TODO');
-    if (form.isValid) {
-      console.log('TODO: submit form!', form.values);
-      onClose();
-    }
-  };
+  const commonNetworkAdapters: IHostNetworkAdapter[] = selectedHosts[0].networkAdapters.filter(
+    ({ name, ipAddress }) =>
+      selectedHosts.every((host) =>
+        host.networkAdapters.some((na) => na.name === name && na.ipAddress === ipAddress)
```
gildub, line 83: I think we need to match only the IP addresses for common network adapters..
Because if for some reason the operator did a typo or entered a different name to that interface then we won't have a match although VMs would be sharing the same Physical IP address.

gildub: Per our discussion, we're using only IP addresses to match common adapters.

gildub (approved):
LGTM

mturley:
I think this is good to merge, and I'll revisit it when I'm back to double check if we need to open any more follow-up issues. @gildub would you mind opening one about that text change?

PR #245, opened 2020-11-20
Check for pipeline step empty description

gildub:
Besides being more robust it also allow older migration data might not have the description field yet.


gildub:
Something went wrong with this one. Closing and adding a new one.

PR #247, opened 2020-11-20
Use locale for mig times and display with ellipses

gildub:
Replaces `formatTimeStamp()` to use Date automated conversion to locale.

times-tz

Resolves https://github.com/konveyor/virt-ui/issues/185

### DIFF
--- a/src/app/common/helpers.ts
+++ b/src/app/common/helpers.ts
@@ -1,11 +1,13 @@
 import dayjs from 'dayjs';
 import advancedFormat from 'dayjs-ext/plugin/advancedFormat';
-import timeZone from 'dayjs-ext/plugin/timeZone';
+import utc from 'dayjs/plugin/utc';
+import timezone from 'dayjs/plugin/timezone';
 import { StatusCategoryType, PlanStatusAPIType, StepType } from '@app/common/constants';
 import { IStatusCondition, IStep, IVMStatus } from '@app/queries/types';
 
+dayjs.extend(utc);
+dayjs.extend(timezone);
 dayjs.extend(advancedFormat);
-dayjs.extend(timeZone);
 
 export const hasCondition = (conditions: IStatusCondition[], type: string): boolean => {
   return !!conditions.find((condition) => condition.type === type);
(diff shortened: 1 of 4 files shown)


mturley:
We just need to add a `z` to the end of the format string, which we need the `advancedFormat` dayjs plugin for: https://day.js.org/docs/en/plugin/advanced-format

gildub:
@mturley, not working with `z` but I might have missed something.

mturley:
@gildub I think you also need to extend the `dayjs/plugin/advancedFormat` plugin

src/app/common/helpers.ts
```diff
@@ -1,7 +1,12 @@
 import dayjs from 'dayjs';
+import advancedFormat from 'dayjs-ext/plugin/advancedFormat';
+import timeZone from 'dayjs-ext/plugin/timeZone';
```
mturley, line 3: @gildub based on the example [here](https://day.js.org/docs/en/plugin/timezone) I think the problem is that you are using a capital Z in `timeZone` and instead you need `timezone`.

gildub: Thanks. Meanwhile still not working, just showing z character.

mturley: That's really weird. I'll take a look when I get a chance. I'm also happy to use something other than dayjs in this repo if we want to switch, we don't use it for much. Someone else in UXD recommended https://date-fns.org/

src/app/common/helpers.ts
```diff
@@ -1,7 +1,14 @@
 import dayjs from 'dayjs';
+import advancedFormat from 'dayjs-ext/plugin/advancedFormat';
```
mturley, line 2: @gildub sorry I missed this before, you were importing from the wrong package here. I made this change locally and it works.

```suggestion
import advancedFormat from 'dayjs/plugin/advancedFormat';
```

gildub: Thank you, I've mended it.

gildub:
@vconzola, please review.

mturley (approved):

mturley:
@vconzola I think this matches what you already described, I'll go ahead and merge, if you need changes we can follow up.

vconzola:
Yeah, this looks good.

PR #258, opened 2020-11-24
Bump node version for Travis

gildub:
Travis is not sending back results to GH. So trying this to see if that makes any changes.


gildub:
This is now obsolete after https://github.com/konveyor/virt-ui/pull/259 merge.

PR #263, opened 2020-11-25
VM Concern severity: Removes Info not avail in backend and uses select filter in SelectVMs page

gildub:
- Remove VM Concern severity Info not avail in backend
- Select VMs now uses select filter for migration analytics

Resolves https://github.com/konveyor/virt-ui/issues/126, resolves https://github.com/konveyor/virt-ui/issues/219


mturley (approved):
LGTM

Issue #146, opened 2020-10-22
Add filter for status column of VMMigrationDetails


gildub:
Following with https://github.com/konveyor/virt-ui/pull/142 adding a filter to the status column could be done post beta version.

gildub:
@vconzola, my concern is in the case of a running migration we don't display that state explicitly in status line above the pipeline summary, and therefore looks awkward to me to filter on something we don't display. But I suppose that could be documented.
What do you think?

![status](https://user-images.githubusercontent.com/1901741/96847921-0ba36c80-1454-11eb-855e-d4a8ba9d7dbb.png)

vconzola:
@gildub I wasn't thinking anything that complicated. :-) was just hoping to filter the status by Not started, In progress, and Completed so users can see which VMs are still migrating during plan execution.

PR #265, opened 2020-11-26
VMMigrationDetails: Add select filter for Status column

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/146

### DIFF
--- a/src/app/Plans/components/VMMigrationDetails.tsx
+++ b/src/app/Plans/components/VMMigrationDetails.tsx
@@ -129,7 +129,6 @@ const VMMigrationDetails: React.FunctionComponent = () => {
       title: 'Status',
       type: FilterType.select,
       selectOptions: [
-        { key: 'Ok', value: 'Ok' },
         { key: 'Completed', value: 'Completed' },
         { key: 'Not Started', value: 'Not Started' },
         { key: 'On Error', value: 'On Error' },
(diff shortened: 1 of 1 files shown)


vconzola:
@gildub Let's hold off on this until we have a chance to talk about it. I'm not sure we're talking about the same page where we want select filtering.

mturley:
@vconzola was there another one? I seem to remember us talking about putting it on this page (#146)

vconzola:
@mturley I thought we were talking about the Plans page where we have the status labels already in the table rows and the user could easily select Not started , Running / In-progress, Failed, or Completed from a dropdown. Ideally we'd have it on both pages and I think the same filter selection list would work in both places.

mturley:
@vconzola I think maybe we talked about putting it in both places and forgot to open an issue for the plans page. Opened an issue: https://github.com/konveyor/virt-ui/issues/276

For this PR, if you go to the details of a plan (list of VMs) the Status filter there has a dropdown now. e.g. http://konveyor-virt-ui-pr-265-preview.surge.sh/plans/plantest-3

vconzola:
@gildub @mturley One of the Status dropdown options is "OK"? What does this mean and when would a migration have status "OK"? I understand Completed, Not Started, On error, and In progress. But OK doesn't seem to fit into the lifecycle.

mturley:
Good point, @gildub it looks like you're using "OK" as a catch-all when the status is unknown. I think we can just exclude it.

gildub:
@vconzola, @mturley, good catch, "Ok" was a left over, absolutely not necessary. Thanks

mturley (approved):

PR #270, opened 2020-11-26
Adjust PipelineSummary title display

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/269

An error during the migration is validated by the presence of error data along with the reasons and the phase where it happened.

After discussing with @fdupont-redhat the issue, now displaying "Running" when the migration is executing instead of no title.

### DIFF
--- a/src/app/common/components/PipelineSummary.tsx
+++ b/src/app/common/components/PipelineSummary.tsx
@@ -35,10 +35,9 @@ export const getPipelineSummaryTitle = (status: IVMStatus): string => {
     return 'Complete';
   }
   if (status.started && !status.completed) {
-    let title: string;
-    title = status.error?.phase ? `Error - ` : '';
+    let title = status.error ? 'Error' : 'Running';
     if (currentStep?.description) {
-      title = `${title}${[currentStep.description.replace(/\.$/, '')]}`;
+      title = `${title} - ${currentStep.description}`;
     }
     return title;
   }
--- a/src/app/queries/mocks/plans.mock.ts
+++ b/src/app/queries/mocks/plans.mock.ts
@@ -60,10 +60,6 @@ if (process.env.NODE_ENV === 'test' || process.env.DATA_SOURCE === 'mock') {
     ],
     phase: 'Mock VM Phase',
     started: '2020-10-10T14:04:10Z',
-    error: {
-      phase: '',
-      reasons: [''],
-    },
   };
 
   const vmStatus2: IVMStatus = {
@@ -102,10 +98,6 @@ if (process.env.NODE_ENV === 'test' || process.env.DATA_SOURCE === 'mock') {
     ],
     phase: 'Mock VM Phase',
     started: '2020-10-10T14:04:10Z',
-    error: {
-      phase: '',
-      reasons: [''],
-    },
   };
 
   const vmStatus3: IVMStatus = {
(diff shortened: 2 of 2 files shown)


gildub:
@vconzola, could you please review. The last version of the PipelineSummary you recommended not to display "Running" in the title to leave more space for displaying the current step. Meanwhile we do display the step description when on error so we have the room to do it when executing.

vconzola (commented):
@gildub I'm fine with adding it if there's room for it. Can you adjust the line spacing when the description wraps? It looks kinda funky now. Also, if the description means there will be more than two lines of text I want to truncate it with an ellipses. Maybe these formatting changes should be a separate issue?

src/app/common/components/PipelineSummary.tsx
```diff
@@ -35,10 +35,9 @@ export const getPipelineSummaryTitle = (status: IVMStatus): string => {
     return 'Complete';
   }
   if (status.started && !status.completed) {
-    let title: string;
-    title = status.error?.phase ? `Error - ` : '';
+    let title = status.error ? 'Error' : 'Running';
```
mturley, line 38: @gildub was it intentional to remove the part of the condition that checks for a truthy `error?.phase` property?

The in-progress migrations in the mock data have bare error objects:
```ts
    error: {
      phase: '',
      reasons: [''],
    },
```

I'm not sure if that's true of the real data, but at least in mock mode we're getting "Error -" in front of non-error statuses because it's just checking if there is an `error` object at all. We should check whether there can be a bare error object like this in real data, and if so put back the `?.phase` part of the condition. If not we should remove those bare errors from our mock data.

mturley (changes requested):

gildub:
@vconzola, from our recent experience with https://github.com/konveyor/virt-ui/issues/267 and to avoid similar issue again, what about we display "Error" and the content as a popover?

src/app/common/components/PipelineSummary.tsx
gildub: @mturley, you're right, the change is because there are no error struct unless an error occurred.
I've fixed the mock data and added a clarification in the description above.

mturley (approved):

mturley:
Merging -- @vconzola if you want us to make the change @gildub is suggesting in his comment above we can open a separate issue

vconzola:
@mturley @gildub The intent of the progress reporting design is that it work the same way that it does in MTC, where a migration is broken into a sequence of steps and each step includes a sequence of phases. In the pipeline view, the current step name is displayed. When a user drills down, all of the steps are shown with the latest phase description (or error message) displayed for each step. Is there a similar taxonomy of steps and phases for MTV migrations? If not, it's something we should work toward post beta. I've never seen what progress/error information the UI receives from the backend. Designing in the dark, as usual.

mturley:
@vconzola the backend provides data in a very similar model: each VM has a pipeline with a list of steps, and each step has a set of sub-tasks (correlated to the phases). We show the current step name in the pipeline summary just like you describe, this PR just adds a prefix of "Running" or "Error" in front of it. Drilling down, There is a list of steps, each with the current phase description or error, just like you describe. We don't show the other sub-tasks in the UI currently because that would be a table-in-a-table-in-a-table.

Is there something missing from that that you're not seeing in the preview? It sounds like the models of the 2 apps are pretty much the same.

vconzola:
@mturley What's missing in the preview is the phase description or error is not shown. It just says, "Mock step phase". I assume that's where the phase description or error will be shown? It also bothers me that appending "Running" or "Error" to the step name above the pipeline causes the text to wrap. You know how i feel about that.  :-)

mturley:
@vconzola ah I see the confusion. The data we get from the backend has a phase name, and if an error has occured in that phase, an error description. There is no description available for a phase that is running normally, I believe. If you look at a pipeline with a failed step, you can see that the phase name becomes a link and the error description is shown in a popover:

Screen Shot 2020-12-01 at 2 09 24 PM

As for the step name wrapping to 2 lines, yeah, that sucks. I'll open a separate issue for it.

PR #271, opened 2020-11-27
Add Restart button for plan migration

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/268

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1901846

### DIFF
--- a/src/app/Plans/components/PlansTable.tsx
+++ b/src/app/Plans/components/PlansTable.tsx
@@ -146,6 +146,7 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
   enum ActionButtonType {
     Start = 'Start',
     Cancel = 'Cancel',
+    Restart = 'Restart',
   }
 
   currentPageItems.forEach((plan: IPlan) => {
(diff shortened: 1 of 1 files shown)


vconzola (commented):
@gildub Does this PR dod anything more than add the Restart button to the UI? I think you and I and @mturley should talk about what other changes need to be made before this is merged.

mturley:
@vconzola reading the code here, it adds the Restart button which will create a new Migration CR, basically it does exactly the same thing as the original Start button.

@gildub @fdupont-redhat when we are restarting a failed migration, do we need to also delete the previous Migration CR? Or is it sufficient to just create a new one?

vconzola:
@mturley If it's creating a new migration CR does that mean we have no visibility into the existing migration CR in order to show status, etc. Ideally, we want the plan status to "pick up" from where it left off before the failure. So if 2 of 5 VMs had been successfully migrated it would show that and start from there. And the the VM details page would show the already migrated VMs as successfully completed.

mturley:
@vconzola as far as I understand, we don't have that capability, this is essentially starting the migration over from scratch. I could be wrong about that. @fdupont-redhat @jortel do you know the answer? Is it possible to pick up where a failed migration left off, and if so do we need to do something special to kick that off other than creating a new Migration CR?

mturley:
@vconzola actually, it turns out I was wrong! I just confirmed with Jeff that creating a new Migration CR will do just what you're asking for. There's no need to delete the old one, the controller will handle all that. So it looks like this is all we need.

mturley:
Were you talking about needing any other changes aside from that?

mturley (approved):

vconzola:
Nope. That was it. Sounds good if we can pick up from where the failed migration left off.

PR #272, opened 2020-11-27
Cascade delete migrations when removing migration plan

gildub:
Note: This issue was initially targeted for post Beta.

Resolves https://github.com/konveyor/virt-ui/issues/190

### DIFF
--- a/src/app/queries/types/common.types.ts
+++ b/src/app/queries/types/common.types.ts
@@ -4,7 +4,7 @@ export interface IMetaTypeMeta {
   kind: string;
 }
 
-interface objectReference {
+interface IObjectReference {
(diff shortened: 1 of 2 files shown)


src/app/queries/types/common.types.ts
```diff
@@ -4,6 +4,13 @@ export interface IMetaTypeMeta {
   kind: string;
 }
 
+interface objectReference {
```
mturley, line 7: Just for consistency, can we stick with the capitalized interface names starting with `I`?
`IObjectReference`

gildub: Of course!

mturley (approved):

PR #273, opened 2020-11-30
Use Popover instead of ellipsis for step content

gildub:
Resolves https://github.com/konveyor/virt-ui/issues/267

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1901967


mturley:
@vconzola does this look good to you? if there are error details on a pipeline step that step name is now a link that opens the error in a popover.

vconzola:
after talking with @mturley, this looks like a good approach.

mturley (approved):

PR #278, opened 2020-12-01
Disable Select migration network when needed

gildub:
- Lift up `commonNetworkAdapters` logic to `VMwareProviderHostsTable` to enable/disable "select migration network" button if there are or not common adapters between hosts.
- Display a Tooltip messages when no selection has been made:
select-message1
- Display a Tooltip message when selected hosts offers no adapters in common:
select-message2

Resolves https://github.com/konveyor/virt-ui/issues/261

### DIFF
--- a/src/app/Providers/components/VMwareProviderHostsTable/VMwareProviderHostsTable.tsx
+++ b/src/app/Providers/components/VMwareProviderHostsTable/VMwareProviderHostsTable.tsx
@@ -63,7 +63,7 @@ const VMwareProviderHostsTable: React.FunctionComponent<IVMwareProviderHostsTabl
       hosts.every((host) => host.networkAdapters.some((na) => na.ipAddress === ipAddress))
     );
 
-  const isCommonNetworkAdapters = () => {
+  const hasCommonNetworkAdapters = () => {
     if (selectedItems.length > 0) {
       return commonNetworkAdapters(selectedItems).length > 0 ? true : false;
     }
@@ -87,7 +87,7 @@ const VMwareProviderHostsTable: React.FunctionComponent<IVMwareProviderHostsTabl
       <Level>
         <LevelItem>
           <ConditionalTooltip
-            isTooltipEnabled={!isCommonNetworkAdapters()}
+            isTooltipEnabled={!hasCommonNetworkAdapters()}
             content={
               selectedItems.length === 0
                 ? 'Select at least one host'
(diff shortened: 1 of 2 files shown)


gildub:
@vconzola, I've used initial text for the tooltips, please let us know if we require more specific messages.

mturley (dismissed):
LGTM pending @vconzola 's review

vconzola:
LGTM.

src/app/Providers/components/VMwareProviderHostsTable/VMwareProviderHostsTable.tsx
```diff
@@ -57,6 +58,18 @@ const VMwareProviderHostsTable: React.FunctionComponent<IVMwareProviderHostsTabl
...
+      hosts.every((host) => host.networkAdapters.some((na) => na.ipAddress === ipAddress))
+    );
+
+  const hasCommonNetworkAdapters = () => {
+    if (selectedItems.length > 0) {
+      return commonNetworkAdapters(selectedItems).length > 0 ? true : false;
+    }
+    return false;
+  };
+
```
mturley, line 72: Sorry to nitpick after my prior approval @gildub , but I realized this could be simpler. Remember that each time the selected items change, the whole component re-renders and these functions are reinitialized. They are always called with the same values and will return the same thing in all calls of a given render. So they don't need to be functions. It could just be one variable:

```ts
const commonNetworkAdapters =
  selectedItems.length > 0
    ? selectedItems[0].networkAdapters.filter(({ ipAddress }) =>
        selectedItems.every((host) => host.networkAdapters.some((na) => na.ipAddress === ipAddress))
      )
    : [];
```

And then below instead of `!hasCommonNetworkAdapters()` you could use `commonNetworkAdapters.length === 0`, and instead of passing `commonNetworkAdapters(selectedItems)` you can just pass `commonNetworkAdapters`.

It's simpler, and it has the benefit of only running the filter function once per render (currently it'll run 3 times).

mturley:
CI results are messed up here because of the GH Actions outage, but should fix themselves when you push another commit

src/app/Providers/components/VMwareProviderHostsTable/VMwareProviderHostsTable.tsx
gildub: That's pretty slick!

mturley (approved):

PR #280, opened 2020-12-01
Properly remove secret when configuring a host with 'reuse creds' checked

### LINKED ISSUE #279: Configuring a VMware host with "reuse vSphere credentials" checked does not clear an existing secret from the CR
This is a bug in https://github.com/konveyor/virt-ui/pull/244. If we have a host that is already configured with a custom secret (Host CR exists with the `spec.secret` field defined) and we configure it again but leave the box checked to reuse provider credentials, the secret should be removed from the CR. Currently it is not.

mturley:
Resolves #279


gildub (approved):
LGTM

PR #290, opened 2020-12-02
Fix paths to tls.key and tls.crt for node server

mturley:
Another artifact of the project renaming


mturley:
Trivial change that will fix a broken build, others are off the clock, self-merging

Issue #283, opened 2020-12-01
Migration details page: pipeline summary text can wrap to multiple lines, should truncate with ellipses/tooltip.


mturley:
See https://github.com/konveyor/virt-ui/pull/270#issuecomment-736750155

mturley:
@vconzola is this the right solution to this? truncate with ellipses and show the full status summary as a tooltip?

vconzola:
...or not append "Running" to the step name as shown in the mockups.

mturley:
Ah ok sorry. I assumed @gildub did that from a conversation with you. Looks like it came from @fdupont-redhat . We can undo that

gildub:
@vconzola, it effectively came up from a discussion with @fdupont-redhat.
There are two diffents issues at play here:
- The inconsistence between of having a summary "title", like `error` in some cases  as opposed to no "title" when migration is `running`. 
- Besides a bug with the error description displayed, when the step description is added (whether just for error or for running cases too), then wrapping occurs. 

So we need truncate and tooltip in any case.

I'll leave it to you guys to sort out if we prefix or not the summary text with "Running".

gildub:
@mturley,  ellipsis are not kicking in here because the PipelineSummary is not simple text. 
So when using `truncate` it will cut the end of the text without adding ellipsis and the tooltip doesn't trigger either by default.
We can add aTooltip withing the summary but we'd have to customize to get ellipsis.

vconzola:
@gildub I don't have a problem with the inconsistency of not having a title for steps that are running and appending "error" for steps that have failed. Ideally, a title wouldn't be needed it either case and the user would be able to identify the error state by the icon. But for color blind accessibility it's useful to append "error" because to a person unable to distinguish red from green the completed and error pipeline icons will look identical.

mturley:
@vconzola that makes sense to me.

@gildub we can go the custom route, we can probably use the `<TruncatedText>` component I made for that.

PR #291, opened 2020-12-03
Pipeline summary wrap

gildub:
Resolves https://github.com/konveyor/forklift-ui/issues/283


gildub:
@vconzola, please check it out and let me know if that works for you.

src/app/common/components/PipelineSummary.tsx
```diff
@@ -35,11 +35,8 @@ export const getPipelineSummaryTitle = (status: IVMStatus): string => {
     return 'Complete';
   }
   if (status.started && !status.completed) {
-    let title = status.error ? 'Error' : 'Running';
-    if (currentStep?.description) {
-      title = `${title} - ${currentStep.description}`;
-    }
-    return title;
+    if (status.error) return `error - ${currentStep?.description}`;
```
mturley, line 38: You lost the capital E on "error" here

mturley:
@gildub I'm seeing 2 problems with this:
* The tooltip appears all the time, even when all the text is visible
* There are no ellipses, the text just gets cut off by the edge of the container

I think you can probably get rid of the `cellTransforms: [truncate]` and instead use [our custom TruncatedText component](https://github.com/konveyor/forklift-ui/blob/master/src/app/common/components/TruncatedText.tsx) around the text instead of using a Tooltip directly.

vconzola:
@gildub A few minor things... (1) "error" should be capitalized. (2) Are errors always fatal - meaning the migration has failed and stopped?  If so, the icon in the pipeline should be a full red circle, not a partially full one. (3) We should be consistent with the use of periods after the step names. I prefer always having NO periods.

mturley:
Interesting point -- I think you're right @vconzola, but I had been looking at those partially-full red circles and thinking of them as "this step failed while in progress" vs a full circle which could mean "this step ran all the way through and resulted in a failure". I don't think we really have a distinction there though.

vconzola:
@mturley I think distinguishing between this step failed while in progress and this step caused the migration to fail is too nuanced to worry about. If an error causes a migration to stop, I'd like to use the full, red icon.

mturley:
Makes sense.

gildub:
@mturley, I initially tried with TruncatedText component but I wasn't happy with the result, I think I know why so let me give it another go.

@vconzola, no worries to change to display a full red circle icon when an error occurred.

I've restored the capital on "error".

mturley:
Sounds good. Let me know if you need any help, the TruncatedText component is a little weird (CSS `width: 100%` works differently depending on the container it's in).

mturley (approved):
Looks good to me. I can't actually get it to truncate with the mocked text unless I make my window very small, but it looks like it would work correctly at normal sizes with longer messages, so this is perfect.

@vconzola can you give it a review? As far as periods in the names, that's coming straight from the backend, so we should open an issue with Jeff to fix that unless we want to filter out all trailing periods in the UI code.

vconzola:
@gildub @mturley LGTM. I'll open an issue for Jeff. What component would that be?

mturley:
@vconzola virt-controller I think. Or I guess it's forklift-controller now.

PR #289, opened 2020-12-02
Add validation to VMware hostname and username fields

mturley:
Resolves https://github.com/konveyor/forklift-ui/issues/285

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1903431

PR #294, opened 2020-12-03
Fix subdomain regex for vmware hostname field

mturley:
Fixes an issue with a regex introduced in https://github.com/konveyor/forklift-ui/pull/289


fdupont-redhat (approved):

gildub (approved):
LGTM

PR #297, opened 2020-12-04
Switch from yarn to npm to support Cachito

fdupont-redhat:
The goal of the pull request is to allow Cachito to manage the cache during build. And today Cachito only supports npm.
The change is then to remove `yarn.lock` and instead add `package-lock.json`, as `yarm` is able to use `package-lock.json`.

To generate the `packages-lock.json`, the command is `npm install --package-lock-only`.


gildub:
@fdupont-redhat, looks like the build works but the CI fails during lint phase.

I believe npm and yarn build deps differently it impacts the toolchain.
And becausse of the latter we better use npm entirely if that's necessary to help for downstream

package.json
```diff
@@ -7,27 +7,28 @@
...
-    "start:mock": "DATA_SOURCE=mock yarn start",
+    "start:remote": "DATA_SOURCE=remote npm run start",
+    "start:mock": "DATA_SOURCE=mock npm run start",
     "start:dev": "./scripts/start-dev.sh",
-    "start:dev:remote": "yarn setup:dev:remote && DATA_SOURCE=remote yarn start:dev",
-    "start:dev:mock": "DATA_SOURCE=mock yarn start:dev",
+    "start:dev:remote": "npm run setup:dev:remote && DATA_SOURCE=remote npm run start:dev",
+    "start:dev:mock": "DATA_SOURCE=mock npm run start:dev",
     "test": "jest",
     "eslint": "eslint --ext .tsx,.js ./src/",
```
mturley, line 23: I think we need `$(npm bin)/eslint` here as well, although it appears to be working as-is, not sure why that is

package.json
```diff
@@ -7,27 +7,28 @@
...
     "format": "prettier --check --write ./src/**/*.{tsx,ts}",
-    "type-check": "yarn tsc --noEmit",
-    "ci": "yarn type-check && yarn lint && yarn test",
+    "tsc": "$(npm bin)/tsc",
+    "type-check": "npm run tsc --noEmit",
+    "ci": "npm run type-check && npm run lint && npm run test",
     "build:bundle-profile": "webpack --config ./config/webpack.prod.js --profile --json > stats.json",
-    "bundle-profile:analyze": "yarn build:bundle-profile && webpack-bundle-analyzer ./stats.json",
-    "clean": "rimraf dist"
+    "bundle-profile:analyze": "npm run build:bundle-profile && webpack-bundle-analyzer ./stats.json",
```
mturley, line 30: we'll probably need `$(npm bin)/webpack-bundle-analyzer` also

scripts/start-dev.sh
```diff
@@ -5,7 +5,7 @@ export NODE_ENV=development
 export DEBUG=1
 export EXPRESS_PORT=9001
 export NODE_TLS_REJECT_UNAUTHORIZED="0"
-yarn concurrently --names "EXPRESS,WEBPACK" -c "green.bold.inverse,blue.bold.inverse" \
+npm run concurrently --names "EXPRESS,WEBPACK" -c "green.bold.inverse,blue.bold.inverse" \
```
mturley, line 8: I think `concurrently` is another package that needs to be run via `$(npm bin)`

mturley (approved):

PR #299, opened 2020-12-04
Bump eslint and prettier to latest

gildub:
Bump eslint and prettier packages
Adjust a couple of cosmetics issue related


gildub:
Self merging as impact is only on dev and we need that for testing downstream builds

fdupont-redhat (approved):

PR #300, opened 2020-12-04
Bump eslint plugins and deps

gildub:
Bumping up eslint plugins and related typescript definitions following eslint bump.

This creates two eslint warnings related to React Hooks:

The 'createMigrationResult' object makes the dependencies of useEffect Hook (at line 54) change on every render. To fix this, wrap the initialization of 'createMigrationResult' in its own useMemo() Hook.


mturley (approved):
LGTM, we can address the warnings in a followup.

PR #302, opened 2020-12-04
Clear errors in wizard on step navigation instead of component unmount

### LINKED ISSUE #296: Plan wizard: errors on final step are not properly being cleared when moving to a previous step
This is evident in preview mode ("this feature is not available" errors are handled the same as real errors). It was working not long ago, so this is a recent regression.

To reproduce, submit the wizard, see the error appear, go back and click Next again. the error is still there and you can't submit the wizard again.

mturley:
Resolves #296 

I actually think the bug was inadvertently fixed in #293, since I couldn't reproduce it anymore before this change. However this is a cleaner and more reliable way to clear these errors anyway, so I think it's still worth merging.

### DIFF
--- a/src/app/Plans/components/Wizard/PlanWizard.tsx
+++ b/src/app/Plans/components/Wizard/PlanWizard.tsx
@@ -8,6 +8,7 @@ import {
   PageSection,
   Title,
   Wizard,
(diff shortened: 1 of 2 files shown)


gildub (approved):
LGTM

PR #303, opened 2020-12-07
Use 'Save' instead of 'Add' for confirm button text when selecting host network

mturley:
@vconzola I realized that we used "Add" for the button text here, probably because we copied some code from the add provider modal, and it doesn't make sense to me. Does "Save" work for you?


gildub (approved):

vconzola:
@mturley Yep, Save is more appropriate.

PR #307, opened 2020-12-07
Add Status column to VMware hosts table

### LINKED ISSUE #306: Add status column to VMware Hosts table, to indicate if credentials for migration network selection are valid/ready
There is currently no way to see in the UI if the username and password the user entered while selecting a migration network for a host are correct. They can't be validated when submitting the modal because the connection test needs to be asynchronous and reflected afterwards as a condition on the Host CR.

We need to add a Status column to that table to show these resulting conditions. It would show Ready most of the time, and show some kind of error if the credentials are invalid.

cc @vconzola

mturley:
Resolves #306.

Provides a place for conditions on a matching Host CR to be surfaced (invalid credentials, connection issues, etc). A special case to note here is that not all hosts will have Host CRs (if they are left with the default network, they won't) so there may be no status available for those hosts. This code assumes that if there is no Host CR then the status is "Ready" for that host.


vconzola (commented):
@mturley What you implemented for this looks fine, but I wonder about the assumption you make in your comment above. Can we really assume that the default network is always "ready"? If it's not ready would we even be able to see anything regarding VMware providers? I think we need to understand the implications better before we just say it's ready.

mturley:
That's fair. I didn't like that it would otherwise just say "unknown" with a warning icon for all of them (always, unless you select a network). Not sure what the better alternative would be, since we can't get any status info on them until we create CRs for them (I think?)

mturley:
Updated to use "Not configured" text in the status column for hosts with no config CR, as described by @fdupont-redhat on scrum this morning. Should be ready to go @vconzola @gildub

vconzola:
@mturley I was trying to follow the conversation in scrum this morning, but couldn't completely. Was does "Not configured" mean? As a user it tells me I need to take an action to configure something, but I don't see anything to configure; hence confusion.

mturley:
Hm, fair point @vconzola . It just means they haven't selected a network, which under the hood means there is no Host CR (no user configuration data for that host, so it'll leave everything at default). Right now the only thing they can configure is the network.

mturley:
Maybe we can just leave that cell entirely blank in that case? It already says (default) for the network, and has bandwidth/MTU blank.

mturley:
Rebased, and removed the "not configured" text per discussion with @vconzola . If there is no host CR the status column will just be blank.

gildub (approved):
LGTM

PR #311, opened 2020-12-07
Uniquely identify host networks by name instead of IP address, remove IP from dropdown

mturley:
From a discussion with @fdupont-redhat . Network adapters connected to the same network on different hosts will have different IP addresses, so we can't use the host network adapter IP as a unique identifier for the network. Selecting more than one host would never result in any common networks to show in the modal.

This PR changes the logic so common networks are identified by networkAdapters with the same name instead of IP, and when creating the Host CRs the IP from each host's matching adapter is used. Also, the IP address has been removed from the dropdown menu of networks, only the name, bandwidth and MTU are shown.


gildub (approved):
LGTM

PR #313, opened 2020-12-08
Helpers and constant cosmetic changes

gildub:
- Renames PlanStatusType from PlanStatusAPIType.
- If/then/else replaced with if/then only when systematically returning value


mturley:
Since there is still confusion around #307 I'm going to go ahead and merge this, looks good.

mturley (approved):

PR #315, opened 2020-12-08
Maintain one linked secret for each host CR, use ownerReferences for secrets, remove "reuse credentials" option

### LINKED ISSUE #314: Configuring host migration networks: should create a unique secret CR per host CR, and not allow reusing vSphere credentials
After a discussion with @fdupont-redhat on this morning's scrum, we need to make some changes to how Host CRs are created in the Select Migration Network modal on the VMware Hosts page.

* Currently, when configuring multiple hosts at the same time, one Secret is created with the credentials you pass and it is referenced in all the Host CRs. This could make it problematic to change the password on one of those hosts later on without affecting the others. Instead, a distinct Secret should be created for each host, named clearly to indicate which host it is for, and with `ownerReferences` set up so it will delete itself if the host CR is deleted.

* Currently there is a checkbox for "Reuse vSphere credentials" which will omit the secret from the Host CR, we've been asked to remove this.

mturley:
Resolves #314

### DIFF
--- a/src/app/Providers/components/VMwareProviderHostsTable/SelectNetworkModal.css
+++ b/src/app/Providers/components/VMwareProviderHostsTable/SelectNetworkModal.css
@@ -1,3 +1,0 @@
-.selectNetworkModal .extraSelectMargin {
-  margin-bottom: 184px;
-}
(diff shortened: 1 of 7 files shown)


mturley:
Waiting to hear back from @jortel about whether it's ok to patch the secrets immediately after creating them before we merge this one.

mturley:
Ok @gildub I confirmed this is how it should work, and tested it successfully against Jeff's cluster. Good to go

gildub (approved):
LGTM

PR #317, opened 2020-12-09
Replace custom checkbox columns with Table's onSelect prop

### LINKED ISSUE #47: Switch expandable selectable tables from custom checkbox columns back to using the `onSelect` prop
The issue that kept me from using the PF Table's `onSelect` prop was that the expanded content had a phantom column that we couldn't control, to the left of our expanded content. I'm not sure if I missed this before, but we might just be able to use the `fullWidth: true` property in the expanded rows, as in [this example](https://www.patternfly.org/v4/documentation/react/components/table#collapsible) to get around the issue. This would also fix the awful layout these tables have on mobile viewports.

mturley:
Resolves #47


gildub (approved):
LGTM

Issue #320, opened 2020-12-10
Unit/Integration test framework.


gildub:
For our unit test framework we're going to use React Testing Library which helps to create solid, light unit tests (and eventually integration tests) which we have confidence with.

As we have discussed, Cypress could be used for end to end tests.

mturley:
It'll be interesting to see where we draw the line between our react-testing-library tests and Cypress tests. Even if some of them are redundant it would be nice to have the React-level tests to prevent regressions immediately since they run in CI in this repo, and the Cypress browser-level tests to make sure things really work in a real app image.

PR #321, opened 2020-12-10
Initial configuration for react testing library

gildub:
- Adds @testing-library/user-event package
- Moves jest.config.js to `./config
- Sets off coverage flag by default in jest.config.js so it's used only when needed by activating the option instead of having to turning off.

Resolves https://github.com/konveyor/forklift-ui/issues/320


mturley:
Huh, the test that is complaining about JSX being invalid ran fine before these config changes. I wonder what that's all about.

gildub:
@mturley, yes sorry, it was missing a couple of things.

package.json
```diff
@@ -18,10 +18,12 @@
     "start:dev": "./scripts/start-dev.sh",
     "start:dev:remote": "npm run setup:dev:remote && DATA_SOURCE=remote npm run start:dev",
     "start:dev:mock": "DATA_SOURCE=mock npm run start:dev",
-    "test": "$(npm bin)/jest --coverage",
     "test:watch": "npm run test -- --watch",
```
mturley, line 21: The `--watch` param doesn't make it through to jest (this runs `npm run test:unit --watch`, which doesn't have the `--` separator). Basically when we use `-- --foo` the script we're calling can't be another `npm run`. We should remove the existing `test:watch` script and add these:

```json
"test:unit:watch": "npm run test:unit -- --watch",
"test:watch": "npm run test:unit:watch",
```

package.json
```diff
@@ -18,10 +18,12 @@
...
     "lint": "$(npm bin)/eslint --ext .tsx,.js ./src/",
     "lint:fix": "npm run lint -- --fix",
+    "test:unit": "$(npm bin)/jest --coverage --rootDir=. --config=config/jest.config.js",
+    "test": "npm run test:unit",
+    "eslint": "$(npm bin)/eslint --ext .tsx,.js ./src/",
```
mturley, line 26: Do we need both the `lint` script and the `eslint` one? we should just pick one and drop the other, imo. I like the consistency of naming the scripts after what they're for, not the tool (so `type-check`, `lint` and `test`, not `tsc`, `eslint` and `jest`).

gildub: No we don't need both.  I agree with naming.

package.json
gildub: Thanks for highlighting this.
I think we just need this:
```
"test": "npm run test:unit -- --coverage",
"test:unit": "$(npm bin)/jest --rootDir=. --config=config/jest.config.js",
"test:watch": "npm run test:unit -- --watch",
```

PR #322, opened 2020-12-10
Add popover to show condition messages on the status column in the providers and hosts tables

### LINKED ISSUE #178: Add popover with reason for provider and host status
Right now, the status column in the providers tables can just say Warning or Critical with no explanation of why. We should add a popover when you click on the icon to indicate the reason (I assume.. can you confirm this is the right interaction @vconzola ?)

Screen Shot 2020-11-02 at 11 52 19 AM

mturley:
Resolves #178 

The cell in the table will continue to show the most severe condition's severity, but now clicking on it will show a popover with all conditions present on the object, icons for their severity, and their full messages.


gildub (approved):
LGTM

PR #323, opened 2020-12-10
Unit tests for VMwareProvidersTable

gildub:


src/app/Providers/components/ProvidersTable/VMware/__tests__/VMwareProvidersTable.tsx
```diff
@@ -0,0 +1,88 @@
+import React from 'react';
```
mturley, line 1: IMO we should name this file with the extension `.test.tsx`, it's convenient to have the test file and the component file named differently so you can tell them apart in the tabs of your editor.

mturley (changes requested):
I think if you rename the test file to `VMwareProvidersTable.test.tsx` this is good to merge, since we can address other cases in a future PR. It's a good foundation.

src/app/Providers/components/ProvidersTable/VMware/__tests__/VMwareProvidersTable.tsx
gildub: I agree as I experience confusion while having the component file opened at same time.

PR #324, opened 2020-12-14
Unit tests for OpenShiftProvidersTable

gildub:
Skipping 'renders storage classes' test for now, there is an issue with that button, might have to use a different approach.


src/app/Providers/components/ProvidersTable/OpenShift/__tests__/OpenShiftProvidersTable.tsx
```diff
@@ -0,0 +1,100 @@
...
+    render(
+      <NetworkContextProvider>
+        <Router history={history}>
+          <OpenShiftProvidersTable {...props} />
+        </Router>
+      </NetworkContextProvider>
+    );
+
+    const sharedClasses = screen.getAllByRole('button', { name: '0' });
+    userEvent.click(sharedClasses[2]);
```
mturley, line 52: It would be nice to have the selectors be a little more semantic instead of hard-coding array indexes like this.. we can leave this for the PR that fixes this test, but it would be better to instead e.g. find the row that contains that provider name, traverse from there to find the button, etc. Traversing the document like a user would with their eyes will make the tests less brittle.

mturley (changes requested):
Same comment as #323, let's name the test files with `*.test.tsx`. Otherwise this is good to go, but let's open an issue for the skipped test.

src/app/Providers/components/ProvidersTable/OpenShift/__tests__/OpenShiftProvidersTable.tsx
gildub: That would be much better meanwhile once a row is queried, react testing libraries can't be used to query on HTMLElement.

mturley: Interesting. Maybe there's a different better way then, we can figure that out later.

PR #331, opened 2020-12-15
Add hook for fetching providers using cluster API

gildub:
https://github.com/konveyor/forklift-ui/issues/327

PR #334, opened 2020-12-16
Load and use Provider CRs as source of truth instead of only loading providers from inventory

### LINKED ISSUE #327: [BZ#1907926] Provider CR disappears when incorrect hostname or IP address is provided while updating VMware provider in UI
RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1907926

Currently the providers table is rendered only from provider objects in the inventory API, and we don't load provider CRs from the cluster API (we only create those CRs). We should add a query for the provider CRs and flip this so the table is driven by those CRs and supplemented by the inventory API objects. This way we always have status for each provider even if it's not available in the inventory API.

mturley:
Resolves #327 (https://bugzilla.redhat.com/show_bug.cgi?id=1907926)

In cases where the inventory failed to include a provider (invalid credentials, etc) the provider was just disappearing from the table. Instead of relying on a provider existing in the inventory API to appear in the table, this PR uses the new query added in #331 to load provider CRs from the cluster API and show a table row for each of those, with inventory data correlated where possible. If a provider CR does not have matching inventory data, those cells in the table will be blank, but the Status cell will still work.

To preserve the simplicity of passing around one array of providers in the table components, I created a new generic interface `ICorrelatedProvider<T>` which extends the provider CRD (`IProviderObject`) and adds an extra property `inventory` of type `T`. So an `ICorrelatedProvider<IVMwareProvider>` is a provider CR with VMware provider inventory data attached to it. See the new helper function `correlateProviders` for the relevant logic.


mturley:
@gildub this involved refactoring a lot of code where inventory providers were being passed around to instead handle cluster providers... we should review it thoroughly to make sure everything still works, but TypeScript was very magically helpful in telling me all the places I needed to fix things so I'm fairly confident I got everything.

gildub:
@mturley, noted, I'll do a thorough review.

mturley:
Had to fix some tests, this should be ready now @gildub

gildub (approved):
Very nice work with the generic interface!
Thanks for addressing the issue while loading vmware provider hosts.
This works like a charm now.

Issue #333, opened 2020-12-15
[BZ#1908020] When click on providers, It should navigation to the respective provider page


mturley:
RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1908020

> When a user click host number of a provider, it navigate to host detail page, however getting back to the same provider is not a single click task.
> 
> For Example, if the user on the host page of a vmware provider name vsphere65, you will see a navigation option on top of page as "Providers<hyperlink>-->vsphere65--->Hosts"
> 
> 1. Click on hyperlink "providers" 
> 2. It will by default navigate you to "openshift virtualization" page
> 3. Switch tab to your provider, in my case it is "vmware"
> 4. Find the target provider row
> 
> For a better user experience, It should be a single click task, at least clicking on "providers" hyperlink should navigate you to the respective provider page 

The simplest way I can think to solve this is to make the selected tab on the providers page driven by a URL route instead of local state, so that navigating back persists the tab you were on previously.

PR #343, opened 2020-12-17
Adds navigation for provider types

gildub:
Fixes https://github.com/konveyor/forklift-ui/issues/333

### DIFF
--- a/src/app/Providers/ProvidersPage.tsx
+++ b/src/app/Providers/ProvidersPage.tsx
@@ -27,14 +27,25 @@ import { IPlan, IProviderObject } from '@app/queries/types';
 import { ResolvedQueries } from '@app/common/components/ResolvedQuery';
 import { getAggregateQueryStatus } from '@app/queries/helpers';
 import { QueryStatus } from 'react-query';
+import { useRouteMatch } from 'react-router';
 
 export const EditProviderContext = React.createContext({
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   openEditProviderModal: (_provider: IProviderObject): void => undefined,
   plans: [] as IPlan[],
 });
+export interface IProvidersMatchParams {
+  url: string;
+  providerType: string;
+}
 
 const ProvidersPage: React.FunctionComponent = () => {
+  const match = useRouteMatch<IProvidersMatchParams>({
+    path: '/providers/:providerType',
+    strict: true,
+    sensitive: true,
+  });
+
   const clusterProvidersQuery = useClusterProvidersQuery();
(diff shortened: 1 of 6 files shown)


vconzola (commented):
LGTM.

src/app/Providers/ProvidersPage.tsx
```diff
@@ -57,8 +68,11 @@ const ProvidersPage: React.FunctionComponent = () => {
   ).filter((type) => !!type) as ProviderType[];
 
   const [activeProviderType, setActiveProviderType] = React.useState<ProviderType | null>(
```
mturley, line 70: Instead of `useState` here, we should probably just use `const activeProviderType = `match?.params.providerType || null` and then everywhere we have `setActiveProviderType` we should instead redirect to the route for that provider type. That way the route is the single source of truth. Right now if you change tabs, go somewhere, and click Back it won't remember the tab change.

gildub: That's a good point.

src/app/app.test.tsx
```diff
@@ -1,13 +1,11 @@
...
 import '@testing-library/jest-dom/extend-expect';
 import { App } from '.';
-import { APP_TITLE } from './common/constants';
 
 describe('App', () => {
   test('renders welcome page without errors', async () => {
     render(<App />);
-    await waitFor(() => screen.getByRole('heading'));
-    expect(screen.getByRole('heading')).toHaveTextContent(APP_TITLE);
+    expect(screen.getByRole('heading', { name: /Forklift/ })).toBeInTheDocument;
```
mturley, line 9: it's weird that APP_TITLE wouldn't work here, these tests would fail if you have `BRAND_TYPE=RedHat` in your .env file. I guess that's fine though, if it's a problem we can fix it later.

src/app/Providers/ProvidersPage.tsx
```diff
@@ -56,14 +69,9 @@ const ProvidersPage: React.FunctionComponent = () => {
     new Set(clusterProviders.map((provider) => provider.spec.type))
   ).filter((type) => !!type) as ProviderType[];
 
-  const [activeProviderType, setActiveProviderType] = React.useState<ProviderType | null>(
-    availableProviderTypes[0]
-  );
-  React.useEffect(() => {
```
mturley, line 62: We still need the useEffect here because when the component first mounts at `/providers` the `match` won't be defined, so we end up with no tab selected.

mturley:
@gildub my remaining comments were pretty trivial to address and I know you're busy, so I just pushed commits here to address them.

mturley (approved):

PR #344, opened 2020-12-17
[Backport] Adds navigation for provider types (#343)

mturley:
* Add navigation for provider types

* Fix tests

* Use better selector for welcome heading

* useHistory() instead of useState() for current provider tab

* Get welcome heading with role's name

* Restore useEffect to select initial tab when on base /providers route

* Use APP_TITLE in test instead of hard-coding Forklift

Co-authored-by: Mike Turley <mike.turley@alum.cs.umass.edu>


gildub (approved):
LGTM

PR #349, opened 2020-12-18
Add status condition popover for plans that are not started

### LINKED ISSUE #347: Add StatusCondition popover to plans that are showing status icons (not started)
If a plan is not ready for one reason or another we should show that by listing conditions in a popover just like we do for providers and hosts. We can reuse the StatusCondition component.

mturley:
Resolves #347


gildub (approved):
LGTM

PR #355, opened 2020-12-18
Load and prefill secrets in the edit provider and select host network forms

### LINKED ISSUE #336: [BZ#1907734] Read and prefill secrets in relevant forms
RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1907734

I was apparently mistaken in my assumption that secrets in the API are write-only for our purposes. Any secret created by the logged in user can be base64-decoded and prefilled in the forms when editing secret-related CRs.

mturley:
Resolves #336 (https://bugzilla.redhat.com/show_bug.cgi?id=1907734). Secrets can be loaded and decoded when opening them for editing, so we can get rid of our logic to only require those fields if the user desires to change them, etc.

### DIFF
--- a/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
+++ b/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
@@ -1,16 +1,6 @@
 import * as React from 'react';
 import * as yup from 'yup';
-import {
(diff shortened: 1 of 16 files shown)


vconzola (commented):
LGTM. Thanks @mturley

PR #358, opened 2021-01-05
Prevent selecting non-ready providers, prevent editing mappings/plans with non-ready associated providers

### LINKED ISSUE #335: Non-ready providers should be excluded from any provider selections in forms
If a provider does not have the Ready condition, inventory API requests for resources in that provider will fail. We should either filter out or disable those options in Select fields for providers (and if filtering out, show a message when no providers remain instead of showing an empty dropdown). This might also be a good opportunity to factor out a `ProviderSelect` common component, since we have some duplicated code for getting the available provider options and setting up that Select.

mturley:
Resolves #335 

* Factors out the four redundant source/target provider select fields we had into one reusable component
* Changes the provider selection to use Provider CRs from the cluster instead of inventory providers, so that a non-ready provider that is missing from the inventory API will still appear
* Disables the providers in the dropdown that are not ready, with a tooltip explaining why
* Disables editing mappings and plans whose associated providers are not ready, with a tooltip explaining why

To see the tooltips in mock mode, here's where to look:
* The source and target provider dropdowns in the plan wizard and the create mapping modal (the second provider in each list is not ready)
* The edit option on the kebab menu on the 6th migration plan, the 2nd network mapping, and the 2nd storage mapping.

Screen Shot 2021-01-05 at 1 06 48 PM

Screen Shot 2021-01-05 at 2 12 39 PM

Screen Shot 2021-01-05 at 2 12 49 PM


gildub (approved):
LGTM

PR #364, opened 2021-01-08
MappingBuilder: Disable incompatible target storage classes, prefill default targets

### LINKED ISSUE #248: Mappings: Filter target storage classes by what is compatible with kubevirt
From a discussion with @fdupont-redhat, it turns out that only certain storage classes will be compatible with kubevirt, so once we can determine which ones are valid we should filter the available mapping choices.

Marking as `blocked` until we have the API resource to determine this filter logic.

### LINKED ISSUE #363: Pre-select default mapping targets in mapping builder
We now have an annotation to check on a storage class to determine if it's the default. Per @fdupont-redhat :

> the annotation for the default storage class is `storageclass.kubernetes.io/is-default-class: true`. If the annotation is absent, it means false.

We can see the CR object and its annotations by adding `?detail=true` on the storageclasses query URL.

When the mapping builder is initialized in the wizard, and when items are added in the mapping modal, the target should prefill the storageclass with this annotation if present. To be consistent, we should also prefill the target networks as "Pod network".

mturley:
Resolves #248
Resolves #363

To see the incompatible storage class message in preview mode, while creating a mapping, look at the second storage class in the list (regardless of the provider you select).


gildub (approved):
LGTM

PR #365, opened 2021-01-11
Forms to have disabled save button unless really changed

### LINKED ISSUE #329: [BZ#1907733] While editing resources, save button is enabled even when no changes are made
RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1907733

The BZ refers to providers but this applies to all of our editing forms. See my comment in the BZ for implementation notes.

gildub:
Resolves #329

- Bumps @konveyor/lib-ui
- Removes "setTouched" from form pre-filling for providers
- Use `useFormState` `isTouched` for the providers: This actually works until the use touch a field (position the pointer and leaves (unblur) that field because it marks the field as touched.

So to make it work, I believe we need to have preffiling happen during the useFormState object creation by passing the initialValues returned from backend.

Edited by @mturley:

Solved by adding a `setInitialValue` function to each field in useFormState in lib-ui. This allows us to prefill values without marking the form `isDirty`, and use the `isDirty` value to manage the disabled state of the submit button. Implemented it for all other prefilled forms (plans, mappings, hosts)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1036,9 +1036,9 @@
       }
     },
     "@konveyor/lib-ui": {
-      "version": "1.10.5",
-      "resolved": "https://registry.npmjs.org/@konveyor/lib-ui/-/lib-ui-1.10.5.tgz",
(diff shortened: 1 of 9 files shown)


mturley:
Removing the WIP tag here and approving. @gildub if you're ok with the patches I made, feel free to merge.

mturley (approved):

mturley:
Also rebased.

gildub (commented):
LGTM

PR #373, opened 2021-01-13
Add upstream and downstream favicons, and make webpack output more readable

### LINKED ISSUE #357: [BZ#1914741] Add a favicon
Currently there is no favicon (icon that appears in the browser tab). It should be a Konveyor icon when built upstream and a Red Hat icon otherwise. Very low priority.

mturley:
Resolves #357 (https://bugzilla.redhat.com/show_bug.cgi?id=1914741)

Adds favicon files for both Konveyor and Red Hat brands, and configures webpack to use the correct one depending on the `BRAND_TYPE` environment variable (falling back on Konveyor if that variable is undefined).

As part of this PR, I was frustrated trying to figure out why my env variable wasn't being passed correctly, because our webpack-dev-server progress output is so noisy that it's impossible to read the top of the output. I decided to disable the progress output of webpack to prevent this issue (no more percent progress, but we still get "waiting", "compiled", and "compiled successfully" messages). This keeps output light and readable, so the environment variables can easily be seen when running webpack.

If this issue is ever resolved: https://github.com/kimmobrunfeldt/concurrently/issues/85 we can turn progress back on because it supposed to be outputting to one line that gets updated instead of hundreds of lines.


config/webpack.common.js
```diff
@@ -147,8 +152,14 @@ module.exports = (env) => {
...
               filename: 'index.html.ejs',
               template: `!!raw-loader!${path.resolve(__dirname, '../src/index.html.ejs')}`,
-            }
-      ),
+            }),
+        favicon: path.resolve(
+          __dirname,
+          process.env.BRAND_TYPE === 'RedHat'
+            ? '../src/favicon-redhat.ico'
+            : '../src/favicon-konveyor.ico'
```
apinnick, line 160: @mturley 
I noticed in the sprint 195 demo that you are using the Konveyor icon rather than the Forklift icon. Will the Konveyor icon be used for all the UIs?

mturley: @apinnick I was only replacing Red Hat branding with Konveyor branding in the upstream, which I figured was equivalent because in the upstream builds we're using Red Hat icons (and not Migration Toolkit icons). It's a good question though, maybe we should be using product-specific favicons.

Either way though the product logo could probably go on the welcome page instead of the generic migration icon.
@vconzola do you have thoughts on these, or know who we should ask?

PR #380, opened 2021-01-13
Add qe-tests folder with qe-committers as CODEOWNERS, add to eslint, prettier and tsc configs

mturley:
This PR establishes a dedicated subdirectory `qe-tests` at the root of the repo for use by the @konveyor/migration-engineering-qe-committers team. CODEOWNERS is set such that the `ui-committers` team should not be required to review PRs which modify only `qe-tests`.

I put a separate `tsconfig.json` in this directory so it can stand alone with different TS options if necessary, and I excluded it from the root `tsconfig.json`. I also added a separate `npm run tsc-qe` script that will type-check only the new directory, and moved the current one to `npm run tsc-src` (now `npm run tsc` runs both, so CI will check both).

I added both `tsconfig.json` files as projects in the `.eslintrc`, so running ESLint will lint both projects. I also added a glob to the Prettier CLI command in `npm run format` so that it will format files in both projects.

@ibragins, note: the option `"types": []` in the `qe-tests/tsconfig.json` file is important, it prevents tsc from walking to the parent directory and looking for typings in the source code. See https://github.com/microsoft/TypeScript/issues/11363#issuecomment-251516924


gildub (approved):
LGTM

PR #381, opened 2021-01-14
Warn instead of blocking when selecting a storage class with no provisioner

mturley:
https://bugzilla.redhat.com/show_bug.cgi?id=1916178

Screen Shot 2021-01-14 at 10 35 23 AM


apinnick:
@mturley @fdupont-redhat  

After checking OpenShift docs, I would like to change the error message slightly.

Could we change the first sentence to "This storage does not support dynamic provisioning"?

mturley:
@apinnick sure. fixed in https://github.com/konveyor/forklift-ui/pull/386 (backport: https://github.com/konveyor/forklift-ui/pull/387)

PR #391, opened 2021-01-18
Pipeline summary: handle special case where the VM has an error but the current step does not

### LINKED ISSUE #390: [BZ#1916963] 'Elapsed time' timer on migration plan doesn't stop after a migration is cancelled from OCPv
https://bugzilla.redhat.com/show_bug.cgi?id=1916963

mturley:
Resolves #390. (https://bugzilla.redhat.com/show_bug.cgi?id=1916963)

There is a special corner case we were not covering, which can be caused by a user suddenly deleting the VM Import CR after the migration starts (among other potential causes). In this case, the migration fails (and so the VM-level status has an error) but the pipeline step is still in a state of being started, not being completed, and not having an error. Because the VM was marked complete and no pipeline steps had errors, they were all showing green icons. And because the step during which the error happened was never marked complete, its elapsed time was continuing to tick forever.

This PR handles this case by making the current step icon red if there is a VM-level error even if there is not a pipeline step-level error. In this case, the VM-level error is displayed in the pipeline status column and the elapsed time is set to the difference between the step start time and the VM completed time.

Screen Shot 2021-01-18 at 5 24 27 PM

### DIFF
--- a/src/app/Plans/components/Step.tsx
+++ b/src/app/Plans/components/Step.tsx
@@ -6,7 +6,6 @@ import {
 } from '@patternfly/react-icons';
 import {
   global_danger_color_100 as dangerColor,
-  global_warning_color_100 as warningColor,
   global_disabled_color_200 as disabledColor,
   global_info_color_100 as infoColor,
   global_success_color_100 as successColor,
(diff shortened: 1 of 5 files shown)


gildub (approved):
LGTM

PR #393, opened 2021-01-19
Upgrade webpack to version 5

gildub:
The driver is that webpack is now 2 years old and there are new features in webpack for better typescript support and also to allow nodejs to benefit from es6 modules.

Upgrades makes few changes
- Environment variables are now declared explicitely in webpack configuration: 
   `new webpack.EnvironmentPlugin(['DATA_SOURCE', 'BRAND_TYPE', 'NODE_ENV'])`
   See https://webpack.js.org/plugins/environment-plugin/ for details
   On the long run we have to redefine those as migration documentation explicitely says "webpack 5 does no longer include a polyfill for this Node.js variable. Avoid using it in the frontend code."
- webpack-dev-server is now triggered using `webpack serve ...`
- mini-css-extract-plugin now needs distincts loader entries to avoid "Error: Automatic publicPath is not supported in this browser"
- To avoid asset duplication with old assets loader (url-loader, file-loader and raw-loader) uses following option: `type: 'javascript/auto'`

For migration details: https://webpack.js.org/migrate/5


gildub:
PR Tests/CI is failing:
ERROR in ./node_modules/@patternfly/react-core/dist/styles/base.css
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
Error: Automatic publicPath is not supported in this browser

gildub:
This PR has been replaced with PR#402 du to issue after moving master branch to main in my repo.

PR #394, opened 2021-01-19
Add Status column to Mappings tables, handle invalid mappings (missing sources/targets)

### LINKED ISSUE #208: How to handle mappings that reference any storages/networks that have been removed?
We should make sure none of our UI breaks if the resources referenced in a mapping (whether part of a plan or not) no longer exist. They are inventory resources, so we can't automatically delete mappings when they are removed via ownerReferences, so we'll probably need something to indicate when rendering/editing them that they are broken.

mturley:
Resolves #208.

Adds a Status column to the mappings table which will show OK or Invalid (if the mapping references any missing inventory resources). If an invalid mapping is expanded, the missing items are pointed out with red "Not available" text:

Screen Shot 2021-01-21 at 2 17 45 PM

Adds a popover on the Invalid status value to explain why it's invalid:

Screen Shot 2021-01-21 at 1 15 13 PM

Prevents an invalid mapping from being edited with a tooltip explaining why:

Screen Shot 2021-01-21 at 1 57 08 PM

Prevents an invalid mapping from being selected in the plan wizard with a tooltip explaining why:

Screen Shot 2021-01-21 at 1 39 23 PM

Adds mock data to reproduce these cases in preview mode.

### DIFF
--- a/src/app/Mappings/components/MappingDetailView/MappingDetailView.css
+++ b/src/app/Mappings/components/MappingDetailView/MappingDetailView.css
@@ -5,3 +5,7 @@
 .mapping-view-arrow-cell {
   padding-top: 12px;
 }
+
+.missing-item {
+  color: var(--pf-global--danger-color--100);
+}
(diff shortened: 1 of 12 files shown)


vconzola (commented):
(1) I’m not sure we’re using link text for popovers in other places, are we? I thought we were “popping over” on the icon? I know this isn’t what PF says to do, but i think it’s what we’ve done elsewhere. just want to be consistent. (2) I’m not crazy about “[missing]“. how about “Not available” without the brackets? Brackets make it look too code-y.

mturley:
This one should be ready to go whenever you get a chance to review @gildub

gildub (approved):
LGTM

PR #400, opened 2021-01-26
Add support for canceling migrations at the VM level

### LINKED ISSUE #192: [Bz#1899371] Add support for canceling a migration
In IMS we had cancellation at a per-VM level. Unclear how this will work exactly here.

Depends on controller support not yet implemented.

https://bugzilla.redhat.com/show_bug.cgi?id=1899371

mturley:
Resolves #192 (https://bugzilla.redhat.com/show_bug.cgi?id=1899371)

Screen Shot 2021-03-03 at 4 18 24 PM

Screen Shot 2021-01-27 at 4 12 34 PM

In preview mode, you can see the canceled state of a VM by looking at the plan "plantest-3".

Screen Shot 2021-03-03 at 3 33 21 PM

### DIFF
--- a/src/app/Mappings/components/MappingsActionsDropdown.tsx
+++ b/src/app/Mappings/components/MappingsActionsDropdown.tsx
@@ -6,7 +6,7 @@ import {
   useDeleteMappingMutation,
   useResourceQueriesForMapping,
 } from '@app/queries';
-import ConfirmDeleteModal from '@app/common/components/ConfirmDeleteModal';
+import ConfirmModal from '@app/common/components/ConfirmModal';
 import { areAssociatedProvidersReady } from '@app/queries/helpers';
 import ConditionalTooltip from '@app/common/components/ConditionalTooltip';
 import { isMappingValid } from './helpers';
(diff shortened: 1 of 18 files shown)


mturley:
Note to self: need to also be able to select VMs that are not started (double check with Jeff.. what if the plan hasn't been started yet at all?)

mturley:
This should be ready to go now, but I need to test it in a real environment.

mturley:
Unblocked, just needs testing.

mturley:
This is almost ready. Last step is making sure the pipeline steps look correct after cancellation. @mansam is going to add a completed timestamp to the current pipeline step at cancel time (so the elapsed time will stop ticking) but the pipeline summary icons need a new condition so the cancelled step doesn't look complete/green. Probably just use the in-progress icon but black or grey, or use a different icon (@vconzola let's follow up when you're back from PTO to see if the icon for a cancelled step in the pipeline should look different).

mansam:
https://github.com/konveyor/forklift-controller/pull/175 has been merged, which marks the current pipeline step complete when a VM is canceled.

vconzola (changes requested):
@mturley Please drop "selected" from the button label. It's not standard and implied by the fact that the button is disabled until one or more VMs is selected.

mturley:
@vconzola thanks, done.

PR #402, opened 2021-01-27
Upgrade webpack to version 5

gildub:
The driver is that webpack is now 2 years old and there are new features in webpack for better typescript support and also to allow nodejs to benefit from es6 modules.

Upgrades makes few changes

- Environment variables are now declared explicitely in webpack configuration:
    `new webpack.EnvironmentPlugin(['DATA_SOURCE', 'BRAND_TYPE', 'NODE_ENV'])`
    See https://webpack.js.org/plugins/environment-plugin/ for details
    On the long run we have to redefine those as migration documentation explicitely says "webpack 5 does no longer include a polyfill for this Node.js variable. Avoid using it in the frontend code."
- webpack-dev-server is now triggered using `webpack serve ...`
- mini-css-extract-plugin now needs distincts loader entries to avoid "Error: Automatic publicPath is not supported in this browser"
- To avoid asset duplication with old assets loader (url-loader, file-loader and raw-loader) uses following option: `type: 'javascript/auto'`

For migration details: https://webpack.js.org/migrate/5


package.json
```diff
@@ -8,11 +8,11 @@
   "private": true,
   "scripts": {
     "dr:surge": "node dr-surge.js",
-    "build": "npm run clean && NODE_ENV=production $(npm bin)/webpack --config ./config/webpack.prod.js",
+    "build": "npm run clean && NODE_ENV=production BRAND_TYPE=Konveyor $(npm bin)/webpack --config ./config/webpack.prod.js",
```
mturley, line 11: I think this would override an exported BRAND_TYPE from outside this line, which I think the container builds use (`BRAND_TYPE=RedHat npm run build`). If we need a default we should probably just fall back on Konveyor when we read the variable if it's undefined.

mturley (changes requested):
Following up on my earlier comment, we should remove the `BRAND_TYPE=Konveyor` from the build script in package.json, and instead provide defaults when configuring the plugin:

```diff
diff --git a/config/webpack.common.js b/config/webpack.common.js
index ef9a7b4..4c70119 100644
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -168,7 +168,11 @@ module.exports = (env) => {
         systemvars: true,
         silent: true,
       }),
-      new webpack.EnvironmentPlugin(['DATA_SOURCE', 'BRAND_TYPE', 'NODE_ENV']),
+      new webpack.EnvironmentPlugin({
+        DATA_SOURCE: 'remote',
+        BRAND_TYPE: 'Konveyor',
+        NODE_ENV: 'production',
+      }),
     ],
     resolve: {
       extensions: ['.js', '.ts', '.tsx', '.jsx'],
```

This will prevent issues overriding values passed in from a parent shell, and it also prevents errors if any of the variables are undefined like they were for me when starting in mock mode.

gildub:
@mturley, I've updated it following your comments, thanks.

mturley (approved):
LGTM

Issue #281, opened 2020-12-01
Restore Migration Analytics features when new backend implementation is settled


mturley:
This issue is for the eventual reversion of PR https://github.com/konveyor/virt-ui/pull/282 (issue #274).

mturley:
One thing to note when we reimplement this: Polling was disabled in the plan wizard due to issues it may cause in the mapping steps. I think that may have been resolved now by our changes for editing plans, which made those steps more resilient. We will probably want to turn polling back on in the wizard, at least on the Select VMs step, in order to make sure the analytics data is always up to date (especially since we added a spinner when the concerns are out of date in https://github.com/konveyor/forklift-ui/pull/242).

mturley:
https://github.com/konveyor/forklift-ui/pull/411 will restore some of this functionality (restores the migration analysis in the Select VMs step of the plan wizard, but does not restore the payload download feature to the providers page).

yaacov:
closing as new development is moving to:
https://github.com/kubev2v/forklift-console-plugin

PR #411, opened 2021-02-04
Restore the Migration Analysis column to the Select VMs step of the plan wizard (and adapt to new data structure)

mturley:
Addresses part of #281.

Example:

Screen Shot 2021-02-04 at 3 22 44 PM


gildub (approved):
LGTM

PR #427, opened 2021-02-16
Bug 1929050: Allow deleting a Host CR by selecting the Management Network option

### LINKED ISSUE #292: Determine and show the default migration network for hosts before they are configured
Related RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1908034

From the discussion here: https://github.com/konveyor/virt-ui/pull/244#issuecomment-731192433

Right now there is no way for the UI to know what network will be used for a host unless it was specified by the user (the default is determined in the backend at migration time), so we just display (default) and no other information. We also can't specify in the list of networks to choose from which is the default.

If there is some way of identifying from vSphere which one is the default network, we should use that to display details in the table before a host is configured, and distinguish the default network in the network selection modal.

Relevant chatter from Slack:

@jortel :
> I wonder if it's simple as always the one named: "Management Network"?

@fdupont-redhat :
> this will be the network of the hostname / IP address of the host in vSphere inventory. So, it's probable that it could be determined by some network calculation. Hacky and not sure how accurate it will be.

### LINKED ISSUE #421: Add the ability to unset the migration network
Once a migration network has been selected for a VMware host, it is only possible to select another network as migration network, not reset to no migration network.

It is possible that the users have invalid credentials and that until they retrieve valid credentials they want to fall back to the default behavior, i.e. connect to vCenter and use the management network. And this is currently not possible.

From a technical point of view, this would mean deleting the Host CR.

RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1929050

mturley:
Resolves #421 (https://bugzilla.redhat.com/show_bug.cgi?id=1929050)
Also resolves #292 (https://bugzilla.redhat.com/show_bug.cgi?id=1908034)

When selecting a migration network for VMware hosts, if the user chooses the management network (and the management network has the same IP for all selected hosts), the credential fields will be disabled (with a tooltip explaining they aren't needed) and upon submission the relevant Host CRs will be deleted, if any.

Screen Shot 2021-02-16 at 2 55 39 PM

Also, a host without a CR will now display "Management Network" with its correct bandwidth and MTU in the table instead of just "(default)".

Screen Shot 2021-02-16 at 3 06 16 PM

### DIFF
--- a/src/app/Providers/components/VMwareProviderHostsTable/SelectNetworkModal.tsx
+++ b/src/app/Providers/components/VMwareProviderHostsTable/SelectNetworkModal.tsx
@@ -11,7 +11,11 @@ import {
 import SimpleSelect, { OptionWithValue } from '@app/common/components/SimpleSelect';
 import { IHost, IHostConfig, IHostNetworkAdapter, IVMwareProvider } from '@app/queries/types';
 
-import { formatHostNetworkAdapter, usePrefillHostConfigEffect } from './helpers';
+import {
+  formatHostNetworkAdapter,
+  isManagementNetworkSelected,
+  usePrefillHostConfigEffect,
+} from './helpers';
 import { useConfigureHostsMutation } from '@app/queries';
 import { QuerySpinnerMode, ResolvedQuery } from '@app/common/components/ResolvedQuery';
 import LoadingEmptyState from '@app/common/components/LoadingEmptyState';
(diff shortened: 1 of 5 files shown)


mturley:
Converted to draft. Instead of relying on the name "Management Network" we should use an inventory property to confirm that a network is indeed the default. See #292

mturley:
Blocked by https://github.com/konveyor/forklift-controller/issues/120.

mturley:
Now unblocked by https://github.com/konveyor/forklift-controller/pull/170.

mturley:
@fdupont-redhat @gildub this one is now ready as well.

fdupont-redhat (approved):

PR #429, opened 2021-02-17
Bug 1929731: Allow partial string matches when filtering target namespace

mturley:
Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1929731

When selecting a target namespace in the plan wizard, partial matches are now supported in the filter.


mturley:
@fdupont-redhat I talked to @vconzola about the search term highlighting here, and he thinks we should only add that feature if we add it everywhere we have text filtering in the UI. I opened an issue for that: https://github.com/konveyor/forklift-ui/issues/430 and I'm reverting it here for now.

fdupont-redhat (approved):

PR #458, opened 2021-03-19
Warm migration followup: warn when selecting warm with incompatible VMs, show migration type in the Review step of the wizard

### LINKED ISSUE #454: Warn when selecting warm migration if incompatible VMs are selected
Separating this task from #308 since it is decoupled from the changes in PR #440 and because I'm waiting for the relevant validation service concern.

mturley:
Resolves #454

The Type step of the wizard now includes warnings under the "warm migration" option if any of the selected VMs have validation concerns relevant to warm migration (currently just a critical message when CBT not enabled).

VMs selected with a critical concern:
Screen Shot 2021-03-23 at 1 17 57 PM

If the validation service re-runs analysis on the selected VMs while the user is on this step (for example, because they updated the settings on the VMs in response to these warnings) a spinner will display in place of the warnings while the validation catches up:

Screen Shot 2021-03-23 at 1 53 37 PM

The Review step of the wizard now also shows the Migration Type at the bottom as part of the summary:

Screen Shot 2021-03-23 at 1 29 19 PM

### DIFF
--- a/src/app/Plans/components/VMMigrationDetails.tsx
+++ b/src/app/Plans/components/VMMigrationDetails.tsx
@@ -117,7 +117,7 @@ const VMMigrationDetails: React.FunctionComponent = () => {
             getTotalCopiedRatio(vmStatus).completed,
             getPipelineSummaryTitle(vmStatus),
           ]
-        : [vmStatus.warm?.precopies.length || 0, getWarmVMCopyState(vmStatus).state]),
+        : [vmStatus.warm?.precopies?.length || 0, getWarmVMCopyState(vmStatus).state]),
     ];
   };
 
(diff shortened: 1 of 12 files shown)


vconzola (commented):
Looks great, as always.

mturley:
@mperezco we wanted to run the look and feel of these warnings by you, what do you think?

mturley:
Removed the warning about snapshots based on discussions around that not being a real issue for warm migration.

mturley:
@gildub @fdupont-redhat this one is ready to go, and is the final feature PR (everything coming after this is a bug fix which may be ok to merge after Friday).

gildub (approved):
LGTM

PR #462, opened 2021-03-22
Update help text on Select VMs step of wizard to be consistent with 'migration assessment' column name

mturley:
Noticed in the usability study review meeting. Trivial fix, so there's no associated issue or BZ.


mturley:
Marking as draft for now until we confirm with Miguel that we want the column to be called "migration analysis" instead of "migration assessment"

mturley:
@vconzola confirmed the name should be "migration assessment".

gildub (approved):

PR #469, opened 2021-03-22
For QE: Add id to the create plan buttons, change the text to consistently say 'Create plan'

mturley:
Currently the plans page has a button "Create migration plan" on the empty state when there are no plans to display, but when there are plans, the button at the top of the table says "Create plan". This PR changes them to consistently say "Create plan" for both, and also adds an id to those buttons for ease of selection in QE tests.


gildub (approved):
LGTM

PR #479, opened 2021-03-24
Fix severity icon logic for StatusCondition component

### LINKED ISSUE #457: Fix logic for StatusCondition severity and icons when there are Required and/or Advisory conditions
As seen in this screenshot, conditions on a provider are not always assigned the correct severity in the UI.

Screen Shot 2021-03-11 at 12 40 48 PM

I need to double check the logic around Required conditions in particular, since the `category` property is Required rather than a severity value, so the severity must be inferred from other properties on the condition.

cc @jortel @vconzola

mturley:
Resolves #457 

The mapping of CR status conditions to severity icons is non-trivial, and this PR simplifies it a bit to correctly show a few of the cases that were confusing users. The new logic is (applied in order, the first matching case here is used):

* If the condition has the `category` "Critical" or "Error", the red error icon is shown.
* If the condition has the `category` "Warn", the yellow warning icon is shown.
* If the condition has the `reason` "Started" or "Running", a spinner is shown with the word "Pending".
* If the condition has the `type` "Ready", the `reason` "Completed" or the `category` "Required", the green check icon is shown.
* If the condition has the category "Advisory", the blue info icon is shown.

In the table, the condition matching the earliest case in this list is assumed to have the highest severity and is shown to summarize all conditions.

Screen Shot 2021-03-24 at 4 35 55 PM

Screen Shot 2021-03-24 at 4 36 05 PM

Screen Shot 2021-03-24 at 4 42 06 PM


gildub (approved):
LGTM

PR #499, opened 2021-04-01
Bug 1945567: Fix errors when saving a shared mapping from within the plan wizard

### LINKED ISSUE #495: [BZ#1945567] Migration plan creation - Not possible to save the network/storage mapping
https://bugzilla.redhat.com/show_bug.cgi?id=1945567

Something is going wrong with saving mappings for later in the plan wizard.

mturley:
Resolves #495 (https://bugzilla.redhat.com/show_bug.cgi?id=1945567)

Fixes a regression introduced in https://github.com/konveyor/forklift-ui/pull/432. When selecting the box to save a mapping for later, the `mappingName` entered in that field is erroneously passed into `getMappingFromBuilderItems` not only for the new shared mapping but for the new plan-owned mapping, and it overrides the `generateName` param. So the wizard attempts to create two mappings with the same name instead of using a generated name for the owned mapping. This PR makes sure that if a `generateName` arg is passed it will be used.


gildub (approved):
LGTM

PR #501, opened 2021-04-01
Plans table: Fix logic for canPlanBeStarted and isPlanBeingStarted states (corner cases for spinner and Restart button)

### LINKED ISSUE #494: Bug: plan restart button only appears if there were failed VMs, so there is no way to re-migrate canceled VMs if the rest of the plan succeeded
Per @jortel, the `Failed` condition on a plan will only be present if at least one VM in the plan failed with errors. If a plan finishes with only succeeded and canceled VMs, that condition will not be present. We currently use that condition to determine if we should show the "Retry" button on a migration plan.

A possible fix would be to show that button if the plan has `Failed` || (`Succeeded` && the number of VMs with `Succeeded` is less than the total number of VMs)

mturley:
Resolves #494


gildub (approved):
LGTM

PR #505, opened 2021-04-05
Bug 1942091: Re-enable polling in the General step of the plan wizard to pick up new namespace options

### LINKED ISSUE #464: [BZ#1942091] General step of the plan wizard: new namespaces created outside the UI are not picked up by polling
https://bugzilla.redhat.com/show_bug.cgi?id=1942091

Noticed in the usability study. If a new namespace is created in OCP while the plan wizard is open, it is not automatically added to the list of options in the target namespace dropdown. If the user cancels and reopens the wizard, the new namespace does appear. Need to make sure polling is behaving properly while on this wizard step.

mturley:
Resolves #464 (https://bugzilla.redhat.com/show_bug.cgi?id=1942091).

In https://github.com/konveyor/forklift-ui/pull/129, we disabled polling on all steps of the wizard (we later re-enabled it for the Select VMs step in https://github.com/konveyor/forklift-ui/pull/411 so the validation concerns could be updated automatically). This was done because some of the polled resources return immutable copies on each poll, and that was breaking selection logic that depended on referential equality. As a side-effect, newly created namespaces were not being picked up by the wizard automatically without reloading the wizard.

It is safe to re-enable polling on the General step because of changes since we disabled it that alleviate this equality issue for provider selections (https://github.com/konveyor/forklift-ui/pull/358). The equality issue was never present for the target namespace field since it uses string values.

This PR leaves polling disabled for the filter VMs step and the mapping steps, because it is uncertain whether re-enabling it will cause bugs there. It isn't as important for inventory resources to be updated automatically without restarting the wizard as it is for namespaces, so it's not worth the risk so close to the GA release. we can address that post-GA if we need to.

### DIFF
--- a/src/app/Plans/components/Wizard/GeneralForm.tsx
+++ b/src/app/Plans/components/Wizard/GeneralForm.tsx
@@ -22,7 +22,6 @@ import { useNamespacesQuery } from '@app/queries/namespaces';
 import { QuerySpinnerMode, ResolvedQueries } from '@app/common/components/ResolvedQuery';
 import ProviderSelect from '@app/common/components/ProviderSelect';
 import { ProviderType } from '@app/common/constants';
-import { usePausedPollingEffect } from '@app/common/context';
 import SelectOpenShiftNetworkModal from '@app/common/components/SelectOpenShiftNetworkModal';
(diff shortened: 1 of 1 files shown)


gildub (approved):
LGTM

PR #508, opened 2021-04-05
[WIP: hacky workaround] Retain the value typed in the target namespace field when clicking away

### LINKED ISSUE #463: [BZ#1943373] General step of wizard: usability issues with creating a new target namespace
https://bugzilla.redhat.com/show_bug.cgi?id=1943373

Observations from the usability study:
* When typing in the target namespace field (typeahead search), one of the options is "Create" which can be clicked to create a namespace from the search text instead of selecting an existing one.
* Clicking away from the field without choosing a dropdown option will clear the field -- this was unexpected and frustrating because users assumed it behaved like a normal text input field. It's not obvious that you need to click on the dropdown item to retain the name you type.
* Users also tried to press Enter to select the 'Create' option which doesn't work either.

This may be an issue in the way the PatternFly Typeahead Select component works. But if we can set an event handler on the change event of that search text field, we could have the form state retain whatever they type in there unless they click one of the options.

@vconzola, if the user just types something in the field and clicks away, do you think it is obvious enough that a new namespace will be created? Should we maybe add some text that appears below to indicate that the namespace in the field does not exist and will be created?

### LINKED ISSUE #507: Target namespace field: validation errors are not appearing when creating a new namespace with an invalid name
Related to #463.

mturley:
WIP. May resolve https://github.com/konveyor/forklift-ui/issues/463 (https://bugzilla.redhat.com/show_bug.cgi?id=1943373).

This is a rough one. We may want to consider pushing this fix to post-GA, or removing the ability to create a new namespace from the plan wizard.

The PatternFly Select component does not support retaining the search text for its typeahead variant when the user leaves the field. In order to get this behavior, we have to use the `onTypeaheadInputChanged` prop introduced in https://github.com/patternfly/patternfly-react/pull/5375, which means upgrading to the latest version of PatternFly. There are a few problems with this:
* The upgrade makes it possible to fix this bug, but it involves a hacky workaround for a PF regression I found that is introduced by the upgrade: https://github.com/patternfly/patternfly-react/issues/5625
* Upgrading PatternFly also breaks a few other things in our UI, particularly where we had to work around other PF bugs that are now fixed. Removing these workarounds is non-trivial.
* It would be better to implement the desired behavior as part of the PatternFly Select component itself, instead of relying on `onTypeaheadInputChanged` and working around the issues it presents. We can't get that done before GA.

In its current state this PR fixes #463, but it's hacky and not ideal, and it does not yet address other issues caused by upgrading PF.

I would prefer to remove the ability to create a new namespace in the wizard for GA and reintroduce it later. This would alleviate the confusion (and users can still create a namespace outside our UI and have it automatically appear here). It would also avoid the need to address #507 right away.


mturley:
Closing this to revisit post-GA.

PR #509, opened 2021-04-06
Fix various minor issues with the target namespace field in the plan wizard

### LINKED ISSUE #507: Target namespace field: validation errors are not appearing when creating a new namespace with an invalid name

mturley:
* Fixes #507: Validation errors were not appearing on the target namespace field when creating a new namespace with an invalid name
* Regex escape characters (`\`) in the search box were causing errors, now they just produce 0 matches
* The Clear (x) button was missing from the field
* Pressing Enter in the field was triggering the popover on the help icon for the transfer network field (weird ref issue with the Popover component, switching the HTML button to a PF Button fixed it)

Does not address the remaining issues in https://github.com/konveyor/forklift-ui/issues/463


fdupont-redhat (approved):
LGTM

PR #511, opened 2021-04-06
Inform the user on the Review step if a new target namespace is being created

mturley:
This is a piece of #463 that we decided not to delay to post-GA since it's low-risk and helpful.

Screen Shot 2021-04-06 at 2 58 32 PM

Also, this PR fixes a regression introduced by #505 where polling while the target namespace field is open and search text has been entered will cause the filter results to reset. Disables polling only while the dropdown is open, so you can still pick up new namespaces by closing the dropdown and reopening it. (in fact, when closing it the polling cycle is restarted immediately so there is no need to wait for the next cycle, new namespaces will be fetched immediately)


mturley:
Marking as draft while I double-check the message-- not sure if namespace is created when the plan is saved or when the migration starts.

mturley:
Confirmed with Jeff that the namespace is created when the plan starts, not when it is created. Updated text, ready for review.

fdupont-redhat (approved):

PR #512, opened 2021-04-06
[Backport] Inform the user on the Review step if a new target namespace is being created (#511)

mturley:
Backports #511


mturley:
Marking as draft while I double-check the message-- not sure if namespace is created when the plan is saved or when the migration starts.

mturley:
Confirmed with Jeff that the namespace is created when the plan starts, not when it is created. Updated text, ready for review.

fdupont-redhat (approved):

PR #524, opened 2021-04-07
Bug 1942137: Add show/hide password feature via new ValidatedPasswordField component

### LINKED ISSUE #452: "Add provider" modal - "Certificate SHA1 Fingerprint" field should be called "SHA-1 fingerprint"
image

The "Hostname or IP address" field is required but does not display the validation message "... is a required field" when it is skipped.

Just noticed another small fix: "Fingerprint" should be lower case.

Suggestion about the "Certificate SHA1 Fingerprint" label. It's a little odd, because it is actually the SHA-1 (this is how VMware always styles the algorithm) fingerprint of the SSL certificate of the ESXi host. "Certificate" does not quite work as an adjective in this context.

You might want to change this to "SHA-1 fingerprint" because all of these fields are for the VMware ESXi host.

### LINKED ISSUE #468: [BZ#1942137] Add button for showing/hiding passwords and tokens
https://bugzilla.redhat.com/show_bug.cgi?id=1942137

Noticed in the usability study. When editing providers and host configurations, passwords and other secrets are still shown as asterisks in the text field. It would make editing easier if there was a button to toggle showing that text in the field instead of the asterisks.

This has been implemented in MTC, we should look at how that was done (may be a simple prop on the PF component).

mturley:
Resolves #468 (https://bugzilla.redhat.com/show_bug.cgi?id=1942137)
Also resolves #452 in the interest of time, because separate PRs for these two fixes would conflict.

Adds a new common component `ValidatedPasswordField` which adds a show/hide icon button on the right side of the field. Replaces the plain password input with this component in 3 places: VMware provider password, OpenShift provider SA token, and ESXi host admin password. 

The first commit puts the toggle icon next to the field label like it's done in MTC today, but then @vconzola pointed out that the PatternFly login page example has the icon in an InputGroup next to the input field (see https://github.com/patternfly/patternfly-react/pull/5481), so the second commit switches to that design.

I had to duplicate some of the code from the lib-ui ValidatedTextInput component since it doesn't support InputGroups. We can add that feature to lib-ui and simplify this code in the future. Opened https://github.com/konveyor/lib-ui/issues/55 to track this.

xvCQrJKku7

### DIFF
--- a/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
+++ b/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
@@ -310,22 +310,39 @@ const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModalProps> 
                     ),
                   }}
                 />
-
                 <ValidatedPasswordInput
                   field={openshiftForm.fields.saToken}
                   label="Service account token"
                   isRequired
                   fieldId="openshift-sa-token"
-                  helpContent={
-                    <>
-                      To obtain SA token, run the following command:
-                      <br />
-                      <i>$ oc serviceaccounts get-token serviceaccount_name -n namespace_name</i>
-                      <br />
-                      <br />
-                      <b>** Be sure to use the namespace in which you created the SA.</b>
-                    </>
-                  }
+                  formGroupProps={{
+                    labelIcon: (
+                      <Popover
+                        bodyContent={
+                          <>
+                            To obtain SA token, run the following command:
+                            <br />
+                            <i>
+                              $ oc serviceaccounts get-token serviceaccount_name -n namespace_name
+                            </i>
+                            <br />
+                            <br />
+                            <b>** Be sure to use the namespace in which you created the SA.</b>
+                          </>
+                        }
+                      >
+                        <Button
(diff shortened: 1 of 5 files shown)


fdupont-redhat (approved):

src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
```diff
@@ -214,16 +216,15 @@ const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModalProps>
...
                   field={vmwareForm.fields.password}
-                  type="password"
                   label="Password"
                   isRequired
                   fieldId="vmware-password"
                 />
                 <ValidatedTextInput
                   field={vmwareForm.fields.fingerprint}
-                  label="Certificate SHA1 Fingerprint"
+                  label="SHA-1 Fingerprint"
```
apinnick, line 227: "Fingerprint" should be "fingerprint", per PF4 guidelines.

mturley: Agh dammit, thanks @apinnick. Will fix in another PR.

PR #526, opened 2021-04-07
Bug 1942135: Remove provider name from title on hosts page to avoid inconsistency with breadcrumb bar

### LINKED ISSUE #460: [BZ#1942135] Breadcrumb bar on VMware Hosts page conflicts with page title
https://bugzilla.redhat.com/show_bug.cgi?id=1942135

From the usability study (https://docs.google.com/presentation/d/1_GdHUjugbX-CB_P3u12qQmfDbSO99Fo_QE0g1CWPfH8/edit?ts=6058cc49#slide=id.gc70a21fdbb_0_36)

> The “Hosts” breadcrumb hierarchy was confusing
> * One user had difficulty interpreting the “Hosts” breadcrumb, where it showed “host_name > Hosts” rather than “Hosts > host_name”. The breadcrumb was also in direct conflict with the page title.
> * Consider adjusting the order of the hosts breadcrumb to reflect the hierarchy. 

Screen Shot 2021-03-22 at 1 07 50 PM

As described, the issue is not quite accurate - the breadcrumb bar does not show a "host_name > Hosts", it shows "provider_name > Hosts". The page title however shows "Hosts - provider_name", so I think the last item in the breadcrumb bar should show this instead of having Hosts be a separate item.

"Providers > VMware > Hosts - provider_name"

cc @vconzola

mturley:
Resolves #460 (https://bugzilla.redhat.com/show_bug.cgi?id=1942135)

### DIFF
--- a/src/app/Providers/HostsPage.tsx
+++ b/src/app/Providers/HostsPage.tsx
@@ -56,7 +56,7 @@ export const HostsPage: React.FunctionComponent = () => {
         </Level>
         <Level className={spacing.mtLg}>
           <LevelItem>
-            <Title headingLevel="h1">Hosts - {match?.params.providerName}</Title>
+            <Title headingLevel="h1">Hosts</Title>
           </LevelItem>
         </Level>
       </PageSection>


vconzola (commented):
@mturley In looking at this in demo mode the breadcrumb change LGTM. But I notice that only one of the VMware providers has the number of hosts as a clickable link. Is this a bug or an artifact of it being demo mode?

fdupont-redhat (approved):

mturley:
@vconzola it's not a bug, that provider is mocking the state where a provider's inventory data isn't ready (it also has Critical under status). In a real environment that link would re-enable once the provider is ready.

apinnick:
Those pesky providers.... :-)

PR #530, opened 2021-04-07
Add provider modal: update field name for SHA-1 Fingerprint

### LINKED ISSUE #452: "Add provider" modal - "Certificate SHA1 Fingerprint" field should be called "SHA-1 fingerprint"

mturley:
Resolves #452


mturley:
Moving this change to be part of https://github.com/konveyor/forklift-ui/pull/524 to prevent conflicts

PR #533, opened 2021-04-08
Correctly handle plan status when it finishes with some VMs canceled

mturley:
This is a corner case I missed when considering canceled VMs in https://github.com/konveyor/forklift-ui/pull/501.


gildub (approved):
LGTM

PR #537, opened 2021-04-08
Bug 1947519: Track migrate/cutover request state locally to the button instead of at the plans page level

mturley:
Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1947519

Currently when clicking the Start, Restart or Cutover buttons on a plan, all the action buttons for all plans turn grey while the CR is being created/mutated. This is because there is one query hook at the plans page level for all plans, since we display any errors with these requests at the top of the plans page. This is confusing because greying out all the buttons on unrelated plans looks weird, and because it isn't apparent that the action is working.

This PR factors out a `MigrateOrCutoverButton` component which contains its own `useCreateMigrationMutation` and `useSetCutoverMutation` hooks, allowing us to track that request state at a per-plan level. To achieve the centralized error container we had before, a React Portal is used which allows the error child of the button component to be rendered at the top of the plans page. Also, instead of greying out the button during the request, it is replaced by a spinner to be more clear that something is happening.

Along with https://github.com/konveyor/forklift-ui/pull/531, this should hopefully clear up any confusion during the transition while a migration is being started.

Although we can't start migrations in mock mode, the spinner behavior and error reporting can still be tested there.

### DIFF
--- a/src/app/Plans/PlansPage.tsx
+++ b/src/app/Plans/PlansPage.tsx
@@ -10,37 +10,23 @@ import {
 } from '@patternfly/react-core';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import { PlusCircleIcon } from '@patternfly/react-icons';
-import { useHistory } from 'react-router-dom';
 
 import {
   useHasSufficientProvidersQuery,
   usePlansQuery,
-  useCreateMigrationMutation,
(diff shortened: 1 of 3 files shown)


vconzola (commented):
@mturley I really like this change. Looks great.

ibolton336 (approved):
VISACK

PR #541, opened 2021-04-08
Make status column clickable while plan is running, navigates to migration details

### LINKED ISSUE #521: Make the progress bar cell clickable instead of having a tooltip to direct the user to click the plan name?
This is a better solution to #443, but since we need to evaluate the UX (hover styles? which area is clickable? maybe the entire row?) I'm saving it for later and moving forward with the tooltip fix.

mturley:
Resolves #521. Nice-to-have.


ibolton336:
VISACK

ibolton336 (approved):

vconzola:
@ibolton336 We should think about doing this in MTC as well to address John/Todd's concern about it not being intuitive how to drill down into a migration.

mturley:
Rebased and resolved a conflict

PR #546, opened 2021-04-20
Update management network identification

fdupont-redhat:
The `managementServerIp` attribute of the host is not the address of the host on the management network, but the address of the vCenter that manages the host. To identify the management network, we need to identify the item in the host `networkAdapters` array which `ipAddress` + `subnetMask` matches the `managementServerIp`.

This pull request updates the logic to compute the network range based on the `ipAddress` and `subnetMask` for each network adapter and pick the one that contains `managementServerIp`.


mturley (approved):
Thanks for the fix @fdupont-redhat ! LGTM

Issue #550, opened 2021-04-21
Workaround: Custom CSS height calculation necessary to keep wizard body contents from creating a double scrollbar.


mturley:
There is a custom max-height to the wizard body container in order to prevent there ever being "double scrollbars" where two nested containers both have overflowing content relative to their height. This was a workaround introduced in https://github.com/konveyor/forklift-ui/pull/551. Without this style hack, this can happen:

<img width="1282" alt="Screen Shot 2021-04-21 at 2 21 21 PM" src="https://user-images.githubusercontent.com/811963/115602932-889d8a80-a2ad-11eb-968f-64739ceea76e.png">

We need to remove this custom CSS and find the real solution to the bug it works around.

yaacov:
closing as new development is moving to:
https://github.com/kubev2v/forklift-console-plugin

PR #551, opened 2021-04-21
Hack around weird double scrollbar bug

mturley:
This PR applies a custom `max-height` to the wizard body container in order to prevent there ever being "double scrollbars" where two nested containers both have overflowing content relative to their height. See https://github.com/konveyor/forklift-ui/issues/550 for a description of the bug. The arbitrary `286px` in the calculation is the total height of the app header, page section header, and wizard footer which are all static heights.

We need to remove this custom CSS and find the real solution to the bug it works around. Capturing that in https://github.com/konveyor/forklift-ui/issues/550.

### DIFF
--- a/src/app/Plans/components/Wizard/PlanWizard.css
+++ b/src/app/Plans/components/Wizard/PlanWizard.css
@@ -1,0 +1,6 @@
+.pf-c-wizard__main-body {
+  /* Fixes double scrollbar when wizard body contents are tall enough for scrolling. */
+  /* This should absolutely not be necessary, there is something funky going on here with heights and overflow. */
+  /* See https://github.com/konveyor/forklift-ui/issues/550 */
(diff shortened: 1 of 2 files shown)


vconzola (commented):
@mturley LGTM for removing double scrollbar. Just to be clear though, if there are 100 namespaces in the dropdown the whole page will just continue to scroll, correct? Or is there a max height for the dropdown after which a second scrollbar will appear in the dropdown itself?

mturley:
@vconzola that's correct. The page grows in height to match the height of the entire menu in the dropdown so that you can scroll to the bottom. The limitation is on the height of the container around that because something weird is going on with its padding. The second scrollbar should never appear no matter how many namespaces and no matter the size of the window.

fdupont-redhat (approved):

PR #564, opened 2021-04-29
For QE: Add ids and aria label attributes to fields in Review step of wizard

mturley:
Resolves #544 

Changes requested by @ibragins 

Field values can be identified by:
```
#review-plan-name
#review-plan-description
#review-source-provider
#review-target-provider
#review-target-namespace
#review-transfer-network
#review-selected-vms-count
#review-network-mapping
#review-storage-mapping
#review-migration-type
```

The dynamic parts of the page can also be identified:
* If the target namespace does not exist, the message showing "this is a new namespace..." will be present with the id `#review-new-target-namespace-message`.
* If you click on `button#review-selected-vms-count` a popover will appear containing the id `#review-selected-vms-list`.

This PR also adds `aria-labelledby` attributes pointing to the labels in the left column for screen-readers.

PR #565, opened 2021-04-29
[Backport] For QE: Add ids and aria label attributes to fields in Review step of wizard (#564)

mturley:
Backports #564. Not a critical fix, but will assist QE in testing the GA release so it's worth backporting.


fdupont-redhat (approved):
Minor change, with great potential for quality. I agree with backporting.

PR #566, opened 2021-04-29
Major dependency upgrades: React, TypeScript, utils, build tools, quality tools + install react-hot-loader

### LINKED ISSUE #162: Update deps and 404 handling to match upstream changes in patternfly-react-seed
The seed project this repo is based on is getting some updates, we should keep up to date with its practices. https://github.com/patternfly/patternfly-react-seed/pull/99

### LINKED ISSUE #441: Remove dead code: highlighting empty mapping targets
Before we added the functionality of always selecting default mapping targets when building a mapping, there was a feature where additional mapping items added when selecting an existing mapping would have their empty target dropdown highlighted in yellow to call the user's attention to them. This will never happen now, so the related code is just cruft.

The `highlight` property in `IMappingBuilderItem` can be removed, and any logic based on it (setting it, and setting the `.highlighted` className used by styles in `MappingBuilder.css`, which can also be removed)

### LINKED ISSUE #554: Style issues after PatternFly upgrade
We may need to adjust something in our CSS imports for PatternFly. After https://github.com/konveyor/forklift-ui/pull/545, certain styles look broken. The first I've noticed is the lack of a border around Select components:

Screen Shot 2021-04-26 at 12 09 31 PM

Compare to the previous version:

Screen Shot 2021-04-26 at 12 09 34 PM

Preview where the issue can be observed: http://konveyor-forklift-ui-pr-545-preview.surge.sh/
Preview from earlier, without the issue: http://konveyor-forklift-ui-pr-551-preview.surge.sh/

This is not a GA blocker, the offending change is not backported to v2.0.0.

mturley:
Resolves #162 
Resolves #441
Resolves #554 (there was a mismatch in PF versions between lib-ui and forklift)

Taking advantage of bug fixes and new features available in most of our dependencies. Keeping things fresh and clean ✨

One huge benefit for DX here: with react-hot-loader configured, the webpack dev server reloads MUCH faster when files are changed than it did before. For a basic JSX change I noticed a drop from ~10-15 seconds down to ~2-3.

The only major dependency this PR does not upgrade is React Query, because [the breaking changes in Reacy Query v3 are much more complex](https://react-query.tanstack.com/guides/migrating-to-react-query-3) and would fit better in their own PR. New issue for that: https://github.com/konveyor/forklift-ui/issues/567


fdupont-redhat (approved):
I cannot judge the code, but keeping things up-to-date is important.

mturley:
Thanks Fabien. There's not much code to judge, these changes are mostly lint and build config stuff. The one exception is the `yup` upgrade, to verify those changes we just need to make sure the form validation in the plan wizard still works (which I've done). Otherwise as long as it builds and passes CI it should be good. 🚿

PR #568, opened 2021-05-02
Use API version v1beta1

fdupont-redhat:
With [forklift-controller#238](https://github.com/konveyor/forklift-controller/pull/238) merged, the API version has changed to `v1beta1`. So, the API fails to retrieve the inventory and to create new custom resources. This merge request updates the UI to use `v1beta1` API version;


fdupont-redhat:
Checked in a real cluster. Self-merging.

PR #573, opened 2021-05-03
Bug 1956268: Don't show 'Running cutover' if cutover time is in the future

mturley:
Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1956268


gildub (approved):
LGTM

Issue #571, opened 2021-05-03
Hooks feature: layout of main Hooks page and table


mturley:
Mockups: https://docs.google.com/presentation/d/1FnAK9Eh-mowADHLtDWqlD9__VQyeHserfPzeWfeg66E/edit#slide=id.gd31781e05c_0_0

PR #574, opened 2021-05-04
Hooks layout & scaffolding

gildub:
Screenshot from 2021-05-04 19-17-30

Fix https://github.com/konveyor/forklift-ui/issues/571

### DIFF
--- a/src/app/Hooks/HooksPage.tsx
+++ b/src/app/Hooks/HooksPage.tsx
@@ -1,0 +1,93 @@
+import * as React from 'react';
+import {
+  Card,
+  PageSection,
+  CardBody,
+  EmptyState,
+  EmptyStateIcon,
+  EmptyStateBody,
+  Title,
+  Button,
+} from '@patternfly/react-core';
+import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
+import { PlusCircleIcon } from '@patternfly/react-icons';
+
+import {
+  useHasSufficientProvidersQuery,
+  useHooksQuery,
+  useClusterProvidersQuery,
+} from '@app/queries';
+
+import HooksTable from './components/HooksTable';
+import { ResolvedQueries } from '@app/common/components/ResolvedQuery';
+import { IHook } from '@app/queries/types';
+import AddEditHookModal from './components/AddEditHookModal';
+
+const HooksPage: React.FunctionComponent = () => {
+  const sufficientProvidersQuery = useHasSufficientProvidersQuery();
+  const clusterProvidersQuery = useClusterProvidersQuery();
+  const hooksQuery = useHooksQuery();
+
+  const [isAddEditModalOpen, toggleAddEditModal] = React.useReducer((isOpen) => !isOpen, false);
+  const [hookBeingEdited, setHookBeingEdited] = React.useState<IHook | null>(null);
+
+  const toggleModalAndResetEdit = () => {
+    setHookBeingEdited(null);
+    toggleAddEditModal();
+  };
+
+  const openEditHookModal = (hook: IHook) => {
+    setHookBeingEdited(hook);
+    toggleAddEditModal();
+  };
+
+  return (
+    <>
+      <PageSection variant="light">
+        <Title headingLevel="h1">Hooks</Title>
+      </PageSection>
+      <PageSection>
+        <ResolvedQueries
+          results={[sufficientProvidersQuery.result, clusterProvidersQuery, hooksQuery]}
+          errorTitles={[
+            'Error loading hooks',
+            'Error loading hooks from cluster',
+            'Error loading plans',
+          ]}
+          errorsInline={false}
+        >
+          <Card>
+            <CardBody>
+              {!hooksQuery.data ? null : hooksQuery.data.items.length === 0 ? (
+                <EmptyState className={spacing.my_2xl}>
+                  <EmptyStateIcon icon={PlusCircleIcon} />
+                  <Title size="lg" headingLevel="h2">
+                    No hooks
+                  </Title>
+                  <EmptyStateBody>
+                    Hooks are contained in Ansible playbooks that can be run before or after the
+                    migration. They are added to a migration plan during plan creation.
+                  </EmptyStateBody>
+                  <Button onClick={() => toggleModalAndResetEdit()} id="create-hooks-button">
+                    Create hook
+                  </Button>
+                </EmptyState>
+              ) : (
+                <HooksTable
+                  hooks={hooksQuery.data?.items || []}
+                  openCreateHookModal={toggleModalAndResetEdit}
+                  openEditHookModal={openEditHookModal}
+                />
+              )}
+            </CardBody>
+          </Card>
+        </ResolvedQueries>
+      </PageSection>
+      {isAddEditModalOpen ? (
+        <AddEditHookModal onClose={toggleModalAndResetEdit} hookBeingEdited={hookBeingEdited} />
+      ) : null}
+    </>
+  );
+};
+
+export default HooksPage;
--- a/src/app/Hooks/HooksPage.tsx
+++ b/src/app/Hooks/HooksPage.tsx
@@ -47,11 +47,7 @@ const HooksPage: React.FunctionComponent = () => {
       <PageSection>
         <ResolvedQueries
           results={[hooksQuery]}
-          errorTitles={[
-            'Error loading hooks',
-            'Error loading hooks from cluster',
-            'Error loading plans',
-          ]}
+          errorTitles={['Error loading hooks']}
           errorsInline={false}
         >
           <Card>
--- a/src/app/Hooks/HooksPage.tsx
+++ b/src/app/Hooks/HooksPage.tsx
@@ -24,8 +24,6 @@ import { IHook } from '@app/queries/types';
 import AddEditHookModal from './components/AddEditHookModal';
 
 const HooksPage: React.FunctionComponent = () => {
-  const sufficientProvidersQuery = useHasSufficientProvidersQuery();
-  const clusterProvidersQuery = useClusterProvidersQuery();
   const hooksQuery = useHooksQuery();
 
   const [isAddEditModalOpen, toggleAddEditModal] = React.useReducer((isOpen) => !isOpen, false);
@@ -48,7 +46,7 @@ const HooksPage: React.FunctionComponent = () => {
       </PageSection>
       <PageSection>
         <ResolvedQueries
-          results={[sufficientProvidersQuery.result, clusterProvidersQuery, hooksQuery]}
+          results={[hooksQuery]}
           errorTitles={[
             'Error loading hooks',
             'Error loading hooks from cluster',
--- a/src/app/Hooks/HooksPage.tsx
+++ b/src/app/Hooks/HooksPage.tsx
@@ -12,11 +12,7 @@ import {
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import { PlusCircleIcon } from '@patternfly/react-icons';
 
-import {
-  useHasSufficientProvidersQuery,
-  useHooksQuery,
-  useClusterProvidersQuery,
-} from '@app/queries';
+import { useHooksQuery } from '@app/queries';
 
 import HooksTable from './components/HooksTable';
 import { ResolvedQueries } from '@app/common/components/ResolvedQuery';
--- a/src/app/Hooks/components/AddEditHookModal.tsx
+++ b/src/app/Hooks/components/AddEditHookModal.tsx
@@ -1,0 +1,151 @@
+import * as React from 'react';
+import * as yup from 'yup';
+import { Modal, Button, Form, Flex, Stack, Popover } from '@patternfly/react-core';
+import { useFormState, useFormField, ValidatedTextInput } from '@konveyor/lib-ui';
+
+import { urlSchema } from '@app/common/constants';
+import { usePausedPollingEffect } from '@app/common/context';
+import {
+  getHookNameSchema,
+  useCreateHookMutation,
+  usePatchHookMutation,
+  useHooksQuery,
+} from '@app/queries';
+
+import { IHook } from '@app/queries/types';
+import { QueryResult } from 'react-query';
+import { HelpIcon } from '@patternfly/react-icons';
+import { QuerySpinnerMode, ResolvedQuery } from '@app/common/components/ResolvedQuery';
+import { IKubeList } from '@app/client/types';
+import { useEditHookPrefillEffect } from './helpers';
+import LoadingEmptyState from '@app/common/components/LoadingEmptyState';
+
+interface IAddEditHookModalProps {
+  onClose: () => void;
+  hookBeingEdited: IHook | null;
+}
+
+const useHookFormState = (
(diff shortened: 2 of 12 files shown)


src/app/Hooks/HooksPage.tsx
```diff
@@ -0,0 +1,93 @@
...
+} from '@app/queries';
+
+import HooksTable from './components/HooksTable';
+import { ResolvedQueries } from '@app/common/components/ResolvedQuery';
+import { IHook } from '@app/queries/types';
+import AddEditHookModal from './components/AddEditHookModal';
+
+const HooksPage: React.FunctionComponent = () => {
+  const sufficientProvidersQuery = useHasSufficientProvidersQuery();
+  const clusterProvidersQuery = useClusterProvidersQuery();
```
mturley, line 28: I don't think we need these providers queries in here necessarily, right?

gildub: You're right, that's left over ;)

src/app/Hooks/HooksPage.tsx
```diff
@@ -0,0 +1,91 @@
...
+      <PageSection>
+        <ResolvedQueries
+          results={[hooksQuery]}
+          errorTitles={[
+            'Error loading hooks',
+            'Error loading hooks from cluster',
+            'Error loading plans',
+          ]}
+          errorsInline={false}
+        >
```
mturley, line 56: @gildub I'm getting errors trying to load the page because the length of `results` and `errorTitles` is not the same here. Makes me think I should have changed the props API for ResolvedQueries so TS could catch this.. but can you change it so there is only one `errorTitle` per `result`? (for now, just `Error loading hooks`)

Screen Shot 2021-05-05 at 8 51 13 AM

gildub: @mturley, sure, done.

Although we'll need the plans to be retrieved to find out any dependencies with the hooks.

gildub:
@vconzola, please review. All screens are availlable with Surge preview of this PR but the empty hook for HooksPage which I added to this PR's description.

vconzola:
@gildub The only things I see concern the create hook button. The label should just be "Create" (context is obvious from the page title), and it's not vertically aligned with the filter and pagination controls. Otherwise, LGTM.

vconzola:
@gildub Yeah, sorry about that - my mistake. The button label should stay as "Create hook" on the empty state page, but change to "Create" above the table.

mturley:
@vconzola our other pages have that button with the resource type in it ("create plan/mapping" and "add provider"). Should we change those too?

vconzola:
SoB. I don't know why I did that. It's easier to make hooks be consistent than to change a bunch of other stuff. So let's keep the button text as "Create hook". Sorry for the churn @gildub .

src/app/Hooks/components/HookActionsDropdown.tsx
```diff
@@ -0,0 +1,100 @@
...
+        aria-label="Actions"
+        toggle={<KebabToggle onToggle={() => setKebabIsOpen(!kebabIsOpen)} />}
+        isOpen={kebabIsOpen}
+        isPlain
+        dropdownItems={[
+          <ConditionalTooltip
+            key="edit"
+            isTooltipEnabled={isEditDeleteDisabled}
+            content={
+              !hook.spec.url
```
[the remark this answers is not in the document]
mturley: Note: see above, maybe we do need to block editing sometimes after all (if the hook is associated with existing plans). Waiting for confirmation from Jeff and Vince on Slack.

src/app/Hooks/components/HookActionsDropdown.tsx
```diff
@@ -0,0 +1,100 @@
...
+              onClick={() => {
+                setKebabIsOpen(false);
+                openEditHookModal(hook);
+              }}
+              isDisabled={isEditDeleteDisabled}
+            >
+              Edit
+            </DropdownItem>
+          </ConditionalTooltip>,
+          <ConditionalTooltip
```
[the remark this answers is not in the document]
mturley: Note: see above, maybe we do need this after all, waiting for confirmation on Slack.

src/app/Hooks/components/HookActionsDropdown.tsx
```diff
@@ -0,0 +1,100 @@
...
+            key="delete"
+            isTooltipEnabled={isEditDeleteDisabled}
+            content={
+              !hook.spec.url
+                ? 'The hook cannot be deleted'
+                : hasRunningMigration
+                ? 'This hook cannot be removed because it has running migrations'
```
mturley, line 63: If we are indeed keeping this behavior, this would probably be better phrased as something like "This hook cannot be removed because it is being used for running migrations". @vconzola

src/app/Hooks/components/HookActionsDropdown.tsx
```diff
@@ -0,0 +1,100 @@
...
+        mutateResult={deleteHookResult}
+        title="Remove hook"
+        confirmButtonText="Remove"
+        body={
+          <>
+            Are you sure you want to remove hook &quot;
+            <strong>{hook.metadata.name}</strong>&quot;?
+          </>
+        }
+        errorText="Error removing provider"
```
mturley, line 94: If we are keeping the shared hook by-reference behavior, we'll need to add logic here that explains deleting the hook will remove it from the associated plans (listing them). Beyond the scope of this PR though probably.

src/app/Hooks/components/HooksTable.tsx
```diff
@@ -0,0 +1,156 @@
...
+      },
+    },
+  ];
+
+  const { filterValues, setFilterValues, filteredItems } = useFilterState(hooks, filterCategories);
+  const getSortValues = (hook: IHook) => [
+    hook.metadata.name,
+    hook.spec.url,
+    hook.spec.branch,
+    '', // plans column,
```
mturley, line 69: If we are keeping this column (I assume we'd drop it if we drop the pass-by-reference behavior), we should leave a TODO so we remember to put a real sort value here (the number of associated plans). The values in that column will be a pain to get actually, we'll have to look them up from a plans query.

src/app/Hooks/components/HooksTable.tsx
```diff
@@ -0,0 +1,156 @@
...
+    { title: 'Branch', transforms: [sortable] },
+    { title: 'Plans', transforms: [cellWidth(20)] },
+    {
+      title: '',
+      transforms: [cellWidth(10)],
+      columnTransforms: [classNames(alignment.textAlignRight)],
+    },
+  ];
+
+  const rows: IRow[] = [];
```
mturley, line 89: Doesn't really matter, but since there are no conditionally-pushed rows in this array (like we have in expandable tables), this could just be `const rows: IRow[] = currentPageItemsa.map(...)` instead of using the `forEach / push`.

src/app/Hooks/components/HooksTable.tsx
```diff
@@ -0,0 +1,156 @@
...
+              spacer={{ default: 'spacerNone' }}
+            >
+              <FilterToolbar<IHook>
+                filterCategories={filterCategories}
+                filterValues={filterValues}
+                setFilterValues={setFilterValues}
+              />
+            </FlexItem>
+            <FlexItem>
+              <div>
```
mturley, line 124: Looking at `CreatePlanButton` and `CreateMappingButton`, to align this button with the filter toolbar we may need `className={spacing.mtMd}` on this div. Seems weird though, I'm not sure why it was necessary in those cases. Maybe some props on the `FlexItem` to align it vertically would work too, in which case we should do the same in the other pages. I'm probably being a perfectionist though, the margin class is fine.

src/app/Hooks/components/helpers.ts
```diff
@@ -0,0 +1,45 @@
...
+    url: form.values.url,
+    branch: form.values.branch,
+  },
+});
+
+interface IEditHookPrefillEffect {
+  isDonePrefilling: boolean;
+}
+
+export const useEditHookPrefillEffect = (
```
mturley, line 23: Nice work getting all this scaffolding in. We definitely will need all this prefilling stuff, etc. Although since this prefill doesn't require waiting for any queries, we may not need the `isDonePrefilling` logic... eh. We can simplify later, there's probably some other reason I did that. 🙃

src/app/queries/hooks.ts
```diff
@@ -0,0 +1,126 @@
...
+      config: {
+        refetchInterval: usePollingContext().refetchInterval,
+      },
+    },
+    mockKubeList(MOCK_HOOKS, 'Hook')
+  );
+  return sortKubeResultsByName(result);
+};
+
+export const useCreateHookMutation = (
```
mturley, line 37: This is all great, thanks for knocking it out! It is making me think we could use some more abstraction on these basic CRUD hooks, but I'm getting ahead of myself again. ☕ ☕

src/app/queries/mocks/hooks.mock.ts
```diff
@@ -0,0 +1,44 @@
...
+    uid: '28fde094-b667-4d21-8f29-28c18f22178b',
+    creationTimestamp: '2021-05-04T19:40:50Z',
+  },
+  spec: {
+    url: 'https://git.example.com/hooks/hooktest-02',
+    branch: 'feature',
+  },
+};
+
+if (process.env.NODE_ENV === 'test' || process.env.DATA_SOURCE === 'mock') {
```
mturley, line 42: I think if we put this `if()` block around the `const` declarations above too, webpack will tree-shake away that whole block of code in prod mode. I'm not sure if that is actually working correctly, but that was the goal of these conditionals, to slim down the prod bundle which doesn't need the mock data.

mturley (changes requested):
Looking good @gildub! Just a few comments.

mturley:
@vconzola it's pretty trivial to change the button text, if you want us to do it right on the other pages.

gildub:
@vconzola, not a worry, let's keep the button in the HookTable as "Create Hook" then.

src/app/Hooks/HooksPage.tsx
```diff
@@ -0,0 +1,87 @@
...
+          <Card>
+            <CardBody>
+              {!hooksQuery.data ? null : hooksQuery.data.items.length === 0 ? (
+                <EmptyState className={spacing.my_2xl}>
+                  <EmptyStateIcon icon={PlusCircleIcon} />
+                  <Title size="lg" headingLevel="h2">
+                    No hooks
+                  </Title>
+                  <EmptyStateBody>
+                    Hooks are contained in Ansible playbooks that can be run before or after the
```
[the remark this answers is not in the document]
mturley: @gildub oops, disregard this, I was looking at an old mockup.

vconzola:
@gildub Just make "hook" lowercase - "Create hook"

mturley:
@gildub since a lot of my stuff above is uncertain pending the discussion on Friday, if you just address the rest I'll merge it as-is and we can follow up to address those points.

src/app/Hooks/HooksPage.tsx
```diff
@@ -0,0 +1,87 @@
...
+    toggleAddEditModal();
+  };
+
+  return (
+    <>
+      <PageSection variant="light">
+        <Title headingLevel="h1">Hooks</Title>
+      </PageSection>
+      <PageSection>
+        <ResolvedQueries
```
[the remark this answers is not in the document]
gildub: Right. Since we're going to need to retrieve the plans too, I'll leave it for now.

src/app/Hooks/components/AddEditHookModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+import { useEditHookPrefillEffect } from './helpers';
+import LoadingEmptyState from '@app/common/components/LoadingEmptyState';
+
+interface IAddEditHookModalProps {
+  onClose: () => void;
+  hookBeingEdited: IHook | null;
+}
+
+const useHookFormState = (
+  clusterHooksQuery: QueryResult<IKubeList<IHook>>,
```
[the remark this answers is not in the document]
gildub: You're right, it's my mistake as I use existing component as template.

src/app/Hooks/components/AddEditHookModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+  return (
+    <Modal
+      className="AddEditHookModal"
+      variant="small"
+      title={`${!hookBeingEdited ? 'Create' : 'Edit'} hook`}
+      isOpen
+      onClose={onClose}
+      footer={
+        <Stack hasGutter>
+          <ResolvedQuery
```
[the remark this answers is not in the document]
gildub: I thought about that too. Would be interesting to see what we could factorize.

src/app/Hooks/components/HookActionsDropdown.tsx
```diff
@@ -0,0 +1,100 @@
...
+const HookActionsDropdown: React.FunctionComponent<IHookActionsDropdownProps> = ({
+  hook,
+  openEditHookModal,
+}: IHookActionsDropdownProps) => {
+  const [kebabIsOpen, setKebabIsOpen] = React.useState(false);
+  const [isDeleteModalOpen, toggleDeleteModal] = React.useReducer((isOpen) => !isOpen, false);
+  const [deleteHook, deleteHookResult] = useDeleteHookMutation(toggleDeleteModal);
+
+  // TODO: This is just a placeholder to be edited when backend for hooks is added
+  const hasRunningMigration = false;
```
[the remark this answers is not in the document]
gildub: Let's see what we come up with during our next Hook meeting.

src/app/Hooks/components/HookActionsDropdown.tsx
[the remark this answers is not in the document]
gildub: Not sure what to do yet. I leave it for now. Could be addressed in separate PR if needed too.

src/app/Hooks/components/HookActionsDropdown.tsx
gildub: Fair enough, waiting for confirmation too.

src/app/Hooks/components/HooksTable.tsx
gildub: Done

src/app/Hooks/components/HooksTable.tsx
gildub: Right, done!

src/app/queries/hooks.ts
gildub: Well, that's it, the more I help you, the further you're go. Let's continue... ;)

src/app/Hooks/components/helpers.ts
gildub: Thanks.

src/app/Hooks/components/HooksTable.tsx
gildub: `spacing.mtMd` fixes it!
I leave addressing it across the board to another PR.

gildub:
@mturley, thank for your review, I've addressed comments relevant prior the hook discussion.

mturley (approved):
LGTM! We can revisit the comments above next week as part of addressing https://github.com/konveyor/forklift-ui/issues/578.

PR #575, opened 2021-05-04
[RFR] Check provider status after adding OCP

nachandr:


qe-tests/cypress/integration/models/providerocpV.ts
```diff
@@ -32,7 +33,22 @@ export class ProviderocpV extends Provider {
...
+  protected populate(providerData: OcpVirtData): void {
+    Provider.openList();
+    const { name } = providerData;
+    cy.contains(name)
+      .parent(trTag)
+      .within(() => {
+        // Validating that provider is in `Ready` state
+        cy.get(dataLabel.status, { timeout: 600 * SEC }).should('have.text', 'Ready');
+        // Verifying that network field is not empty and count is not 0
+        cy.get(dataLabel.networks).invoke('text').then(parseFloat).should('be.gt', 0);
```
mturley, line 45: `parseFloat` is probably not necessary, you could use `parseInt`, but it doesn't really matter.

mturley (approved):
Looks good! We should probably eventually address the eslint errors shown in "Unchanged files with check annotations", but that's beyond the scope of this PR.

PR #585, opened 2021-05-06
Hooks wizard step scaffolding, form state, and modal to add hooks to the plan

mturley:
For the purposes of this wizard step, I'm using the term "instance" to refer to a hook that has been added to the plan and its associated options (which step to run on, any future fields in the modal).

This PR adds the Hooks step to the wizard, with an empty state and a modal to add a new hook instance. It is currently missing the table of hook instances (once an instance is added, JSON of the new form state is shown as a placeholder), and the modal doesn't yet support editing an instance (prefill logic).

There is some magic going on here to reduce duplication between this modal and the `AddEditHookModal` on the main Hooks page. I've factored out the form state for the "hook definition" fields (currently name, url and branch fields, but I know we're changing those) into a common `useHookDefinitionFields` hook. This returns an object of those `useFormField` calls, which is separately passed into the two different `useFormState` calls in `AddEditHookModal` and the new `PlanAddEditHookModal`, with the latter including additional fields.

I've also factored out the JSX for these field inputs into `HookDefinitionInputs` which is also used in both modals. That component is weird... it can hide the name field where necessary with a prop, and it can take children to render between the name field and other fields. We may need to change that structure a bit if we need more control over the order of the fields, or just give up on abstracting that. The `fields` prop of that component is fancy though, it can take any `fields` object from a `useFormState` call that includes the shared fields. That's why TS is happy with us passing the `form.fields` from the forms in both of these modals.

The `PlanAddEditHookModal` contains its own form state for the instance currently being added/edited. When the save button is clicked, the values from that form are passed to a handler in the `HooksForm` so they can be added to the actual plan wizard form state. This involves more magic: There is a type `PlanHookInstance` which is just the value types inferred from the `useFormState` in the `PlanAddEditHookModal`. That is the type used for the wizard's `forms.hooks.fields.instances` field, it's simply an array of those instances. So when we change the fields in these forms, the plan wizard's form state won't need to be changed, everything will match up. The `useFormState` call in `PlanAddEditHookModal` is the source of truth for these fields.

The Add functionality works! The values from the modal's form are just pushed into the plan's form state. Edit and remove will also be pretty similar to implement. You can even populate the form correctly by selecting an existing hook, using a pattern I borrowed from `MappingForm` (there is a possible opportunity to factor out a common component there).

I'm saving the Edit and Remove actions, and the table of hook instances, for another PR.

### DIFF
--- a/src/app/Mappings/components/MappingBuilder/MappingBuilder.tsx
+++ b/src/app/Mappings/components/MappingBuilder/MappingBuilder.tsx
@@ -20,6 +20,7 @@ export interface IMappingBuilderItem {
 export const mappingBuilderItemsSchema = yup
   .array<IMappingBuilderItem>()
   .required()
+  .min(1)
   .test('no-empty-selections', 'All sources must be mapped to a target.', (builderItems) =>
     builderItems ? builderItems.every((item) => item.source && item.target) : false
   );
--- a/src/app/common/components/TableEmptyState.tsx
+++ b/src/app/common/components/TableEmptyState.tsx
@@ -10,6 +10,7 @@ import {
   Button,
 } from '@patternfly/react-core';
 import { SearchIcon } from '@patternfly/react-icons';
+import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 
 export interface ITableEmptyStateProps {
   icon?: EmptyStateIconProps['icon'];
@@ -21,8 +22,8 @@ export interface ITableEmptyStateProps {
 
 const TableEmptyState: React.FunctionComponent<ITableEmptyStateProps> = ({
   icon = SearchIcon,
-  titleText = 'No results found',
-  bodyText = 'No results match the filter criteria. Remove filters or clear all filters to show results.',
+  titleText,
+  bodyText,
   onClearFiltersClick,
   isHiddenActions = false,
 }: ITableEmptyStateProps) => (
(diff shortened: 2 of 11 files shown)


mturley:
@vconzola I know the fields for the hook definition will change, but this is ready for an initial UX review for the Add modal. The one interesting thing to note is that when you select an existing hook, the only field that isn't prefilled is the "step when the hook will run", but it's not obvious right away. So I set it so that you'll get the validation message on that field ("Step is a required field") when selecting an existing hook, to call your attention to it. This won't happen if you select to create a new hook. See screenshots above, and the preview is working.

mturley:
@gildub I opened #586 and #587 to track the parts of this wizard step that I'm leaving out of this PR. If you want you could take on one of them.

src/app/Mappings/components/MappingBuilder/MappingBuilder.tsx
```diff
@@ -20,6 +20,7 @@ export interface IMappingBuilderItem {
 export const mappingBuilderItemsSchema = yup
   .array<IMappingBuilderItem>()
   .required()
+  .min(1)
```
gildub, line 23: Just a detail but this doesn't seem relevant to this PR.

src/app/common/components/TableEmptyState.tsx
```diff
@@ -10,6 +10,7 @@ import {
   Button,
 } from '@patternfly/react-core';
 import { SearchIcon } from '@patternfly/react-icons';
+import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
```
gildub, line 13: Same comment here, changes on this file do not seem relevant to this PR.

gildub (approved):
LGTM.

Just a couple of comments regarding changes seems out of the scope of this PR.
You don't have split to push those in a different PR. Up to you.

src/app/Mappings/components/MappingBuilder/MappingBuilder.tsx
mturley: After #566 (including a `yup` version upgrade) the handling of required arrays changed, and an empty array became a valid value for a required array field. This meant that without the `min(1)` here you could skip right past the mapping steps in the wizard without filling them in. I figured as part of adding a new wizard step I should make sure the wizard step flow is correct. I'll put it in another PR if you want though.

src/app/common/components/TableEmptyState.tsx
mturley: Oops! You're right. I was originally using TableEmptyState in HooksForm but decided it was easier to just directly use an EmptyState than make this component more complicated. Meant to revert this, I'll do so.

Issue #562, opened 2021-04-28
Hooks feature: API integration


mturley:
Part of #325. To be done after (or as part of) #561.

mturley:
See the TODO comments in `@app/queries/plans.ts` added by #595 for steps on implementing this.

PR #604, opened 2021-05-19
Plan creation with hooks

gildub:
Partially addresses https://github.com/konveyor/forklift-ui/issues/562


mturley:
@gildub to fully resolve #562 we also need to manipulate the hook CRs in `usePatchPlanMutation` according to those TODOs. Is that planned for another PR? If so can you edit the description here so it won't close #562, or open a follow-up issue?

mturley (approved):
LGTM for creating plans.

PR #612, opened 2021-05-23
[RFR] Automated test to cancel and restart a migration plan with multiple VMs

nachandr:


qe-tests/cypress/integration/models/plan.ts
```diff
@@ -145,6 +146,29 @@ export class Plan {
...
+      });
+  }
+
+  protected plan_details(name: string): void {
+    cy.get(tdTag).contains(name).parent(tdTag).parent(trTag).click();
+  }
+
+  protected cancel(name: string): void {
+    this.plan_details(name);
+    cy.get(`[aria-label="Select row 0"]`, { timeout: 20000 }).should('be.enabled').check();
```
mturley, line 167: FYI, this selector may break in the future if we improve the aria-label here to include the row's name (which I think we may end up doing).

qe-tests/cypress/integration/models/plan.ts
```diff
@@ -179,4 +203,27 @@ export class Plan {
...
     this.waitForSuccess(name);
   }
+
+  restart(name: string): void {
+    cy.get(tdTag)
+      .contains(name)
+      .parent(tdTag)
+      .parent(trTag)
+      .within(() => {
+        clickByText(button, 'Restart');
```
mturley, line 213: Ideally this traversal from the cell with the name to the row and down to the action cell shouldn't be necessary, if we have a good aria-label on that Restart button to indicate which plan it restarts. Making a note to possibly improve this in the future.

qe-tests/cypress/integration/models/plan.ts
```diff
@@ -179,4 +203,27 @@ export class Plan {
...
+  }
+
+  cancel_and_restart(planData: PlanData): void {
+    const { name } = planData;
+    Plan.openList();
+    this.run(name);
+    this.cancel(name);
+    this.waitForCanceled(name);
+    this.restart(name);
+    cy.wait(10000);
```
mturley, line 224: Why wait 10 seconds in particular? What exactly are you waiting for here?

mturley (approved):
@nachandr this looks good to me! I just have a few notes to myself for later and one question. I don't think it needs to block this from being merged though.

Issue #596, opened 2021-05-14
Show hooks in Review step of plan wizard, and View plan details modal


mturley:
Breaking this out of #595 because it would conflict with #589.

mturley:
This was closed in #617

PR #615, opened 2021-05-27
Hooks details

gildub:
- Adds proper hooks fields such as Definition (Ansible playbook or image) and Type (Pre or Post migration) 
- Hooks are optional so their fields should be displayed only when hooks exist for a plan
- Mappings and Hooks header fields must be displayed in bold

Screenshot from 2021-06-01 10-41-58

Screenshot from 2021-06-01 10-39-51

Resolves https://github.com/konveyor/forklift-ui/issues/596

### DIFF
--- a/src/app/Plans/components/PlanDetails.tsx
+++ b/src/app/Plans/components/PlanDetails.tsx
@@ -1,7 +1,8 @@
 import * as React from 'react';
-import { Grid, GridItem, Popover, Button, List, Text, ListItem } from '@patternfly/react-core';
+import { Grid, GridItem, Popover, Button, List, ListItem, Text } from '@patternfly/react-core';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import { StatusIcon } from '@konveyor/lib-ui';
+import text from '@patternfly/react-styles/css/utilities/Text/text';
 
 import MappingDetailView from '@app/Mappings/components/MappingDetailView';
 import { IMetaObjectMeta, IPlan, IVMwareVM, MappingType, POD_NETWORK } from '@app/queries/types';
@@ -9,6 +10,7 @@ import {
   useInventoryProvidersQuery,
   useMappingsQuery,
   useResourceQueriesForMapping,
+  useHooksQuery,
   useVMwareVMsQuery,
 } from '@app/queries';
 import { usePausedPollingEffect } from '@app/common/context';
@@ -47,6 +49,13 @@ const PlanDetails: React.FunctionComponent<IPlanDetailsProps> = ({ plan }: IPlan
   const warmCriticalConcernsFound = plan.spec.warm
     ? warmCriticalConcerns.filter((label) => someVMHasConcern(selectedVMs as IVMwareVM[], label))
     : [];
+
+  const hooks = useHooksQuery();
+  const getHookDefinition = (name) => {
+    const hook =
+      hooks.data?.items.find((hook) => (hook.metadata as IMetaObjectMeta).name === name) || null;
+    return hook ? !!hook.spec.playbook : false;
+  };
 
   const networkMappingResources = useResourceQueriesForMapping(MappingType.Network, networkMapping);
   const storageMappingResources = useResourceQueriesForMapping(MappingType.Storage, storageMapping);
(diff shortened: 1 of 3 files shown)


mturley:
You can mark this as resolving #596! I'll review asap. Thanks @gildub.

mturley:
@vconzola can you give it a visual review? I think maybe we should switch the order of the columns so "Migration step" comes first, like we did for the hooks table in the wizard.

src/app/Plans/components/PlanDetails.tsx
```diff
@@ -48,6 +50,13 @@ const PlanDetails: React.FunctionComponent<IPlanDetailsProps> = ({ plan }: IPlan
     ? warmCriticalConcerns.filter((label) => someVMHasConcern(selectedVMs as IVMwareVM[], label))
     : [];
 
+  const hooks = useHooksQuery();
+  const getHookDefinition = (name) => {
```
mturley, line 54: I'd either call this something like `isPlaybookHook` or have it return `'playbook' | 'image'` explicitly. `getHookDefinition` doesn't indicate what the boolean returned means. Just a nitpick though, it doesn't really matter.

If you do want to go with the string union, we could define it somewhere as `HookType` or something. It's hard-coded in `usePlanHookInstanceFormState` currently.

src/app/Plans/components/PlanDetails.tsx
```diff
@@ -192,24 +201,41 @@ const PlanDetails: React.FunctionComponent<IPlanDetailsProps> = ({ plan }: IPlan
...
-              <GridItem span={5} id="migration-plan-hooks-steps-label">
-                <label className="pf-c-form__label">Migration step</label>
-                {plan.spec.vms[0].hooks?.map((hook) => hook.step)}
+              <GridItem md={9} id="review-plan-hooks" aria-labelledby="migration-hooks-label">
+                <div>
+                  <Grid>
+                    <GridItem span={5} id="migration-plan-hooks-definition-label">
+                      <Text className={text.fontWeightBold}>Definition</Text>
+                      {plan.spec.vms[0].hooks?.map((hook) => (
+                        <Text key={Math.random()}>
```
mturley, line 217: We shouldn't use a random number as a key, the key should be unique to the array element. Definitely overkill here performance-wise, but it's a good habit anyway. I'd use `key={hook.step}` maybe, since we'll currently only ever have two.

src/app/Plans/components/PlanDetails.tsx
```diff
@@ -192,24 +201,41 @@ const PlanDetails: React.FunctionComponent<IPlanDetailsProps> = ({ plan }: IPlan
...
+                            ? 'Ansible playbook'
+                            : 'Custom container image'}
+                        </Text>
+                      ))}
+                    </GridItem>
+                    <GridItem span={2} className="migration-hooks-align" />
+                    <GridItem span={5} id="migration-plan-hooks-steps-label">
+                      <Text className={text.fontWeightBold}>Migration step</Text>
+                      {plan.spec.vms[0].hooks?.map((hook) => (
+                        <Text key={Math.random()}>
```
mturley, line 228: Same here.

src/app/Plans/components/Wizard/Review.tsx
```diff
@@ -140,6 +142,36 @@ const Review: React.FunctionComponent<IReviewProps> = ({
...
+            <GridItem md={3} id="migration-type-label">
+              Hooks
+            </GridItem>
+            <GridItem md={9} id="review-plan-hooks" aria-labelledby="migration-hooks-label">
+              <div>
+                <Grid>
+                  <GridItem span={5} id="migration-plan-hooks-definition-label">
+                    <Text className={text.fontWeightBold}>Definition</Text>
+                    {forms.hooks.values.instances?.map((hook) => (
+                      <Text key={Math.random()}>
```
mturley, line 156: Same with the random keys in this file as well.

mturley (commented):
Some minor nitpicks, take them or leave them. Also pending Vince's review.

src/app/Plans/components/PlanDetails.tsx
gildub: Good point it should be isPlayBookHook for a bolean returned type. 
Actually in the following PR#617 I changed it for string union, so I change it here.

gildub:
This PR is replaced with https://github.com/konveyor/forklift-ui/pull/617

PR #616, opened 2021-06-01
fix(cleanup): fix skip to content link page refresh issue

seanforyou23:
This PR is mostly just some small housekeeping items. The link in quickstart for the example json content was out of date, I updated and also made it a relative path so it works as expected on forks.

I'm unsure what [.env.example](https://github.com/konveyor/forklift-ui/blob/main/.env.example) is supposed to point to?

I also noticed when selecting the SkipToContent link the whole page refreshes, this should be fixed now.

Lastly, the spinner just needed an accessible name
Screen Shot 2021-06-01 at 12 59 35 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,16 +18,16 @@ cd forklift-ui
 npm install
 ```
 
-Create a meta.dev.json file in the config directory using [`config/meta.dev.example.json`](https://github.com/konveyor/forklift-ui/blob/main/config/meta.example.json) as a template. Set the `inventoryApi` property to the root URL of your forklift-controller inventory API, and set the `clusterApi` property to the root URL of your host OpenShift cluster API. And also to be able to use VMware provider data to be analysed by Migration Analytics set the `inventoryPayloadApi` property to the root URL of your forklift-controller inventory Payload API.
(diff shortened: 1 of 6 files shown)


mturley:
Ah, `.env` used to be how we pass the stuff that is now in `meta.json` (changed because it was simpler for the operator to use JSON), and for a while we still supported the dotenv plugin in case you wanted to persist those variables. Ended up removing it for some reason or another, so we should just remove references to the .env file in the README and instead describe that you can pass those environment variables via, well, the environment.

Also I see you're lacking maintainer rights, I'll fix that!

gildub (approved):
Welcome onboard!
LGTM.

mturley:
@seanforyou23 looks like this is failing CI because of eslint errors. I forgot to mention we're using Prettier and enforcing its formatting using `eslint-config-prettier` and `eslint-plugin-prettier`. You can auto-fix these by running `npm run format`. I also just have Format on Save configured in VS Code using the Prettier extension, but that's not for everyone. If you want it just for this project though, you can do that in workspace config.

mturley (approved):
LGTM other than the lint errors!

PR #617, opened 2021-06-01
Adds hook details and refactors shared elements to a common component

gildub:
- Adds proper hooks fields such as Definition (Ansible playbook or image) and Type (Pre or Post migration) 
- Hooks are optional so their fields should be displayed only when hooks exist for a plan
- Mappings and Hooks header fields must be displayed in bold
- Renames `PlanDetails` to `ConnectedPlanDetails` (open for better name suggestion)
- Moves <Grid> element from `Review` and `ConnectedPlanDetails` to new `PlanDetails`

Screenshot from 2021-06-02 14-18-14

Screenshot from 2021-06-02 15-34-40

Screenshot from 2021-06-02 15-36-42

### DIFF
--- a/src/app/Plans/components/PlanActionsDropdown.tsx
+++ b/src/app/Plans/components/PlanActionsDropdown.tsx
@@ -16,7 +16,7 @@ import { useClusterProvidersQuery, useDeletePlanMutation } from '@app/queries';
 import ConfirmModal from '@app/common/components/ConfirmModal';
 import ConditionalTooltip from '@app/common/components/ConditionalTooltip';
 import { areAssociatedProvidersReady } from '@app/queries/helpers';
-import PlanDetails from './PlanDetails';
+import ConnectedPlanDetails from './ConnectedPlansDetails';
 
 interface IPlansActionDropdownProps {
   plan: IPlan;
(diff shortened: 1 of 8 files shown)


src/app/Plans/components/PlanActionsDropdown.tsx
```diff
@@ -129,7 +129,7 @@ const PlansActionsDropdown: React.FunctionComponent<IPlansActionDropdownProps> =
           </Button>,
         ]}
       >
-        <PlanDetails plan={plan} />
+        <ConnectedPlanDetails plan={plan} />
```
mturley, line 132: I do think `ConnectedPlanDetails` needs a better name, it is reminiscent of Redux connected components which are a bit different. Maybe we could just put this `<Modal>` that is currently wrapped around it inside it, and call it `PlanDetailsModal`?

gildub: Definitely!
Thanks for the suggestion.

gildub:
@vconzola, as mentioned initially by @mturley in https://github.com/konveyor/forklift-ui/pull/615#issuecomment-852489256 :
Wouldn't it make sense to have "Migration step" column displayed first?

mturley (approved):
Approving and merging to unblock other PRs, @vconzola let's revisit this when you have time and follow up on any feedback in another PR.

vconzola:
@gildub Yep, @mturley  is right. Migration step should be first in the plan details views.

gildub:
@vconzola, noted, I'll move Migration step to be first and "None" (no headings, no bold) when there are no hooks.

PR #619, opened 2021-06-02
Hooks details: first column is hook' step; None value when no hooks

gildub:
Following up from https://github.com/konveyor/forklift-ui/pull/617#issuecomment-853309195
- Hook step is now in first column then hook definition
- None is displayed when there is no hook for plan
- Also renames React component IDs from `review` to `details`

Screenshot from 2021-06-02 22-14-47

Screenshot from 2021-06-02 22-16-23


mturley:
@gildub FYI these `review-*` ids were added for QE tests to use, it's fine to change them where it makes sense but we need to make sure we let @ibragins know when we do. I'll ping him.

mturley (approved):

gildub:
@mturley, yeah I was wondering about that, I checked they weren't involved in unit tests but thanks for following up with QE!

mturley:
Yeah I'm not seeing them in `qe-tests` yet either, but he asked me to add them for that reason so 🤷 
No problem 😄

vconzola:
Thanks @gildub. Looks good.

PR #620, opened 2021-06-03
PlanDetails: Refactor using IPlan object

gildub:
Using `generatePlan` and `generateMappings` to create a non resilient `IPlan` object to feed `PlandDetails` instead of fields.

### DIFF
--- a/src/app/Plans/components/Wizard/Review.tsx
+++ b/src/app/Plans/components/Wizard/Review.tsx
@@ -2,14 +2,15 @@ import * as React from 'react';
 import { TextContent, Text, Form } from '@patternfly/react-core';
 
 import { PlanWizardFormState } from './PlanWizard';
-import { IPlan, IVMwareVM, Mapping, POD_NETWORK } from '@app/queries/types';
+import { IMetaObjectMeta, IPlan, IVMwareVM, Mapping } from '@app/queries/types';
 import { MutationResult } from 'react-query';
 import { IKubeResponse, KubeClientError } from '@app/client/types';
 import { QuerySpinnerMode, ResolvedQueries } from '@app/common/components/ResolvedQuery';
-import { generateMappings } from './helpers';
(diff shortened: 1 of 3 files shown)


src/app/Plans/components/Wizard/Review.tsx
```diff
@@ -31,14 +32,27 @@ const Review: React.FunctionComponent<IReviewProps> = ({
 }: IReviewProps) => {
   usePausedPollingEffect();
 
+  // Create non resilient mappings and plan to display details before commit (or not)
   const { networkMapping, storageMapping } = generateMappings({ forms });
+  const plan: IPlan = generatePlan(
+    forms,
+    {
+      name: (networkMapping?.metadata as IMetaObjectMeta).name,
```
mturley, line 40: `generateMappings` won't produce a mapping with a `metadata.name` field if the wizard has no existing mapping being edited or no new mapping name being entered (instead it will use the `generateName` field). this `as IMetaObjectMeta` assertion here prevents TS from catching this. So this `name` field will be undefined in that case...

However, that doesn't really matter here because the `plan.spec.map` that is generated here isn't being used at all. So honestly we could just pass `{ name: '', namespace: '' }` for these mapping refs instead. If so we should probably leave a comment explaining that real mappings are not necessary for this ephemeral plan object.

I'm going to merge anyway so I can prevent merge conflicts with #597, but we should probably clean this up later.

mturley (approved):
LGTM, minor comment to address later.

src/app/Plans/components/Wizard/Review.tsx
gildub: @mturley, noted, I've added https://github.com/konveyor/forklift-ui/pull/627

mturley: Thanks!

PR #621, opened 2021-06-03
Pass detail=1 instead of detail=true for inventory API calls

mturley:
Based on discussion with @jortel , this will allow us to specify more variants for the level of detail in the future.


fdupont-redhat (approved):
LGTM

Issue #217, opened 2020-11-13
Text in a table cell should never wrap to multiple lines.


mturley:
By @vconzola's request. We need to look through all our tables and fix instances of this, basically every table cell should be able to truncate and have its contents show in a tooltip rather than wrapping. Not sure how this impacts cell widths and such.

mturley:
I created a `<TruncatedText>` component for this, not sure if it will work for all our cases.

mturley:
I think we can close this now, right @seanforyou23 ?

Issue #168, opened 2020-10-29
Visual bug: on medium-sized windows, some of our tables can stretch and overflow their container, making the last few columns unreachable


mturley:
Example, on the providers page:

![Screen Shot 2020-10-29 at 1 38 32 PM](https://user-images.githubusercontent.com/811963/97611177-33627980-19ec-11eb-8717-ddf4627875b0.png)

Some suggestions on how to handle it from UXD devs:

![Screen Shot 2020-10-29 at 1 40 54 PM](https://user-images.githubusercontent.com/811963/97611328-673d9f00-19ec-11eb-8163-1c883d24f88e.png)
![Screen Shot 2020-10-29 at 1 45 35 PM](https://user-images.githubusercontent.com/811963/97611823-0cf10e00-19ed-11eb-961b-c91e10baf1d3.png)

seanforyou23:
The attached PR helps this, but at the end of the day there are simply too many columns to fit all at once in horizontal fashion when the viewport reaches certain sizes. We could either hide some less important columns when there isn't enough room or set an earlier breakpoint as mentioned in the attached thread. I suppose we could also consider moving some of this column data into compound expandable cells. Are any of these columns less critical to be in view at all times?

mturley:
@vconzola see above. we should consider adjusting the columns of the VMware providers page in particular. Marking this issue as `needs-design` for followup.

vconzola:
Why does the Endpoint column not truncate like the other columns. Fixing that would be my first choice. Also, it looks like an earlier breakpoint auto-hides the navigation. That's a good option too. If we do both of those things can we avoid f-ing with any of the other columns? Btw, have I mentioned I hate PF tables? It ridiculous that a design system for enterprise apps can't handle a table with seven columns because of the asinine font size and spacing rules.

mturley:
Yeah, that's what the compact table variant is for which we could also try here, but it still has the font size issue. I like the look, but it definitely isn't as flexible as it should be

seanforyou23:
I updated that PR to set these tables into "compact" mode, and also found another table with needed the truncation treatment - definitely looking better IMO. I can look into adjusting the various breakpoints, but it can have cascading effects so maybe I'll do that in a dedicated PR to make reviewing easier. Let me know if you see any issues!

vconzola:
[...]

Issue #179, opened 2020-11-02
VM migration details page: center VM row text vertically


mturley:
To make this table a little cleaner, let's use some CSS to vertically center the text of the rows in `VMMigrationDetails` which are currently top-aligned (not sure what the best solution is).

![Screen Shot 2020-11-02 at 11 59 14 AM](https://user-images.githubusercontent.com/811963/97896549-3a96c980-1d03-11eb-9d5e-771aa8c3caf2.png)

mturley:
Removing the compact table piece of this, I assumed that's what was meant by the feedback for #181.

gildub:
@mturley, after I couldn't find reference documentation about how to use `columnTransforms` to be effective.
Then I tried some css rules by directly passing classes through IRows and ICells props.
This is incredibly time consuming and frustrating.

mturley:
@gildub don't worry about it then. It's not important enough to waste time on. I'll take a look later

PR #623, opened 2021-06-03
chore(styling): tidy styles for tables

seanforyou23:
This PR applies some styling enhancement to various tables in the UI. I tried to fine-tune them as best I could without making major changes to the way the data is output. It's not perfect in the sense that at certain viewport widths, the rows still slightly overflow their container, but it's to a lesser degree now.

### Before/after Migration details table:
Screen Shot 2021-06-03 at 1 23 36 PM

Screen Shot 2021-06-03 at 1 23 16 PM

### Before/after Providers table:
Screen Shot 2021-06-02 at 2 34 43 PM

Screen Shot 2021-06-09 at 11 50 18 AM

Steps toward resolving https://github.com/konveyor/forklift-ui/issues/168 and https://github.com/konveyor/forklift-ui/issues/217

### DIFF
--- a/src/app/Providers/components/ProvidersTable/OpenShift/OpenShiftProvidersTable.tsx
+++ b/src/app/Providers/components/ProvidersTable/OpenShift/OpenShiftProvidersTable.tsx
@@ -7,8 +7,10 @@ import {
   sortable,
   compoundExpand,
   classNames as classNamesTransform,
+  cellWidth,
   ICell,
   IRow,
+  truncate,
 } from '@patternfly/react-table';
 import { DatabaseIcon, NetworkIcon } from '@patternfly/react-icons';
 import tableStyles from '@patternfly/react-styles/css/components/Table/table';
@@ -19,9 +21,11 @@ import ProviderActionsDropdown from '../ProviderActionsDropdown';
 import StatusCondition from '@app/common/components/StatusCondition';
 import { MappingType } from '@app/queries/types';
 import { getMostSeriousCondition, hasCondition, numStr } from '@app/common/helpers';
+import { centerCellTransform } from '@app/utils/utils';
 
 import './OpenShiftProvidersTable.css';
-import { PlanStatusType } from '@app/common/constants';
+import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
+import { PlanStatusType, ProviderType } from '@app/common/constants';
 import { isSameResource } from '@app/queries/helpers';
 import OpenShiftNetworkList from './OpenShiftNetworkList';
 import SelectOpenShiftNetworkModal from '@app/common/components/SelectOpenShiftNetworkModal';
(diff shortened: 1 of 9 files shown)


gildub (changes requested):
Looks good.

Just a minor formating error which makes CI fails, a missing comma on line 11 of /home/runner/work/forklift-ui/forklift-ui/src/app/Providers/components/VMwareProviderHostsTable/VMwareProviderHostsTable.tsx

gildub:
@seanforyou23, thanks, could you please rebase to update branch.

src/app/Providers/components/ProvidersTable/OpenShift/OpenShiftProvidersTable.tsx
```diff
@@ -168,7 +200,7 @@ const OpenShiftProvidersTable: React.FunctionComponent<IOpenShiftProvidersTableP
 
   return (
     <>
-      <Level>
+      <Level className="pf-u-pb-sm">
```
mturley, line 203: Not that it matters much but I think we can import this class from `'@patternfly/react-styles/css/utilities/Spacing/spacing'` like we do elsewhere

mturley (changes requested):
Nice work man, it's looking better. I do think there's some odd stuff going on with vertical spacing though. In various tables we have cell contents that are positioned at the top or bottom of the cell instead of the center of the cell. At the very least they should all be consistent in each table (the openshift providers table has some columns misaligned with others) but ideally all the contents within the red lines here should have the same amount of space above and below them in their cell:

Screen Shot 2021-06-08 at 12 06 21 PM

Screen Shot 2021-06-08 at 12 08 06 PM

I'm realizing that part of this was already an issue (https://github.com/konveyor/forklift-ui/issues/179) but I think it's worth fixing while we're in here.

Also on the Plans page, the Progress column looks a bit odd with more space between the progress bar and its label than above and below the bar. I think that's one table that we can maybe leave in non-compact form. The migration details table and mappings table might also have been fine non-compact, it was mostly the providers tables that needed the extra space. What do you think @vconzola ?

mturley:
Thanks! Hm, @vconzola what do you think about the text in the rows of the migration details page? I was imagining we wanted the contents of the cells all to be vertically centered (so the column with the pipeline dots would have its text higher than the columns that don't have content under the text) but as it is currently, the text is all aligned and it just looks like all the other columns have blank space lined up with the pipeline dots.

vconzola:
@mturley You're correct. My intention was that cell contents be center aligned vertically even if it means the text above the pipeline isn't exactly in line with the text in other cells. Just a personal preference - I think it looks better this way.

seanforyou23:
There are a few different ways we could go about implementing the vertical alignment for the various data cells and I felt this PR start to grow - if you're cool with it I'd like to break that out into a follow-up PR to keep it manageable from a review standpoint.

mturley:
Looks like there are some merge conflicts now... sorry, I probably should have merged this ahead of #597 and fixed them on my end. Let me know if you want a hand.

seanforyou23:
All good - I rebased and went ahead and gave the Red Hat Virtualization provider table the same treatment.

mturley (approved):

Issue #112, opened 2020-10-08
Super-low-priority: Mappings pages should be unified into one page with tabs, consistent with providers


mturley:
Lowest priority in the whole backlog. Totally fine if we don't get to this

mturley:
Maybe obsolete? (do we want to entirely remove the mappings pages)

gildub:
@vconzola, what's the status on that one?

vconzola:
@gildub Let's just let this one sit for a while longer. Post-beta, I want to talk to Fabien and Marco about getting rid of Mappings altogether and just having users create mappings as part of plan creation, especially if there will be additional parameters (e.g., volume mode and access mode) that need to be specified for mapped storage classes.

PR #626, opened 2021-06-04
feat(Mappings): consolidate storage/network mappings pages

seanforyou23:
This PR updates the primary nav to go from two sub-menus for the different types of mappings to a single "Mappings" view with tabs for the mapping type options, modeled after the Providers page layout. There shouldn't be any functional changes, but please do test this out as I'm new to the project 😅 

Screen Shot 2021-06-04 at 10 19 30 AM

Should help close https://github.com/konveyor/forklift-ui/issues/112

[issue-112] marry network/storagem mapping pages

### DIFF
--- a/src/app/Mappings/Mappings.test.tsx
+++ b/src/app/Mappings/Mappings.test.tsx
@@ -1,0 +1,20 @@
+import * as React from 'react';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import Mappings from '@app/Mappings/Mappings';
+import { MappingType } from '@app/queries/types';
+
+describe('Mappings', () => {
+  const noop = jest.fn();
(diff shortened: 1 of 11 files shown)


vconzola (changes requested):
@seanforyou23 The only thing I see that looks a little hinky is that the page title changes with tab selection. I think ti should just say "Mappings" all the time. I do like that the context (storage vs. network) is included in the create modals.

gildub (commented):
This looks very good.
I've tested storage and mapping, creation, edition and deletion, it looks and works just fine.
Ready to merge once title issue mentioned by @vconzola is addressed.
Thanks

vconzola (commented):
LGTM. Thanks @seanforyou23 .

vconzola (approved):

gildub (approved):

PR #628, opened 2021-06-07
Update Plan Hooks: edit/add/delete

gildub:
Closes https://github.com/konveyor/forklift-ui/issues/562

### DIFF
--- a/src/app/queries/plans.ts
+++ b/src/app/queries/plans.ts
@@ -4,12 +4,14 @@ import { IKubeList, IKubeResponse, IKubeStatus, KubeClientError } from '@app/cli
 import { dnsLabelNameSchema, META } from '@app/common/constants';
 import { usePollingContext } from '@app/common/context';
 import { MutationResultPair, QueryResult, useQueryCache } from 'react-query';
+
 import {
   mockKubeList,
   nameAndNamespace,
   useKubeResultsSortedByName,
   useMockableMutation,
   useMockableQuery,
+  isSameResource,
 } from './helpers';
 import { MOCK_PLANS } from './mocks/plans.mock';
 import { IHook, IPlan, Mapping, MappingType } from './types';
@@ -18,6 +20,8 @@ import { getMappingResource } from './mappings';
 import { PlanWizardFormState } from '@app/Plans/components/Wizard/PlanWizard';
 import { generateHook, generateMappings, generatePlan } from '@app/Plans/components/Wizard/helpers';
 import { IMetaObjectMeta } from '@app/queries/types/common.types';
+import { useDeleteHookMutation } from '@app/queries';
+import { useHooksQuery } from './hooks';
 
 const planResource = new ForkliftResource(ForkliftResourceKind.Plan, META.namespace);
 const networkMapResource = getMappingResource(MappingType.Network).resource;
@@ -136,6 +140,9 @@ export const usePatchPlanMutation = (
   const client = useAuthorizedK8sClient();
   const queryCache = useQueryCache();
   const { pollFasterAfterMutation } = usePollingContext();
+  const hooks = useHooksQuery();
+  const [deleteHook] = useDeleteHookMutation();
+
   return useMockableMutation<IKubeResponse<IPlan>, KubeClientError, IPatchPlanArgs>(
     async ({ planBeingEdited, forms }) => {
       const { networkMapping, storageMapping } = generateMappings({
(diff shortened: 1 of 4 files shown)


src/app/queries/plans.ts
```diff
@@ -136,35 +140,111 @@ export const usePatchPlanMutation = (
...
+        );
+        return response;
+      });
+      await Promise.all(newHooksWithOwnerRef);
+
+      // Delete hooks removed from plan
+      const planHooksToDelete = planBeingEdited.spec.vms[0].hooks?.filter((vmsHook) => {
+        if (
+          forms.hooks.values.instances.length === 0 ||
+          forms.hooks.values.instances.filter((hook) => {
```
mturley, line 222: Have you been able to test this? Reading through the code here it doesn't seem like this would work correctly. If there is at least one hook instance that was not prefilled, then `!hook.PrefilledFromHook` will be true and short circuit the name comparison, and the length of this filtered array will not be zero regardless of the value of `vmsHook`. So if you've created a new hook and deleted an old one, the old one's CR won't actually be deleted.

I think the logic you meant to use here was `if (hook.prefilledFromHook && ...)` instead of `if (!hook.prefilledFromHook || ...)`. But, I think you can simplify this entire block to:

```ts
      const planHooksToDelete = planBeingEdited.spec.vms[0].hooks?.filter(
        (vmsHook) =>
          !forms.hooks.values.instances.find((instance) =>
            isSameResource(instance.prefilledFromHook?.metadata, vmsHook.hook)
          )
      );
```

src/app/queries/plans.ts
```diff
@@ -136,35 +140,111 @@ export const usePatchPlanMutation = (
...
+          }).length === 0
+        ) {
+          return true;
+        }
+        return false;
+      });
+      const hooksToDelete = hooks.data?.items.filter((hook) =>
+        planHooksToDelete?.find((vmHook) => isSameResource(vmHook.hook, hook.metadata) || null)
+      );
+      hooksToDelete?.forEach((hook) => deleteHook(hook));
```
mturley, line 239: I see the `deleteHook` function comes from your `useDeleteHookMutation` above, but that means it is asynchronous. We're not doing anything with the status object from it, and this plan mutation will resolve successfully even if the deletion fails. We want to make this deletion synchronously as part of the plan mutation, so that if there is any error it will bubble up and be shown in the `ResolvedQuery` component just like if something failed with the plan.

I think the solution here is to entirely get rid of `useDeleteHookMutation` and instead make the `client.delete(...)` calls right here and await them in a `Promise.all`, just like you did for patching the hooks above. And since you can do this using only the name/namespace ref, you don't even need the `const hooksToDelete`, you can just map directly over the `planHooksToDelete` from above.

mturley (changes requested):
Looking good, just a few minor things!

src/app/queries/plans.ts
gildub: For a given hook, 
if there are no hooks instances in form then the given hook is to be deleted (case all hooks are deleted from plan)
if there are hooks instances:
  then if the instance doesn't have prefilled data then it's a new hook
          or if it the instance has prefilled data and it's the same hook name then the hook is to be kept

I used that logic to create `planHooksToDelete`. 
The hooksToDelete is to map hooks (to delete) from the VM hooks (to delete) since they're different structures.

It works fine, I effectively tested it.

That said, your approach is neat (and makes me realize I'm still progressing :) ).
I'll use your then.
Thanks

src/app/queries/plans.ts
gildub: Yes that makes total sense,
I think I wrongly went on `useDeleteHook` road because of the way delete happens in other parts.
That's overkill as we just need to be synchronous
So now using `client.delete()` directly.

src/app/queries/plans.ts
mturley: Maybe it's just an issue of code readability and I missed something then :) no problem.

I think it did work fine unless you had both created a new hook instance AND deleted a hook instance in the same plan edit operation. In that case, the `forms.hooks.values.instances.filter(...).length === 0` would never be true, because there is a new instance that causes `if (!hook.prefilledFromHook)` to be true and short circuit the `name === name` comparison in every iteration of the filter loop. So it doesn't matter which `vmsHook` we're filtering on, the logic says that none of them should be deleted as long as any non-prefilled instance is present. I may be misreading it though. Anyway, simplifying sounds good, thanks for taking my suggestion :)

mturley (approved):
Looks great! Thanks @gildub !

Issue #218, opened 2020-11-13
Select VMs step of wizard: in smaller screen sizes, the filter dropdown has some strange behavior


mturley:
It has a weird transparent background:

![Screen Shot 2020-11-13 at 3 46 21 PM](https://user-images.githubusercontent.com/811963/99119760-baf0e080-25c7-11eb-9957-1e8e0133537b.png)

Also, if you click the X in the text field, instead of just clearing the text field it also shows a filter chip that was otherwise hidden (should appear immediately)

gildub:
This occurs also with PlansPage filter.

![filter-resize-plans](https://user-images.githubusercontent.com/1901741/100335713-93d9dc00-2fd5-11eb-9371-d30b5f1d1e61.png)

The filter seems to have several issues in smaller screen size:
- Looks like issues with the breakdown points definitions
- Transparent background is confusing as it mixes with the page data

mturley:
Super low priority since I don't think we will claim to support mobile devices for MTV. But post-beta, we should see if this is also an issue in upstream PF.

PR #630, opened 2021-06-07
fix(toolbar): improve toolbar items layout

seanforyou23:
This PR swaps Level layout for Split in the filter toolbar area for the toolbars in PlansTable, VMMigrationDetails, and SelectVMsForm.

Best way to test is check at various sceen widths the migration plans view, a migration detail view, and the select VM step in plan creation flow, they all received a same treatment.

I also fixed a small issue where an Alert wasn't being announced to screen reader users.

Should help close https://github.com/konveyor/forklift-ui/issues/218

Screen Shot 2021-06-09 at 3 25 04 PM

Screen Shot 2021-06-09 at 3 25 49 PM

### DIFF
--- a/src/app/Plans/components/CreatePlanButton.tsx
+++ b/src/app/Plans/components/CreatePlanButton.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import { Button, ButtonProps } from '@patternfly/react-core';
-import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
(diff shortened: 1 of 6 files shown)


mturley:
I think @vconzola had a strong opinion that the create button should be left-aligned there with the filters, maybe we do need that grid.

vconzola:
@mturley that's correct - button should be on the left, next to the filters.

mturley (approved):
LGTM! Thanks Mike

PR #632, opened 2021-06-08
chore(mappings): allow for specifying mapping type during mapping creation flow

seanforyou23:
This PR adds a dropdown to the mapping creation flow that's pre-selected to whatever resource type is selected in the tabs.

Screen Shot 2021-06-08 at 11 14 42 AM

### DIFF
--- a/src/app/Mappings/MappingsPage.tsx
+++ b/src/app/Mappings/MappingsPage.tsx
@@ -8,18 +8,13 @@ import {
   Tab,
   TabTitleText,
 } from '@patternfly/react-core';
-import { Mapping, MappingType } from '@app/queries/types';
+import { Mapping, MappingType, MapType } from '@app/queries/types';
 import CreateMappingButton from '@app/Mappings/components/CreateMappingButton';
 import AddEditMappingModal from '@app/Mappings/components/AddEditMappingModal';
 import Mappings from '@app/Mappings/Mappings';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 
 const MappingsPage: React.FunctionComponent = () => {
-  enum MapType {
-    'Network',
-    'Storage',
-  }
-
   const [activeTabKey, setActiveTabKey] = React.useState<React.ReactText>(0);
   const [activeMapType, setActiveMapType] = React.useState('Network');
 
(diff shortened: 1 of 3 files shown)


src/app/Mappings/MappingsPage.tsx
```diff
@@ -8,18 +8,13 @@ import {
   Tab,
   TabTitleText,
 } from '@patternfly/react-core';
-import { Mapping, MappingType } from '@app/queries/types';
+import { Mapping, MappingType, MapType } from '@app/queries/types';
```
mturley, line 11: I just noticed that you had introduced MapType in the last PR. I don't think it's necessary, right? `MappingType` is the same enum. I've been considering switching it to a string union (`'Network' | 'Storage'`) anyway which would be even simpler. Same for a lot of the other enums.

src/app/Mappings/MappingsPage.tsx
```diff
@@ -8,18 +8,13 @@ import {
...
 import Mappings from '@app/Mappings/Mappings';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 
 const MappingsPage: React.FunctionComponent = () => {
-  enum MapType {
-    'Network',
-    'Storage',
-  }
-
   const [activeTabKey, setActiveTabKey] = React.useState<React.ReactText>(0);
```
mturley, line 18: The tabKey can be a string, and we don't need both activeTabKey and activeMapType since they duplicate the same state. See the tabs on ProvidersPage, it's all driven from `activeProviderType` which is just a string.

mturley (changes requested):
Looking good, but there is a bit of redundancy in the enum and the state we can get rid of.

Also @vconzola would you mind reviewing it for visuals? How do you feel about the type and name being side by side in the grid? I think the only reason we had source and target provider selections side by side is because that way they line up with the mapping builder columns that appear below.

src/app/Mappings/MappingsPage.tsx
seanforyou23: Ah yes, tabKey as string, good call! Updated.

src/app/Mappings/MappingsPage.tsx
seanforyou23: Right, I was following suit from when tabKey was a number. I simplified it to use a string and removed the redundant enum. Good call.

mturley (approved):
LGTM! I'll just wait for Vince's review to merge.

vconzola (approved):
@mturley Even though it's not "standard", I think it looks OK with Type and Name side-by-side. It balances the rest of the form content. If we hear complaints that people are missing the Name field we can change it later. But I doubt that will happen.

PR #636, opened 2021-06-10
selectvms table adjustments

seanforyou23:
This PR updates the select vm table to truncate column headers for better space mgmt. Further addresses https://github.com/konveyor/forklift-ui/issues/217. 

Screen Shot 2021-06-09 at 10 20 07 PM

Also, I pulled some recently added styles for provider hosts table back out, they seemed necessary before but now was causing issues. I tested in FF and chrome, seems to be find but could use a second set of 👀


mturley (approved):
LGTM 👍

Issue #641, opened 2021-06-10
No padding above "create mapping" button in empty state of mappings pages


mturley:
@seanforyou23 I missed this when reviewing #626. That's one flaw in our mock mode, you can't get to the empty states to see if they look right.

The button spacing here should match the empty states on the other pages, with some padding between the button and the empty state body text. This is present on the empty state of both the Network and the Storage tab of the new mappings page.

![Screen Shot 2021-06-10 at 3 02 10 PM](https://user-images.githubusercontent.com/811963/121582501-17e93380-c9fd-11eb-83e6-bc847a406afd.png)

PR #643, opened 2021-06-11
fix(mappings): remove unnecessary div around create mappings button

seanforyou23:
This PR fixes a spacing issue for the empty state of the mappings page. By removing the unnecessary div, PatternFly styles seem to do the expected thing without any modification. Should help close https://github.com/konveyor/forklift-ui/issues/641

Screen Shot 2021-06-11 at 8 28 23 AM


vconzola (approved):
LGTM.

gildub (approved):
LGTM

mturley:
Are you sure this doesn't break the tooltip that can appear when the button is disabled because there are insufficient providers? (Can't test now, just posting before I forget)

PR #653, opened 2021-06-15
Filter VMs tree view: exclude hosts, be explicit about which kinds of nodes are selectable, properly exclude deselected child folders

mturley:
Based on discussions with @jortel.

This PR solves two issues with the Filter VMs step of the wizard.

1. The structure of the RHV tree has VMs as peers of hosts rather than children, and the VMware cluster tree will be changing to share this flattened structure. This means that selecting a host would never result in any VMs showing up on the next step, because host tree nodes will never have VM children. We decided it makes sense to just remove Hosts from the UI tree entirely, since they are not very meaningful for the user anyway.
2. There is currently a bug in the cluster/host tree where selecting a datacenter and then deselecting one or all of the child clusters results in the datacenter still being selected. This doesn't make sense because datacenters will never have VM children in that tree. However, in the folder tree, we need to retain this behavior because there can be VM children of a parent folder that are not in a child folder, so we want to be able to exclude child folders and retain the VMs in the parent.

To solve these with as little hard-coded logic about the tree node kinds as possible, I introduced two new helpers: `isIncludedNode` and `isNodeSelectable`. `isIncludedNode` is a constant, which just excludes hosts from the tree (preventing them from appearing in the tree and from being considered leaf nodes for selection logic purposes). `isNodeSelectable` can be passed into these tree helpers, so that it can differ by tree type or whatever other conditions we need. The one defined by `useIsNodeSelectableCallback` (which is only defined as a hook so it can use React.useCallback to prevent triggering the useEffect that performs all this logic on every render) only allows leaf nodes (clusters) to be selected in the cluster tree, and allows any folders or datacenters to be selected in the folder tree. The `isFullyChecked` and `isPartiallyChecked` logic has been abstracted out to use these so that the non-selectable nodes correctly render as checked/indeterminate/unchecked based on the checked state of their selectable descendants.

You can verify this behavior by checking the box on a datacenter, then unchecking one of its cluster children and observing that the datacenter checkbox becomes indeterminate rather than staying selected, and observing that this *doesn't* happen with folders.

I also cleaned up the tree mock data a bit, there were duplicate VMs in the folder tree that made it difficult to verify this excluded-child fix.

I noticed a few more opportunities to simplify some of this tree code in future PRs, but I didn't want this PR to get any larger so I will follow up later. In particular, I left a TODO comment about optimizing `getVMTreePathInfoByVM` which could be performed once for the entire tree at query time rather than on every render. This optimization would remove a lot of the need for caching / useEffect in SelectVMsStep as well.

### DIFF
--- a/src/app/Plans/components/Wizard/FilterVMsForm.tsx
+++ b/src/app/Plans/components/Wizard/FilterVMsForm.tsx
@@ -14,9 +14,10 @@ import {
   findMatchingNode,
   findMatchingNodeAndDescendants,
   findNodesMatchingSelectedVMs,
-  flattenVMwareTreeNodes,
+  getSelectableNodes,
   getSelectedVMsFromPlan,
   isNodeFullyChecked,
+  useIsNodeSelectableCallback,
 } from './helpers';
 import { PlanWizardFormState } from './PlanWizard';
 
@@ -31,13 +32,8 @@ interface IFilterVMsFormProps {
 }
 
 // TODO switch to new cluster tree, update mock data, test against Jeff's cluster
-// TODO hide hosts entirely
 // TODO figure out if it makes sense to change Select VMs list so it comes from filtering on VM properties matching selected tree nodes
 //      instead of using the leaf nodes. otherwise, see if it makes sense to limit to only the direct child leaves.
-// TODO figure out if we need to change how selection and deselection works in terms of auto-toggling descendants, and having indeterminates.
-//      maybe the only state we save is which (displayed) leaves are selected, and the rest is render logic.
-//      (i.e. there is no such thing as selecting a parent, a parent is only shown as selected if all its children are).
-// See notes in Typora scratch
 
 // TODO also for the folder path column, make a hash of folder ids to tree nodes, so we can walk up their parents to build the path string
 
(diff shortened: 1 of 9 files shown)


vconzola (commented):
@mturley There's still an aspect of this that feels weird to me on the folders tab. I don't think it's intuitive that you can have an indeterminate parent node whose children are all selected or a selected parent node whose children are all deselected. I understand why it's this way, but am not sure users will. How difficult would it be to add labels to each node in the tree showing how many VMs the current selection encompasses? So when you first select a node it will show the sum of its own VMs and all of its descendents' VMs. And then as children are selected or deselected the sums will be updated at each level.

mturley:
@vconzola I think it won't be too hard to add those numbers, the tree view supports badges like that. The question will be exactly how to calculate those.. it seems odd to me that the numbers would change depending on what is selected, maybe it's sufficient to show how many VMs are direct children of each node? There wouldn't be any number on the datacenter or on folders that have all their VMs in subfolders.. hm. That also sounds confusing now that I type it out.

So you're suggesting that the numbers only indicate how many VMs are being included from that subtree, not how many are available in that part of the tree? That makes sense I think... I'll have to figure out a good way to perform that math without making the code even more messy :)

Let's talk on Monday so I can confirm exactly what you're asking for before I go down the wrong path. I do think it should probably be a separate PR, too.

gildub (approved):
Code looks good.

I'm a bit parted, not because of this PR directly but more the direction we're going. 
Maybe I'm wrong but I wonder what's better between component being generic and adapting to various trees vs having trees handling a certain type of tree. Brainstorm!

mturley:
@gildub yeah, there's certainly some discussion to be had there. I have a followup PR or two in mind to clean up what we have and make it more readable at least, so it would be easier to talk about. These helper functions are generic enough that they should work with any tree-structured JSON that has these `kind` properties and (currently) `kind: 'VM'` leaf nodes to filter by. I tried to keep the logic specific to VMware and RHV as decoupled from the basic tree traversal stuff as possible.

BUT, I was talking to @jortel and we think that for 2.2 and beyond we can provide properties on the VM inventory objects that make it possible to filter by selected clusters/folders/etc and list with the columns we need, without using the `kind: 'VM'` leaf nodes in the tree at all (removing them would improve backend performance). That would eliminate the need for many of these helper functions.

So I think maybe it makes sense to continue down the path of maintaining and simplifying what we have, and by the time we get it into that state maybe it'll be obvious whether it's worth keeping or replacing. In theory it's not that damn complicated, so I apologize for creating some spaghetti around it :)

Issue #638, opened 2021-06-10
Deleting the last provider of a given type leaves you on a broken route


mturley:
I deleted the only RHV provider in a cluster, and I stayed on the /providers/ovirt page but the RHV tab disappeared and the table rendered with no empty state. It should redirect you to `/providers` so the router can take you to either the default tab or the empty state.

PR #654, opened 2021-06-16
Go to default tab when deleting a last provider

gildub:
Fixes https://github.com/konveyor/forklift-ui/issues/638


mturley (approved):
Thanks!

Issue #567, opened 2021-04-29
Upgrade to React Query v3


mturley:
This will involve handling some significant breaking changes, but I think we can leverage some of the new cache features and API changes to make our query code even more readable and maintainable.

https://react-query.tanstack.com/guides/migrating-to-react-query-3

PR #657, opened 2021-06-18
React query v3 upgrade

seanforyou23:
This PR upgrades `react-query` to v3. There really isn't a way to introduce this in pieces as far as I can tell, sorry for the big diff 😅 It seems like a massive change, and while it isn't small most of this is the same change in a bunch of places throughout the codebase. Should help close https://github.com/konveyor/forklift-ui/issues/567

@mturley major thx for helping me work through this!

### Highlights

- react query devtools is now part of the react-query pkg
- Rendering components inside the test suite now requires wrapping JSX in a stub `QueryClientProvider`
- QueryStatus enum was removed, just use string literals for states now (success, error, loading etc)
- `QueryResult` => `UseQueryResult`
- `MutationResult` => `UseMutationResult`
- `ReactQueryCacheProvider` => `QueryClientProvider`
- Can now utilize [`select`](https://react-query.tanstack.com/guides/migrating-to-react-query-3#query-data-selectors) config property to apply transforms to returned data, eliminating the need for complex sort helpers
- I started to pair back some of the function return type definitions where we used generics heavily. It was [recommended](https://tkdodo.eu/blog/react-query-and-type-script#type-inference) by maintainers of react-query since ts inferencing does the same job with less code

### DIFF
--- a/src/app/Plans/components/Wizard/PlanWizard.tsx
+++ b/src/app/Plans/components/Wizard/PlanWizard.tsx
@@ -12,7 +12,7 @@ import {
   WizardStepFunctionType,
 } from '@patternfly/react-core';
 import { Link, Prompt, Redirect, useHistory, useRouteMatch } from 'react-router-dom';
-import { QueryResult, QueryStatus } from 'react-query';
+import { UseQueryResult } from 'react-query';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import { useFormField, useFormState } from '@konveyor/lib-ui';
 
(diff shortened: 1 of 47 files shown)


src/app/Plans/components/Wizard/PlanWizard.tsx
```diff
@@ -230,15 +227,15 @@ const PlanWizard: React.FunctionComponent = () => {
 
   const onSave = () => {
     if (!planBeingEdited) {
-      createPlan(forms);
+      createPlanMutation.mutate(forms);
```
seanforyou23, line 230: 👍

mturley (approved):
Nice work man!

PR #660, opened 2021-06-22
fix(table): enter grid layout when there is not enough horizontal space

seanforyou23:
This PR sends the Red Hat Virtualization and VMware tables into grid layout up through large viewports (992px) to avoid table data from overflowing its container. The OpenShift Virtualization table remains as it was. Should help close https://github.com/konveyor/forklift-ui/issues/168

Screen Shot 2021-06-22 at 10 27 59 AM


mturley (approved):
LGTM!

PR #661, opened 2021-06-22
Filter VMs step: show badge with number of VMs provided by each tree node's selected descendants

### LINKED ISSUE #658: Filter VMs step: add number badges to the tree items indicating how many available VMs are coming from those selections
See @vconzola 's comment here: https://github.com/konveyor/forklift-ui/pull/653#pullrequestreview-686754691

> @mturley There's still an aspect of this that feels weird to me on the folders tab. I don't think it's intuitive that you can have an indeterminate parent node whose children are all selected or a selected parent node whose children are all deselected. I understand why it's this way, but am not sure users will. How difficult would it be to add labels to each node in the tree showing how many VMs the current selection encompasses? So when you first select a node it will show the sum of its own VMs and all of its descendents' VMs. And then as children are selected or deselected the sums will be updated at each level. 

I don't want to block #653 on this, but we should come back to it.

mturley:
Resolves #658.

@vconzola I'm curious about your feedback here. With the mock data we have it's easiest to test this with the folder view. Note that the V2V-DC datacenter has 2 direct VM children that are not in any folders, and all other VMs are under a leaf node folder.

Based on [your comment](https://github.com/konveyor/forklift-ui/pull/653#pullrequestreview-686754691) I think this is what you were asking for? Any tree node that is selected or has selected descendants (showing checked or indeterminate) will have a number next to it with the total number of VMs coming from the selections under that node. If a node isn't selected it won't have a badge (so we don't have "0"s all over the tree), but if it is selected and provides no VMs it gets a "0" badge to be extra clear. (not sure how often this will come up in real data anyway, we just have empty stuff in our mock data).

It feels a little weird, but maybe is fine? I wonder if we should add "VMs" as a suffix on the number to be more clear what those numbers mean?

Also, I was initially messing around with a slightly different approach where instead of showing anything about descendants, each node just lists the number of VMs that will be included by selecting that node (and not its children). So the numbers will not change based on selections, they just describe where the VMs are in the tree. You can see that older preview here: https://forklift-tree-counts-v1-preview.surge.sh/ (edit: tore this down)

For both of these previews, the easiest way to get to the tree quickly is by editing plantest-02 and clicking Filter VMs in the wizard nav. As mentioned above it's more apparent in the Folders tab.

### DIFF
--- a/src/app/Plans/components/Wizard/FilterVMsForm.tsx
+++ b/src/app/Plans/components/Wizard/FilterVMsForm.tsx
@@ -91,8 +91,9 @@ const FilterVMsForm: React.FunctionComponent<IFilterVMsFormProps> = ({
     const { isItemSelected, selectedItems } = treeSelection;
     const selectedDescendants = flattenInventoryTreeNodes(node).filter(isItemSelected);
     const numVMs = getAvailableVMs(selectedDescendants, vmsQuery.data || [], treeType).length;
+    const rootNodeSuffix = ` VM${numVMs !== 1 ? 's' : ''}`;
     if (numVMs || isItemSelected(node) || (isRootNode && selectedItems.length > 0)) {
-      return `${numVMs}${isRootNode ? ` VM${numVMs !== 1 ? 's' : ''}` : ''}`;
+      return `${numVMs}${isRootNode ? rootNodeSuffix : ''}`;
     }
     return null;
   };
--- a/src/app/Plans/components/Wizard/FilterVMsForm.tsx
+++ b/src/app/Plans/components/Wizard/FilterVMsForm.tsx
@@ -14,6 +14,7 @@ import {
   findMatchingNode,
   findMatchingNodeAndDescendants,
   findNodesMatchingSelectedVMs,
+  flattenInventoryTreeNodes,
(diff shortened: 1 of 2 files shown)


mturley:
The weirdness I still see here is that the only way to tell there are VMs directly under a datacenter/folder that aren't in one of the child folders is to notice that the number on that node is greater than the sum of the numbers on its children. And if a folder contains no direct VMs and its number comes entirely from its descendants, that's also not obvious right away. But maybe that's still more clear than the alternative approach I linked above.

mturley:
UX approved by Vince privately.

seanforyou23 (commented):
Looks good to me, do we want to address the SonarCloud issue?

mturley:
So far I've been mostly ignoring SonarCloud except for security warnings, until we have time to clean up the full codebase to make it happy. But I'll take a look at these ones

seanforyou23 (approved):

gildub (approved):
LGTM

Need @vconzola review.

mturley:
@gildub I got a verbal approval from Vince while discussing this. Rebased, will merge when CI is green.

PR #662, opened 2021-06-22
fix(plans): conditionally display select network link in plans wiz

seanforyou23:
This PR hides the "Select a different network" link if the specified namespace isn't found in the available namespaces of the first (general) step of the plans wizard. It also ensures values persist when removing focus from the target namespace field. Should help close https://github.com/konveyor/forklift-ui/issues/463

Screen Shot 2021-06-22 at 2 39 32 PM

Screen Shot 2021-06-22 at 2 39 48 PM

### DIFF
--- a/src/app/Plans/components/Wizard/GeneralForm.tsx
+++ b/src/app/Plans/components/Wizard/GeneralForm.tsx
@@ -45,6 +45,11 @@ const GeneralForm: React.FunctionComponent<IGeneralFormProps> = ({
   const clusterProvidersQuery = useClusterProvidersQuery();
   const namespacesQuery = useNamespacesQuery(form.values.targetProvider);
 
+  const targetNsFoundInQueriesNs = () =>
+    namespacesQuery.data?.reduce((acc, namespace) => {
+      return acc ? true : form.values.targetNamespace === namespace.name;
+    }, false);
+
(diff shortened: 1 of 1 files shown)


src/app/Plans/components/Wizard/GeneralForm.tsx
```diff
@@ -45,6 +45,11 @@ const GeneralForm: React.FunctionComponent<IGeneralFormProps> = ({
   const clusterProvidersQuery = useClusterProvidersQuery();
   const namespacesQuery = useNamespacesQuery(form.values.targetProvider);
 
+  const targetNsFoundInQueriesNs = () =>
```
mturley, line 48: This may be nitpicky, but I don't think this needs to be a function. It's called exactly once within the same scope, it could just be a boolean variable. I guess we already do that with `getFilteredOptions` so if this feels cleaner that's fine.

Also I think it would be simpler to use `.find()` instead of `.reduce()`, that way you don't need the ternary and it will stop looping as soon as it finds a match.

```ts
const targetNsExists = !!namespacesQuery.data?.find(
  (namespace) => form.values.targetNamespace === namespace.name
);
```

mturley (approved):
LGTM 👍

Issue #36, opened 2020-09-09
Add "expand/collapse all" control in the header of the leftmost column of the select VMs table in step 2 of the plan wizard.


mturley:
Similar to #35, we will need a change in PF to support this since we can't modify the header of that column (it's generated by the `expandable` transform in `Table`). Will open an issue there and follow up here.

Once it's available, [as noted here](https://github.com/konveyor/virt-ui/pull/21#issuecomment-689722506) we can use the `selectAll` and `areAllSelected` values from `useSelectionState` to drive this button, and it should use the `<AngleRightIcon>` and `<AngleDownIcon>` icons depending on its state.

mturley:
Upstream issue to track: https://github.com/patternfly/patternfly-react/issues/4798

mturley:
The upstream issue was closed because it is moving back to the design stage. It sounds like they are talking about making it a toolbar button above the table though, instead of in the table header. We could probably implement it like that today, if you want to @vconzola .

vconzola:
@mturley It's up to you. Definitely a nice-to-have for beta if you just want to wait until the design is finalized in PF.

mturley:
sure. I'll leave this issue open until I see it finalized in PF, we can wait.

mturley:
This was concluded in PF back in October, the decision was to support it as a toggle in the toolbar above the table. See this demo: https://github.com/patternfly/patternfly-react/pull/5030

PR #663, opened 2021-06-23
feat(table): allow for expanding/collapsing all rows in VM selection

seanforyou23:
This PR adds an "expand/collapse" all rows toggle to the select VMs table. It also moves the select all rows function to a bulk select component and refactors the FilterToolbar layout. Should help close https://github.com/konveyor/forklift-ui/issues/36

Screen Shot 2021-06-24 at 4 22 54 PM

### DIFF
--- a/src/app/Plans/components/Wizard/SelectVMsForm.tsx
+++ b/src/app/Plans/components/Wizard/SelectVMsForm.tsx
@@ -5,10 +5,14 @@ import {
   Text,
   Level,
   LevelItem,
-  Split,
-  SplitItem,
+  Button,
+  Dropdown,
+  DropdownToggle,
+  DropdownToggleCheckbox,
+  DropdownItem,
+  ToolbarItem,
 } from '@patternfly/react-core';
-import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
+
 import {
   Table,
   TableHeader,
@@ -20,6 +24,10 @@ import {
   wrappable,
   truncate,
 } from '@patternfly/react-table';
+
+import { AngleDownIcon, AngleRightIcon } from '@patternfly/react-icons';
+import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
+import '@app/Plans/components/Wizard/SelectVMsForm.css';
 
 import {
   SourceVM,
@@ -184,7 +192,7 @@ const SelectVMsForm: React.FunctionComponent<ISelectVMsFormProps> = ({
             title: 'Folder path',
             type: FilterType.search,
             placeholderText: 'Filter by folder path ...',
-            getItemValue: (item) => {
+            getItemValue: (item: SourceVM) => {
               const { folderPathStr } = getVMTreeInfo(item);
               return folderPathStr ? folderPathStr : '';
             },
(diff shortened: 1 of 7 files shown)


mturley:
@vconzola do you mind giving it a visual review? does the spacing in that toolbar look right?

vconzola (dismissed):
@seanforyou23 @mturley Why are we not using the Bulk selection control (https://www.patternfly.org/v4/guidelines/bulk-selection/) for the VMs table? If we used that along with the Expand/collapse all control (https://www.patternfly.org/v4/components/table/react-demos/#expandcollapse-all) in the toolbar, I think spacing would take care of itself.

mturley:
Hmm I wasn't aware of that control, it seems weird that that exists but there is also a `canSelectAll` prop that we're using to put that checkbox in the header

seanforyou23:
It's a good point @vconzola, I'm not sure why bulk select wasn't used historically but can definitely look into using it here. To @mturley's point, we probably don't want both of these mechanisms to select/deselect all so probably just drop the one from the table header column in favor of the one in the toolbar, right?

I notice the pagination is actually also not housed within the Toolbar itself, so I could take this opportunity to refactor that area of the code so that the bulk expand, bulk select, filter controls, and pagination are all inside the toolbar. This should allow us to drop the Split component usage here and make better use of that `toolbarItems` prop that was recently added for FilterToolbar.

My question is, can we accept this incrementally? If the functionality is ok I'd love to get this bit merged, and then send a follow-up PR that implements bulk select and refactors the FilterToolbar - but no problem if that's not desirable, I can do it all as part of this PR if that's preferred.

vconzola:
@seanforyou23 I'm fine with doing it incrementally, and what you have is OK. But I'd request one change - can you vertically align the filter control with the left side of the checkboxes? It would look a little cleaner. It would also look better if the expand all was in the same row as the select all? I'm a little concerned it's too far away from the table contents to be noticed. But again, this can be an incremental change too.

mturley:
I agree with all of this, we can merge incrementally to keep PRs small but have the end goal of using the Toolbar properly and cleanly for all our tables.

src/app/Plans/components/Wizard/SelectVMsForm.tsx
```diff
@@ -294,6 +308,11 @@ const SelectVMsForm: React.FunctionComponent<ISelectVMsFormProps> = ({
     }
   });
 
+  const toggleCollapseAll = (collapse: boolean) => {
```
mturley, line 311: @seanforyou23 there is some out of sync state here since the toggle state is separate from the expanded/collapsed state of the rows. For example, if I click the toggle to expand all the rows, but then I collapse an individual row or all the rows by hand, the toggle is still showing in the expanded position.

IMO we don't need the `expandCollapseToggle` state, and we can instead use the `areAllSelected` (renamed to `areAllExpanded`) property from `useSelectionState` to determine which toggle to show. i.e. if all rows are expanded, show it in the AngleDownIcon mode, otherwise the AngleRightIcon mode. This will make it behave like the indeterminate select-all checkbox: if there are any unselected/unexpanded rows, clicking it selects/expands all, and the deselect/collapse-all action is only available if all are selected/expanded.

seanforyou23: Ah ok, I'm starting to see the value in useSelectionState now. Very cool!

seanforyou23:
Ok, I've pushed some updates. I placed the pagination in compact mode in this one place to prevent wrapping of the toolbar items. Is it ok? There may be more I can do in terms of hiding/collapsing toolbar items at given viewport sizes, can explore if needed.

@mturley I got rid of that extra state, but unfortunately it didn't fix the issue of selecting "page" items in the bulk select. Notice the bottom pagination reflects the correct number of selected items, but the checkboxes themselves are not actually checked. In all other scenarios the checkboxes seem to accurately reflect the checked state.. Am I using `setSelectedItems` correctly?

Screen Shot 2021-06-24 at 10 52 27 AM

mturley:
Code LGTM. @vconzola, how does this look visually? I would think we should align the expand/collapse-all caret with the individual row carets, and the bulk select checkbox with the row checkboxes, no?

seanforyou23:
Ah, I see what you mean @mturley - updated as you suggested, looks much better now.

mturley:
@seanforyou23 it looks like there's a border around the expand-all toggle now, was that intentional?

This is looking great btw!

seanforyou23:
Yea, it felt like it was just floating before, so I added the control variant styling. Easier to see the bounds of the selectable area now.

mturley (approved):
LGTM! Will wait for a final review from @vconzola before merging.

PR #664, opened 2021-06-23
Fix redirect loop when rendering /handle-login route with an error param and no user param

mturley:
Also renders login errors in an Alert on the page instead of just in the console.

To test this, you can simulate a login error by navigating to:
```
/handle-login?error=%7B%22message%22%3A%22request+to+https%3A%2F%2Fkubernetes.default.svc.cluster.local%2F.well-known%2Foauth-authorization-server+failed%2C+reason%3A+self+signed+certificate+in+certificate+chain%22%2C%22type%22%3A%22system%22%2C%22errno%22%3A%22SELF_SIGNED_CERT_IN_CHAIN%22%2C%22code%22%3A%22SELF_SIGNED_CERT_IN_CHAIN%22%7D
```

This would previously trigger a redirect to `/` which would attempt to perform the login again, ending up on this route again and causing a loop.

### DIFF
--- a/src/app/common/LoginHandlerComponent.tsx
+++ b/src/app/common/LoginHandlerComponent.tsx
@@ -1,6 +1,15 @@
+import { Alert, AlertActionLink } from '@patternfly/react-core';
+import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import React, { useEffect } from 'react';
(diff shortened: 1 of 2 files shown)


mturley:
@fdupont-redhat if you want to merge this before you cut the release branch that would be great, it definitely needs to go in

gildub (approved):
LGTM

PR #667, opened 2021-06-24
Upgrade to lib-ui 5.0.0, replace setInitialValue/reset with prefill/clear in PlanWizard and other prefilled forms

### LINKED ISSUE #625: Regression: changing source/target providers while editing an existing plan no longer causes the rest of the wizard steps to be cleared
All of the options/selections on the filter/select VMs and mappings steps of the wizard are dependent on the selected source and target providers on the General step. If the user changes one of those provider selections, the wizard should clear all their selections in those dependent steps so they can start fresh. We implemented this with a useEffect [here](https://github.com/konveyor/forklift-ui/blob/edfb0b1/src/app/Plans/components/Wizard/PlanWizard.tsx#L192-L202):

```ts
  // When providers change, reset all forms containing provider-specific options
  React.useEffect(() => {
    if (!isFirstRender.current && isDonePrefilling) {
      forms.filterVMs.reset();
      forms.selectVMs.reset();
      forms.networkMapping.reset();
      forms.storageMapping.reset();
    }
    isFirstRender.current = false;
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [forms.general.values.sourceProvider, forms.general.values.targetProvider]);
```

However, in https://github.com/konveyor/forklift-ui/pull/365 we indirectly changed the behavior of what that `reset()` function does by introducing the `setInitialValue` function in lib-ui's useFormState. When you reset a form, it sets all the fields to their `initialValue`, which in the case of a plan being edited is the values prefilled from the existing plan, not the initial values from an empty wizard (because we switched from using `setValue` to `setInitialValue` in the `useEditingPlanPrefillEffect`. So if the user is editing a plan, when they change providers it doesn't actually clear anything. This can cause weird side effects as form state and query results from a different provider are hanging around when the user moves through the wizard.

The bug is unfortunately present in 2.0.0, but it seems like everything is fine in that version because there is only one type of source provider and one type of target provider. So loading things like `useMappingResourcesQuery` will have the same results no matter which source provider you choose. However, with https://github.com/konveyor/forklift-ui/pull/597 we are introducing RHV source providers, and I have found that if you are editing a VMware plan and change to a RHV source provider, it totally breaks the storage mappings step of the wizard since it fails to prefill required sources from the new query (storage domains rather than datastores).

To fix this, we need to make it so those `reset()` calls will actually reset the form to actual blank initial values instead of what was in the `planBeingEdited`. We can't just go back to using `setValue` in the `useEditingPlanPrefillEffect` though, because that will reintroduce the bug fixed in #365. Instead we need a way to retain the `initialValue`s when prefilling the plan, without setting `isDirty` when we prefill values.

I think a viable solution might be to add a `setIsDirty` parameter of the `setValue` function, which defaults to true. That way, we could use `setValue` for the prefilling, with that param set to false, and prevent the `isDirty` problem that was causing the save buttons to enable prematurely, but the `initialValue`s in the form state would still be an empty wizard's values so the `reset()` calls would have the intended effect. This would require modifying lib-ui and upgrading to a new version here.

mturley:
Resolves #625, making use of the useFormState changes in https://github.com/konveyor/lib-ui/pull/65.

Editing a plan and changing the target provider now fully clears the rest of the wizard as if you are creating a new plan, rather than resetting those values to the ones from the plan being edited (which will cause errors as it won't match the new provider).

Also it looks like upgrading the lib-ui package switched our lockfile from v1 to v2 again (first committed in https://github.com/konveyor/forklift-ui/pull/613), I think it's because I'm using a newer version of npm than you guys are @seanforyou23 @gildub . Do you think I should downgrade or do you guys mind upgrading?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,19828 @@
 {
   "name": "forklift-ui",
   "version": "0.0.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "version": "0.0.1",
+      "license": "Apache",
+      "dependencies": {
+        "@hot-loader/react-dom": "^17.0.1",
+        "@konveyor/lib-ui": "^5.0.0",
+        "@patternfly/react-core": "^4.128.2",
+        "@patternfly/react-icons": "^4.10.11",
+        "@patternfly/react-styles": "^4.10.11",
+        "@patternfly/react-table": "^4.27.24",
+        "@patternfly/react-tokens": "^4.11.12",
+        "axios": "^0.21.1",
+        "css-minimizer-webpack-plugin": "^3.0.1",
+        "dayjs": "^1.10.5",
+        "dayjs-ext": "^2.2.0",
+        "ejs": "^3.1.6",
+        "express": "^4.17.1",
+        "fast-deep-equal": "^3.1.3",
+        "http-proxy-middleware": "^2.0.0",
+        "js-yaml": "^4.1.0",
+        "netmask": "^2.0.2",
+        "node-fetch": "^2.6.1",
+        "q": "^1.5.1",
+        "react": "^17.0.2",
+        "react-dom": "^17.0.2",
+        "react-router-last-location": "^2.0.1",
+        "react-syntax-highlighter": "^15.4.3",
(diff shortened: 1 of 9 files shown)


gildub (approved):
LGTM

PR #669, opened 2021-06-24
Consistently use correct provider names based on upstream/downstream branding (oVirt/RHV and KubeVirt/OCPv)

mturley:
Currently even in Konveyor-branded upstream builds that can migrate from oVirt to KubeVirt, the UI displays downstream-branded provider names (Red Hat Virtualization and OpenShift Virtualization). This PR sets the correct name for each provider type in the `PROVIDER_TYPE_NAMES` constant based on the `BRAND_TYPE` env variable, and replaces hard-coded use of these names across the UI with references to that constant.


mturley:
cc @fdupont-redhat

fdupont-redhat (approved):

PR #673, opened 2021-06-29
Reduce TS types and mocks for inventory data to only the properties we actually use

### LINKED ISSUE #618: Enumerate the exact inventory API fields we actually need/use, and remove unused fields from TS interfaces
Discussing with @jortel, we'd like to remove unused data from the inventory API responses used by the UI for performance reasons, and having an understanding of exactly what the UI needs would also make it easier to make changes to the inventory API without worrying about risk of breaking the UI. We should simplify the interfaces in https://github.com/konveyor/forklift-ui/tree/main/src/app/queries/types so they are more readable and represent only what the UI actually needs.

mturley:
Resolves #618

As requested by @jortel. We had a bunch of properties in our TS types for these inventory objects that we weren't actually using, and it bloated the mock data as well as restricting backend changes unnecessarily since we didn't know what would break if things were removed/changed.

This PR reduces these types as much as possible based on the actual references in the app code. I also wrote up a wiki page to list the properties we're keeping for reference. https://github.com/konveyor/forklift-ui/wiki/UI-usage-of-inventory-API


seanforyou23 (approved):
Spring cleaning!

Issue #655, opened 2021-06-16
Improve RHV provider field names and tooltips


apinnick:
![image](https://user-images.githubusercontent.com/31725355/122190710-bd3b5600-ce9a-11eb-9183-0546fdabc21c.png)

The host name/IP, user name, and password are for the RHV Manager (upstream: Engine). You may want to add that to the field names or as a tooltip to the fields. 

The CA certificate field has a tooltip that is not very helpful because the RHV CA certificate location is documented in the RHV Administration Guide. Suggest you change the tooltip to: "The CA certificate is the `/etc/pki/ovirt-engine/apache-ca.pem` file on the Manager machine."

seanforyou23:
@apinnick I attached a PR for this, not sure I got it exactly right, but please review when you find time. I couldn't tag you as reviewer on the PR for some reason.. Thanks!

apinnick:
@seanforyou23 
I made a few comments about following UI and our style conventions on your PR. Other than those issues, LGTM.

mturley:
@apinnick would you mind opening a BZ for this?

PR #674, opened 2021-06-30
fix(providers): update form field labels for better readability

seanforyou23:
This PR updates the form field labels for the add provider modal. For VMware I used the "vCenter" prefix, and for RHV/oVirt I used the respective prefix for those as well. I also changed the helper text for Name/username fields based on some language from the docs, should be a little clearer for users who may be unfamiliar with what some of these properties do.

Screen Shot 2021-07-06 at 12 32 08 PM

Screen Shot 2021-07-06 at 12 32 15 PM

should help close https://github.com/konveyor/forklift-ui/issues/655

### DIFF
--- a/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
+++ b/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
@@ -53,6 +53,35 @@ const PROVIDER_TYPE_OPTIONS = PROVIDER_TYPES.map((type) => ({
   value: type,
 })) as OptionWithValue<ProviderType>[];
 
+const oVirtLabelPrefix = process.env.BRAND_TYPE === 'RedHat' ? 'RHV Manager' : 'oVirt Engine';
+const vmwareLabelPrefix = 'vCenter';
+const getLabelName = (type: 'hostname' | 'username' | 'pwd', prefix?: string) => {
+  let label = '';
+  switch (type) {
+    case 'hostname': {
+      label = prefix ? `${prefix} host name or IP address` : 'Host name or IP address';
+      break;
+    }
+    case 'username': {
+      label = prefix ? `${prefix} user name` : 'User name';
+      break;
+    }
+    case 'pwd': {
+      label = prefix ? `${prefix} password` : 'Password';
+      break;
+    }
+    default: {
+      label = 'Unknown label';
+    }
+  }
+  return label;
+};
+
+const isVmWare = (providerType: ProviderType | null) => providerType === 'vsphere';
+const isOvirt = (providerType: ProviderType | null) => providerType === 'ovirt';
+const brandPrefix = (providerType: ProviderType | null) =>
+  isOvirt(providerType) ? oVirtLabelPrefix : isVmWare(providerType) ? vmwareLabelPrefix : undefined;
+
 const useAddProviderFormState = (
   clusterProvidersQuery: ReturnType<typeof useClusterProvidersQuery>,
   providerBeingEdited: IProviderObject | null
@@ -76,9 +105,24 @@ const useAddProviderFormState = (
 
   const sourceProviderFields = {
     ...commonProviderFields,
-    hostname: useFormField('', hostnameSchema),
-    username: useFormField('', usernameSchema.required()),
-    password: useFormField('', yup.string().max(256).label('Password').required()),
+    hostname: useFormField(
+      '',
+      hostnameSchema.label(getLabelName('hostname', brandPrefix(providerTypeField.value)))
+    ),
+    username: useFormField(
+      '',
+      usernameSchema
+        .required()
+        .label(getLabelName('username', brandPrefix(providerTypeField.value)))
+    ),
+    password: useFormField(
+      '',
+      yup
+        .string()
+        .max(256)
+        .label(getLabelName('pwd', brandPrefix(providerTypeField.value)))
+        .required()
+    ),
   };
 
   return {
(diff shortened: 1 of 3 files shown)


src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
```diff
@@ -249,23 +249,23 @@ const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModalProps>
                 {fields?.hostname ? (
                   <ValidatedTextInput
                     field={fields.hostname}
-                    label="Hostname or IP address"
+                    label="RHV Manager Hostname or IP address"
```
apinnick, line 252: Following Patternfly4 and RHV conventions, this should be "RHV Manager host name or IP address"

src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
```diff
@@ -249,23 +249,23 @@ const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModalProps>
...
                     isRequired
                     fieldId="hostname"
                   />
                 ) : null}
                 {fields?.username ? (
                   <ValidatedTextInput
                     field={fields.username}
-                    label="Username"
+                    label="RHV Manager Username"
```
apinnick, line 260: Change to "RHV Manager user name"

src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
```diff
@@ -249,23 +249,23 @@ const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModalProps>
...
                     isRequired
                     fieldId="username"
                   />
                 ) : null}
                 {fields?.password ? (
                   <ValidatedPasswordInput
                     field={fields.password}
-                    label="Password"
+                    label="RHV Manager Password"
```
apinnick, line 268: RHV Manager password

mturley:
@apinnick what about in upstream builds? We're using the oVirt branding on providers instead of RHV when building the UI with Forklift rather than MTV branding. Is "oVirt Manager" still the correct term in that case?

mturley (changes requested):
@seanforyou23 one current drawback of useFormState is that we have to duplicate the field names in both the schema and the field labels themselves. You'll need to modify these in both places for the validation errors to appear correctly (try focusing and unfocusing a required field to see an example).

In this case, if you look above where we define `const sourceProviderFields` (fields used for both VMware and RHV/oVirt providers), we have `hostnameSchema`. If you go to where that is declared you'll see it's a yup schema with `.label('Hostname or IP address')`, which defines how the field is named in validation error messages.

Since we now need that field to have different labels for the different provider types, what we probably need to do is remove the `.label()` from that imported schema and move `hostname` from `sourceProviderFields` into both the `vsphere: useFormState(...)` and `ovirt: useFormState(...)` below, with separate `hostnameSchema.label(...)`s in there.

Similarly we need to change the `.label()` on the username and password fields used here as well, maybe we need to just get rid of `sourceProviderFields` and have them all defined in both forms.

Left a comment above for Avital about oVirt/RHV branding as well, these labels may need to be conditional based on `ENV.BRAND_TYPE`.

seanforyou23:
Super helpful, thx @mturley - I'll make adjustments and send an update soon.

mturley:
@apinnick should we also put different labels on these fields when adding a VMware provider? They are the same, just "Hostname or IP address", "Username" and "Password".

Screen Shot 2021-07-01 at 2 24 46 PM

mturley (approved):
Code LGTM! I'll wait for Avital's answer about VMware, we might as well do this all at once if we need a BZ for each backported PR now.

Also, to follow up later on the label duplication issue, I opened this: https://github.com/konveyor/lib-ui/issues/69

apinnick:
> @apinnick what about in upstream builds? We're using the oVirt branding on providers instead of RHV when building the UI with Forklift rather than MTV branding. Is "oVirt Manager" still the correct term in that case?

@mturley The correct upstream term for RHV Manager is oVirt Engine.

apinnick:
> @apinnick should we also put different labels on these fields when adding a VMware provider? They are the same, just "Hostname or IP address", "Username" and "Password".
> 
> Screen Shot 2021-07-01 at 2 24 46 PM

@mturley These labels are not correct for VMware. I'm sure that we updated the provider fields a while ago. All lthose fields are for vCenter. That is very important because we don't want the user to mistake "host" for ESXi host.

image

mturley:
@apinnick 

> These labels are not correct for VMware

Which ones, the ones in my screenshot? I do remember we updated them at some point, but I just wasn't sure if we should give them more specific labels since the equivalent fields for RHV are getting more specific labels.

Those field descriptions are from the documentation, right? Should they also be in the UI under help icons like the name and fingerprint fields?

apinnick:
@mturley 
> Which ones, the ones in my screenshot? I do remember we updated them at some point, but I just wasn't sure if we should give them more specific labels since the equivalent fields for RHV are getting more specific labels.

It's not so much that they are wrong (I overstated that) but that they needed documentation because users were confused about whether they needed the vCenter or ESXi credentials, etc.

If the fields do not have to be re-used for other providers and if the RHV fields will be specific, I think it makes sense to add "vCenter" to the labels:

- vCenter host name or IP address
- vCenter admin user name
- vCenter admin password

> 
> Those field descriptions are from the documentation, right? 

Yes, that is a screenshot of the user guide. 

> Should they also be in the UI under help icons like the name and fingerprint fields?

If you are not planning to update the field labels, then they should be in the tooltips or help icons. But I think it would be easier for the user if you update the field labels with "vCenter ..."

seanforyou23:
@apinnick I pushed some updates and updated the PR description to reflect the latest changes, let me know what you think when you have time to take another look. Thank you!

mturley (approved):
LGTM, thanks @seanforyou23 !

@apinnick if you have more feedback here we can address it in a followup PR, I just want to move this along.

PR #675, opened 2021-06-30
Simplify and optimize Filter/Select VMs code: pre-index tree and VM data only once per new data, drive everything from the pre-indexed data

### LINKED ISSUE #704: With large data sets, the app can freeze for several seconds while indexing tree and VM data
In the current main branch, this happens upon reaching the `SelectVMsForm` in the wizard and calculating `availableVMs` and `treePathInfoByVM` in useEffect calls. A spinner is showing while it freezes, so it's acceptable for now. It also only seems to happen when there is a very large amount of data. See my branch https://github.com/mturley/forklift-ui/tree/tmp-perf-compare for mock data with 4000+ VMs to reproduce this with.

This is being impacted by https://github.com/konveyor/forklift-ui/pull/675 which moves that processing into the query hooks, but still has this issue. Planning to fix it as a followup so that PR doesn't get more bloated.

mturley:
Resolves https://github.com/konveyor/forklift-ui/issues/704

This PR resolves the technical debt introduced in kubev2v/forklift-ui#600 and kubev2v/forklift-ui#609, which were temporary hacks to fix performance issues in v2.0. Here I am attempting to do those things right, with even better performance and hopefully much more readable code.

In the plan wizard, the Filter and Select VMs steps involve a lot of data traversal. The inventory tree queries and the VMs query provide raw data that must be recursively traversed, and the current code does that stuff fairly redundantly and in messy useEffects that attempt to cache the results asynchronously. It works okay, but there are issues (see https://github.com/konveyor/forklift-ui/issues/704). In particular, when there are a lot of VMs these algorithms start to have a real performance impact since they are constantly looping over the whole VM list and walking the whole tree on every render, in some cases multiple times per render.

This PR moves all of the expensive traversal of tree and VM data into the `useSourceVMsQuery` and `useInventoryTreeQuery` hooks themselves, by way of the new `select` option provided by react-query v3. The new helpers `indexVMs` and `indexTree` are called there, and the result is memoized and only recalculated if the query returned new data on any given fetch. These hooks now return `IndexedTree` and `IndexedSourceVMs` objects, which contain a copy of the original data along with various properties with that data indexed by convenient keys. And because these query hooks are mounted in the GeneralForm, all this indexing happens in the background as soon as a source provider is selected, and by the time the user proceeds to the next wizard steps there is no loading spinner to be seen at all.

All of the logic elsewhere in the wizard steps and their helpers that was previously doing repetitive traversal of the tree or VMs list now references the indexed fields like `vmsById` or `descendantsBySelfLink`, performing these lookups in O(1) time instead of O(n) or O(n^2).

I have a branch in my fork called [optimize-tree-indexing--tmp-perf](https://github.com/mturley/forklift-ui/tree/optimize-tree-indexing--tmp-perf) with the contents of this PR plus a large dataset of mock tree and VM data (over 4000 VMs). I deployed a preview of this branch here: https://tmp-forklift-pr675-large-data-preview.surge.sh/plans/create

If you try operating the plan wizard you can see that everything is responsive and feels instant. Compare that to the same dataset cherry-picked on the current main branch: http://tmp-forklift-issue704-large-data-preview.surge.sh/plans/create . On this branch when you reach the Select VMs step of the wizard it sits on a spinner for a long while (the browser is actually frozen, you can see that CSS hover effects no longer work during this time; see https://github.com/konveyor/forklift-ui/issues/704).

The one downside of this approach is that all the tree helpers now have to have the `indexedTree` passed around as an extra parameter to do lookups with. In a future PR I may clean up the way those helpers are defined so we don't have all those repetitive parameters, but this PR was getting bloated so I'll leave that alone for now.

@gildub @seanforyou23 in an earlier version of this branch I played around with the `use-async-memo` package, but I ended up removing it after all since I got things performing well enough without it. Something to keep in mind for any future performance struggles though, it is a nice tool.

And yes, I did write unit tests for the indexing helpers!

### DIFF
--- a/src/app/Plans/components/Wizard/FilterVMsForm.tsx
+++ b/src/app/Plans/components/Wizard/FilterVMsForm.tsx
@@ -14,9 +14,7 @@ import {
   findMatchingNode,
   findMatchingNodeAndDescendants,
   findNodesMatchingSelectedVMs,
-  flattenInventoryTreeNodes,
   getAvailableVMs,
-  getSelectableNodes,
   getSelectedVMsFromPlan,
   isNodeFullyChecked,
   useIsNodeSelectableCallback,
@@ -33,8 +31,6 @@ interface IFilterVMsFormProps {
   planBeingEdited: IPlan | null;
 }
 
-// TODO also for the folder path column, make a hash of folder ids to tree nodes, so we can walk up their parents to build the path string
-
 const FilterVMsForm: React.FunctionComponent<IFilterVMsFormProps> = ({
   form,
   sourceProvider,
@@ -50,7 +46,7 @@ const FilterVMsForm: React.FunctionComponent<IFilterVMsFormProps> = ({
   const isNodeSelectable = useIsNodeSelectableCallback(form.values.treeType);
 
   const treeSelection = useSelectionState({
-    items: getSelectableNodes(treeQuery.data || null, isNodeSelectable),
+    items: treeQuery.data?.flattenedNodes.filter(isNodeSelectable) || [],
     externalState: [form.fields.selectedTreeNodes.value, form.fields.selectedTreeNodes.setValue],
     isEqual: (a: InventoryTree, b: InventoryTree) => a.object?.selfLink === b.object?.selfLink,
   });
@@ -67,7 +63,7 @@ const FilterVMsForm: React.FunctionComponent<IFilterVMsFormProps> = ({
       } else if (vmsQuery.isSuccess && treeQuery.isSuccess) {
         const selectedVMs = getSelectedVMsFromPlan(planBeingEdited, vmsQuery);
         const selectedTreeNodes = findNodesMatchingSelectedVMs(
-          treeQuery.data || null,
+          treeQuery.data.tree,
           selectedVMs,
           isNodeSelectable
         );
@@ -89,7 +85,10 @@ const FilterVMsForm: React.FunctionComponent<IFilterVMsFormProps> = ({
   const getNodeBadgeContent = (node: InventoryTree, isRootNode: boolean) => {
     const { treeType } = form.values;
     const { isItemSelected, selectedItems } = treeSelection;
-    const selectedDescendants = flattenInventoryTreeNodes(node).filter(isItemSelected);
+    const selectedDescendants = [
+      node,
+      ...(treeQuery.data?.descendantsBySelfLink[node.object?.selfLink || ''] || []),
+    ].filter(isItemSelected);
     const numVMs = getAvailableVMs(selectedDescendants, vmsQuery.data || [], treeType).length;
     const rootNodeSuffix = ` VM${numVMs !== 1 ? 's' : ''}`;
     if (numVMs || isItemSelected(node) || (isRootNode && selectedItems.length > 0)) {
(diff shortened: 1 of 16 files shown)


mturley:
I'm pretty sure the code is done here, but I want to test it with a large dataset and write up a good description before I mark it ready.

mturley:
Note to self: since several of these helpers have to pass around `indexedTree`, `isNodeSelected`, `isNodeSelectable` and `getNodeBadgeContent`, it might be useful later on to create a closure like `useTreeHelpers` or `getTreeHelpers` (depending on whether it needs to call hooks) that takes these params and keeps them in scope for all internal functions, to shorten the code a bit. Then we can call that and destructure helpers from it instead of importing them.

src/app/Plans/components/Wizard/FilterVMsForm.tsx
```diff
@@ -87,17 +85,35 @@ const FilterVMsForm: React.FunctionComponent<IFilterVMsFormProps> = ({
...
+    const selectedDescendants = allDescendants.filter(isItemSelected);
+    const numVMs = getAvailableVMs(
+      treeQuery.data,
+      selectedDescendants,
+      vmsQuery.data,
+      treeType
+    ).length;
     const rootNodeSuffix = ` VM${numVMs !== 1 ? 's' : ''}`;
     if (numVMs || isItemSelected(node) || (isRootNode && selectedItems.length > 0)) {
       return `${numVMs}${isRootNode ? rootNodeSuffix : ''}`;
```
seanforyou23, line 103: Doesn't this represent the number of VMs selected from _descendents_ of this node in the tree? At first I thought it was a summation of all VMs.. Maybe we can make the badge text more consistent in some way? Maybe we should use a Label component here instead of Badge since it's not only a numeric count inside it.

src/app/Plans/components/Wizard/FilterVMsForm.tsx
```diff
@@ -130,35 +146,32 @@ const FilterVMsForm: React.FunctionComponent<IFilterVMsFormProps> = ({
...
-                treeViewItem.id || '',
-                isNodeSelectable
-              );
-              if (nodesToSelect.length > 0) {
-                treeSelection.selectMultiple(nodesToSelect, !isFullyChecked);
+            } else if (treeQuery.data) {
+              const matchingPath = treeQuery.data.pathsBySelfLink[treeViewItem.id || ''];
+              if (matchingPath) {
+                const matchingNode = matchingPath[matchingPath.length - 1];
+                const isFullyChecked = isNodeFullyChecked(
```
seanforyou23, line 161: Not sure if this is the right line that handles this, but shouldn't the root node be indeterminate state if the children are partially selected?

Screen Shot 2021-07-22 at 3 33 11 PM

seanforyou23 (commented):
Looking good! Left a couple comments, only other thing is a overall question about selection state. If a data center folder is selected that has "loose" VMs inside it, the badge count shows the number of VMs, but it's not so clear where they are coming from. Maybe we could shuffle these "loose" VMs into an "uncategorized" folder or something? It's just a little strange that the count bubbles are sometimes attached to VM leaf nodes and sometimes attached to the data center.

Screen Shot 2021-07-22 at 3 41 54 PM

src/app/queries/tree.ts
```diff
@@ -1,15 +1,79 @@
...
+      if (!b || !b.object) return 1;
+      return a.object.name < b.object.name ? -1 : 1;
+    }),
+});
+
+export interface IndexedTree<T extends InventoryTree = InventoryTree> {
+  tree: T;
+  flattenedNodes: T[];
+  vmSelfLinks: string[];
+  pathsBySelfLink: Record<string, T[] | undefined>; // Flattened list of nodes leading to each node
```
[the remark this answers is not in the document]
mturley: Renamed and pushed.

src/app/queries/tree.ts
```diff
@@ -1,15 +1,79 @@
...
+  descendantsBySelfLink: Record<string, T[] | undefined>; // Flattened list of nodes under each node
+  vmDescendantsBySelfLink: Record<string, T[] | undefined>; // Flattened list of only VM nodes under each node
+  getDescendants: (node: InventoryTree, includeNode?: boolean) => InventoryTree[];
```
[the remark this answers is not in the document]
mturley: Renamed and pushed.

src/app/Plans/components/Wizard/FilterVMsForm.tsx
mturley: That is a fantastic point, and something we should follow up on. Opened an issue: https://github.com/konveyor/forklift-ui/issues/718

src/app/Plans/components/Wizard/FilterVMsForm.tsx
mturley: Yep, that's totally a bug. I think I should actually fix it in this PR before I forget how all this works. Commit coming shortly.

mturley: Fixed and pushed.

seanforyou23 (approved):
😎

mturley:
Feels good man 🚿

Issue #423, opened 2021-02-16
Update migration status code to use new VM-level `conditions` instead of hard-coded heuristics


mturley:
@jortel informed me that there are new `conditions` on each VM in the Migration CR (and `status` section of the Plan CR). We use a lot of logic based on other status properties currently to decide which states to render, and we can simplify this by using those clear conditions instead.

mturley:
@jortel I'm starting to worry about warm migration and hooks for this sprint. Are you ok with me putting this at lower priority than those, possibly bumping it to after feature freeze (considering it a blocking bug)?

mturley:
Note to self: as part of fixing this, we should also simplify the plan-level status logic. See TODO comments added in this commit: https://github.com/konveyor/forklift-ui/pull/440/commits/a6bbec5cd539dc93985255121042dcd5defd0f50

seanforyou23:
>Note to self: as part of fixing this, we should also simplify the plan-level status logic. See TODO comments added in this commit: a6bbec5

Made some progress toward this work with #678 - but it was a little more involved than I first realized so the vm-level conditions refactor work remains. Going to leave this issue open and remove myself from assignment for now in case anyone else wants to help tag team it, otherwise I can resume this work at a later point.

PR #678, opened 2021-07-01
Mig status conditions refactor

seanforyou23:
This PR refactors some logic around how the plan status gets reported on the migration plans page. It should be a transparent change in terms of user experience, just that now the code is a step toward more readable. 

Steps toward closing https://github.com/konveyor/forklift-ui/issues/679 and https://github.com/konveyor/forklift-ui/issues/423

### DIFF
--- a/src/app/Plans/components/PlansTable.tsx
+++ b/src/app/Plans/components/PlansTable.tsx
@@ -5,7 +5,6 @@ import {
   Pagination,
   Progress,
   ProgressMeasureLocation,
-  ProgressVariant,
   Text,
   ToolbarItem,
 } from '@patternfly/react-core';
@@ -34,7 +33,6 @@ import { useSelectionState } from '@konveyor/lib-ui';
 import PlanActionsDropdown from './PlanActionsDropdown';
 import { useSortState, usePaginationState } from '@app/common/hooks';
 import { IPlan } from '@app/queries/types';
-import { PlanStatusDisplayType, PlanStatusType } from '@app/common/constants';
 import CreatePlanButton from '@app/Plans/components/CreatePlanButton';
 import { FilterToolbar, FilterType, FilterCategory } from '@app/common/components/FilterToolbar';
 import { useFilterState } from '@app/common/hooks/useFilterState';
@@ -48,12 +46,7 @@ import {
 } from '@app/queries';
 
 import './PlansTable.css';
-import {
-  canPlanBeStarted,
-  getPlanStatusTitle,
-  getWarmPlanState,
-  isPlanBeingStarted,
-} from './helpers';
+import { getPlanStatusTitle, getPlanState, getButtonState, getMigStatusState } from './helpers';
 import { isSameResource } from '@app/queries/helpers';
 import StatusCondition from '@app/common/components/StatusCondition';
 import MigrateOrCutoverButton from './MigrateOrCutoverButton';
@@ -126,10 +119,10 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
       ],
       getItemValue: (item) => {
         if (item.status?.conditions) {
-          if (hasCondition(item.status.conditions, PlanStatusType.Failed)) return 'Failed';
-          if (hasCondition(item.status.conditions, PlanStatusType.Ready)) return 'Ready';
-          if (hasCondition(item.status.conditions, PlanStatusType.Executing)) return 'Running';
-          if (hasCondition(item.status.conditions, PlanStatusType.Succeeded)) return 'Succeeded';
+          if (hasCondition(item.status.conditions, 'Failed')) return 'Failed';
+          if (hasCondition(item.status.conditions, 'Ready')) return 'Ready';
+          if (hasCondition(item.status.conditions, 'Executing')) return 'Running';
+          if (hasCondition(item.status.conditions, 'Succeeded')) return 'Succeeded';
         }
         return '';
       },
@@ -196,47 +189,12 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
   const rows: IRow[] = [];
 
   currentPageItems.forEach((plan: IPlan) => {
-    let buttonType: PlanActionButtonType | null = null;
-    let title = '';
-    let variant: ProgressVariant | undefined;
-
-    const conditions = plan.status?.conditions || [];
     const latestMigration = findLatestMigration(plan, migrationsQuery.data?.items || null);
+    const isWarmPlan = plan.spec.warm;
 
-    // TODO This whole if-else pile should instead be reduced to a union type like WarmPlanState but generalized.
-    // TODO the PlanStatusType should be a union PlanConditionType and the PlanStatusDisplayType should perhaps not be a thing.
-    // TODO note, buttonType and title will only be used if <Progress> is shown in the row below, so their values don't matter for pre-cutover.
-    if (hasCondition(conditions, PlanStatusType.Ready) && !plan.status?.migration?.started) {
-      buttonType = 'Start';
-    } else if (hasCondition(conditions, PlanStatusType.Executing)) {
-      buttonType = null;
-      title = PlanStatusDisplayType.Executing;
-      if (plan.spec.warm) {
-        title = 'Running cutover';
-        if (!latestMigration?.spec.cutover) {
-          buttonType = 'Cutover';
-        }
-      }
-    } else if (hasCondition(conditions, PlanStatusType.Succeeded)) {
-      title = PlanStatusDisplayType.Succeeded;
-      variant = ProgressVariant.success;
-    } else if (hasCondition(conditions, PlanStatusType.Canceled)) {
-      title = PlanStatusDisplayType.Canceled;
-    } else if (hasCondition(conditions, PlanStatusType.Failed)) {
-      title = PlanStatusDisplayType.Failed;
-      variant = ProgressVariant.danger;
-    } else if (plan.status?.migration?.started) {
-      if (canPlanBeStarted(plan)) {
-        title = 'Finished - Incomplete';
-        variant = ProgressVariant.warning;
-      } else {
-        console.warn('Migration plan unexpected status:', plan);
-      }
-    }
-
-    if (buttonType !== 'Start' && canPlanBeStarted(plan)) {
-      buttonType = 'Restart';
-    }
+    const planState = getPlanState(plan, latestMigration, migrationsQuery);
+    const buttonType = getButtonState(planState);
+    const { title, variant } = getMigStatusState(planState, isWarmPlan);
 
     const { statusValue = 0, statusMessage = '' } = ratioVMs(plan);
 
@@ -247,9 +205,7 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
 
     const isExpanded = isPlanExpanded(plan);
 
-    const warmState = getWarmPlanState(plan, latestMigration, migrationsQuery);
-    // TODO this is redundant with getWarmPlanState's 'Starting' case, maybe generalize that helper.
-    const isBeingStarted = isPlanBeingStarted(plan, latestMigration, migrationsQuery);
+    const isBeingStarted = planState === 'Starting';
 
     rows.push({
       meta: { plan },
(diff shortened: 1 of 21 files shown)


mturley:
@seanforyou23 can you remove the imports that are no longer being used? looking at the review diff you can see GitHub Actions has surfaced those eslint warnings

src/app/Plans/components/PlansTable.tsx
```diff
@@ -196,47 +196,14 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
...
-      } else {
-        console.warn('Migration plan unexpected status:', plan);
-      }
-    }
-
-    if (buttonType !== 'Start' && canPlanBeStarted(plan)) {
-      buttonType = 'Restart';
-    }
+    const planState = getPlanState(plan, latestMigration, migrationsQuery);
+    const newButtonType = getButtonState(planState, plan, latestMigration);
```
mturley, line 205: I think we can call this `buttonType` since there won't be an old and a new one anymore :)

src/app/Plans/components/helpers.ts
```diff
@@ -1,33 +1,133 @@
...
-// TODO maybe generalize this for cold migrations too
-type WarmPlanState =
-  | 'NotStarted'
-  | 'Starting'
-  | 'Copying'
-  | 'AbortedCopying'
-  | 'StartingCutover'
-  | 'Cutover'
-  | 'Finished';
+export const getMigStatusState = (plan: IPlan) => {
```
mturley, line 22: Can this be driven from the `PlanState` as well instead of directly from `hasCondition` cases? If they don't map 1:1 we may want to add/change the states of a plan so that they do. I see you have a TODO below to add different types of Finished states, e.g. Canceled. That may be worth doing before we introduce duplicate logic. (I guess there already is some duplicate logic... we could do that in a followup PR if you want)

src/app/Plans/components/helpers.ts
```diff
@@ -1,33 +1,133 @@
...
+      type = 'Restart';
+      break;
+    }
+    case state === 'StartingCutover': {
+      type = null; // spinner
+      break;
+    }
+    // "Cutover" name is warm-specific, but this state applies to both warm and cold
+    // it could just be called 'Migrating' or 'PipelineRunning'
+    case state === 'Cutover': {
```
mturley, line 108: I think we should go ahead and rename `Cutover` so it's not warm-specific, that could get confusing if we forget to do it later.

src/app/Plans/components/helpers.ts
```diff
@@ -1,33 +1,133 @@
...
+    }
+    default: {
+      type = null;
+    }
+  }
+
+  return type;
+};
+
+export const getPlanState = (
```
mturley, line 130: On line 135 here in `getPlanState`, you return null if `!plan.spec.warm`. This means `planState` is null for every single cold plan. Interestingly enough this doesn't seem to be affecting any of the mocked plans in the preview, because:

1. the only two cases in the Status column we show are the `<StatusIcon>` case (which you also show if `!plan.status?.migration?.started`) and the `<Progress>` case (which is the base case, it's showing for all the other plans because `planState` is null)
2. the Progress case uses `getMigStatusState` which isn't using the `planState` at all anyway.
3. the `getButtonState` function is using `planState`, but the only 2 cases it's hitting for cold plans are `'Restart'` (which you're also using `plan.status?.migration?.started` for), `Start` (which you're also using `plan.status.conditions` for in a separate case from the NotStarted one) and null, which is your default case anyway.

But it will behave incorrectly in other cases we aren't mocking: for example, we'll no longer get a "Preparing for migration" case when a cold plan would be in the `Starting` state. We may want to add a mock plan to show this state.

You should remove the `!plan.spec.warm` condition from `getPlanState` so that you actually have a defined `planState` for every plan, and then I would think we should drive the `getButtonState` and `getMigStatusState` entirely from `planState` and not touch any of the `plan.status` stuff in it if we can help it, so all of that logic is in `getPlanState`.

We should also probably remove all the usages of `isPlanBeingStarted` and drive those all from the `'Starting'` planState. Single source of truth, single source of plan state logic :)

src/app/queries/constants.ts
```diff
@@ -1,4 +1,4 @@
-export const POLLING_INTERVAL = 5000;
+export const POLLING_INTERVAL = process.env.NODE_ENV !== 'production' ? 10000 : 5000;
```
mturley, line 1: I like this!

mturley (changes requested):
A couple of things to consider here. This is really starting to come together though!

mturley:
One other thing: https://github.com/konveyor/forklift-ui/issues/423 has to do specifically with the VM-level conditions in the pipeline status (for the VMMigrationDetails page), not the plan-level conditions we're working with here. That will probably require a separate PR to fully close after we get this one nailed down.

src/app/Plans/components/PlansTable.tsx
seanforyou23: lol whoops, leftovers from debug mode - will update

src/app/Plans/components/helpers.ts
seanforyou23: Yep, totally makes sense - I'll go ahead and do it as part of this PR

src/app/Plans/components/helpers.ts
mturley: It does actually look like some of the logic won't work right for cold plans as well. Any cold plan with the Executing condition will end up in the Starting state.. you may want to have that whole `if` block affect only `plan.spec.warm` plans, and have another one that puts cold plans in the `RunningPipeline` state (or whatever we rename `Cutover` to).

seanforyou23:
@mturley gotcha, I saw a comment in 423 that mentioned to simplify the plan-level status logic and just assumed that was what we've been working with so far. Can definitely get all the VM conditions sorted out as part of this PR as well. Are those the status markings on the Mappings page table?

I noticed a 404 for the fallback/default favicon at some point during development, and now that I pointed it to an up to date location I see it overrides the correct one at runtime. I'll get that fixed, let me know if I need to split that out into another PR.

mturley:
> Can definitely get all the VM conditions sorted out as part of this PR as well. Are those the status markings on the Mappings page table?

No, the Mappings status is unique and weird, it's driven from a function that just checks that all the networks/storages in the mapping really exist. That's not coming from any API-side status data. The VM conditions the issue refers to have to do with the states on the VMMigrationDetails page (click on a running plan to see it, it's the one with the pipeline steps). I think those should maybe be a separate PR since it's new logic instead of just consolidating the logic we have. If it's easy enough to add though we can try.

I'm fine with fixing the favicon in here if you want since it's a trivial thing. We can probably just remove it entirely from that file because I think webpack sticks one in there at compile time.

src/app/Plans/components/helpers.ts
seanforyou23: Ok I _think_ I addressed most all of this. There's still a couple of places in the code that I think could be cleaned up a little further, take a look when you get time and let me know what you think!

src/app/Plans/components/PlansTable.tsx
```diff
@@ -265,34 +222,46 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
...
+              </PlanStatusNavLink>
+            ) : !plan.status?.migration?.started || planState === 'NotStarted' ? (
+              <StatusCondition status={plan.status} />
+            ) : planState === 'Copying' ? (
+              <PlanStatusNavLink plan={plan}>
+                Running - performing incremental data copies
+              </PlanStatusNavLink>
+            ) : planState === 'StartingCutover' ? (
+              <PlanStatusNavLink plan={plan}>Running - preparing for cutover</PlanStatusNavLink>
+            ) : (
```
mturley, line 253: A couple things about the `isBeingStarted` states:
* I don't think we want to render the <Progress> bar when the plan is being started, it should always show one of the "preparing for ___" states in that case. In fact I think the "cold plan being started" case will be covered by line 239 and you can get rid of the whole case starting on line 228.
* Remember "incremental data copies" is a warm migration thing. You have the isWarmPlan cases backwards, on line 239 you want `!isWarmPlan` instead of `isWarmPlan`. That will make the line 231 case only hit for warm plans which is correct, but if you want to be more explicit there you can also add `&& isWarmPlan` on that line.
* On line 245 I think we should get rid of the `!plan.status?.migration?.started` condition and just use `planState === 'NotStarted'`, we should be in that state if that started property is falsy anyway. See below comment in getPlanState.

We may want to add some new mock plans to show the "starting" state for warm and cold, or at least one or the other.

src/app/Plans/components/helpers.ts
```diff
@@ -1,122 +1,175 @@
...
-  if (!plan || !plan.spec.warm) return null;
+  migration: IMigration | null
+): PlanState | null => {
+  if (!plan) return null;
+
   if (!migration) return 'NotStarted';
-  if (isPlanBeingStarted(plan, migration, migrationsQuery)) {
-    return 'Starting';
-  }
+
```
mturley, line 114: To match my above comment, this should be `if (!migration || !plan.status?.migration?.started) return 'NotStarted';`

Also, it doesn't look like this function ever returns the `'Starting'` state. I see you deleted the `isPlanBeingStarted` function, I think you still need that code, just as a private helper that is only called in `getPlanState`. That probably belongs as an `if(isPlanBeingStarted(plan)) return 'Starting'` here after the NotStarted case.

src/app/Plans/components/helpers.ts
mturley: Resolving this because you got most of it, will open another thread for new comments.

src/app/Plans/components/helpers.ts
```diff
@@ -1,122 +1,175 @@
...
   const conditions = plan.status?.conditions || [];
+
   if (
-    (hasCondition(conditions, PlanStatusType.Canceled) ||
-      hasCondition(conditions, PlanStatusType.Failed)) &&
+    (hasCondition(conditions, 'Canceled') || hasCondition(conditions, 'Failed')) &&
```
mturley, line 118: AbortedCopying is only relevant for warm plans. This will incorrectly put cold canceled/failed plans in the AbortedCopying state because only warm plans can ever have `migration.spec.cutover`.

src/app/Plans/components/helpers.ts
```diff
@@ -1,122 +1,175 @@
...
     return 'AbortedCopying';
   }
-  if (!!migration && !!plan.status?.migration?.completed) return 'Finished';
-  if (hasCondition(conditions, PlanStatusType.Executing)) {
+
+  if (!!plan.status?.migration?.started && canPlanBeStarted(plan)) {
+    return 'Finished-Incomplete';
+  }
+
+  if (!!migration && !!plan.status?.migration?.completed) {
```
mturley, line 128: This logic will return Finished-Succeeded if a plan has a `completed` timestamp, but plans get that timestamp even if they are failed or canceled. `plan.status?.migration?.completed` just means the plan is no longer running. It may be simplest to move your Canceled and Finished-Failed cases from line 152-159 to above this code so the rest of the logic can assume there was no failure.

Wait, hang on... you have two cases that return Finished-Succeeded? I think if you move all 3 of the cases from the bottom (Finished-Succeeded, Canceled, Finished-Failed) above this one you can delete this one.

src/app/Plans/components/helpers.ts
```diff
@@ -1,122 +1,175 @@
...
     );
     const cutoverTimePassed =
       migration?.spec.cutover && new Date(migration.spec.cutover).getTime() < new Date().getTime();
+
     if (cutoverTimePassed && !pipelineHasStarted) {
       return 'StartingCutover';
     }
     if (cutoverTimePassed && pipelineHasStarted) {
-      return 'Cutover';
+      return 'PipelineRunning';
```
mturley, line 143: Again, cutover time is a warm migration thing, cold migrations will never have a cutover time and therefore they will never hit this case.

Oh.. but you return PipelineRunning as a base case, so that's okay... yeah I suppose that's okay.

The Copying case also doesn't check if a plan is warm, but it's okay because a cold plan won't have a precopies array... hmm.

I think it might be more clear if you put the 3 warm-specific states (AbortedCopying, StartingCutover, Copying) under an `if(plan.spec.warm) {}` block. Cold plans don't need to bother running that logic and it'll be more readable imo.

mturley (changes requested):
A couple more things. I think you're almost there. Let me know if you want to discuss them or pair program, it might be easier for me to make the code changes than explain them.

src/app/Plans/components/PlansTable.tsx
```diff
@@ -247,9 +205,8 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
 
     const isExpanded = isPlanExpanded(plan);
 
-    const warmState = getWarmPlanState(plan, latestMigration, migrationsQuery);
-    // TODO this is redundant with getWarmPlanState's 'Starting' case, maybe generalize that helper.
-    const isBeingStarted = isPlanBeingStarted(plan, latestMigration, migrationsQuery);
+    const isBeingStarted = planState === 'Starting';
+    console.log(plan.metadata.name, isBeingStarted, planState);
```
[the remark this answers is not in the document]
seanforyou23: Oh yea, anything goes with WIP 😆

src/app/Plans/components/PlansTable.tsx
seanforyou23: Agree, not having the "starting" state left me having to guess at a few things - I'd love to get one for warm/cold just to be exhaustive. Maybe in a follow-up PR?

src/app/Plans/components/helpers.ts
```diff
@@ -1,92 +1,175 @@
...
+
   const conditions = plan.status?.conditions || [];
-  if (
-    (hasCondition(conditions, PlanStatusType.Canceled) ||
-      hasCondition(conditions, PlanStatusType.Failed)) &&
-    !migration.spec.cutover
-  ) {
-    return 'AbortedCopying';
+
+  if (isWarm) {
```
mturley, line 123: If you're not going to combine this case with the other isWarm cases, you can just put `isWarm &&` into the if statement you have inside here instead of wrapping it

src/app/Plans/components/helpers.ts
```diff
@@ -1,92 +1,159 @@
...
+
+  if (!migration || !plan.status?.migration?.started) return 'NotStarted';
+  if (isPlanBeingStarted(plan, migration, migrationQuery)) return 'Starting';
+
   const conditions = plan.status?.conditions || [];
+
   if (
-    (hasCondition(conditions, PlanStatusType.Canceled) ||
-      hasCondition(conditions, PlanStatusType.Failed)) &&
+    isWarm &&
```
mturley, line 109: I think we need to replace:
```ts
if (
  isWarm &&
  (hasCondition(conditions, 'Canceled') || hasCondition(conditions, 'Failed')) &&
  !migration.spec.cutover
) {
  return 'AbortedCopying';
}
```
with:
```ts
if (isWarm && !migration.spec.cutover) {
  if (hasCondition(conditions, 'Canceled')) {
    return 'CanceledCopying';
  }
  if (hasCondition(conditions, 'Failed')) {
    return 'FailedCopying';
  }
}
```

src/app/Plans/components/VMMigrationDetails.tsx
```diff
@@ -103,12 +102,10 @@ const VMMigrationDetails: React.FunctionComponent = () => {
...
   const migrationsQuery = useMigrationsQuery();
   const latestMigration = findLatestMigration(plan || null, migrationsQuery.data?.items || null);
-  const warmPlanState = getWarmPlanState(plan || null, latestMigration, migrationsQuery);
+  const planState = getPlanState(plan || null, latestMigration, migrationsQuery);
   const isShowingPrecopyView =
     !!plan?.spec.warm &&
-    (warmPlanState === 'Starting' ||
-      warmPlanState === 'Copying' ||
-      warmPlanState === 'AbortedCopying');
+    (planState === 'Starting' || planState === 'Copying' || planState === 'AbortedCopying');
```
mturley, line 108: I think we need to replace `planState === 'AbortedCopying'` with `planState === 'CanceledCopying' || planState === 'FailedCopying'`

src/app/common/constants.ts
```diff
@@ -38,22 +38,21 @@ export enum StatusCategoryType {
...
-}
+export type ConditionType = 'Ready' | 'Executing' | 'Succeeded' | 'Failed' | 'Canceled';
+
+export type ColdPlanState = Exclude<PlanState, 'Copying' | 'AbortedCopying' | 'StartingCutover'>;
+
+export type PlanState =
+  | 'NotStarted' // Just created, no action taken yet
+  | 'Starting' // User clicked start button, Migration CR exists but no status data exists yet
+  | 'Copying' // Warm-specific. "Running incremental copies" in the UI, the first stage of a warm migration
+  | 'AbortedCopying' // Warm-specific. Plan was canceled during copying
```
mturley, line 49: I think we need to rename `AbortedCopying` to `FailedCopying` and add a new state `CanceledCopying`

mturley (changes requested):
As discussed on Slack, I think these are the changes necessary to distinguish between Canceled and Failed during the copying phase of a warm plan.

src/app/common/constants.ts
seanforyou23: Making all these states explicit seems like a great improvement, thanks for all the help!

mturley (approved):
LGTM! Since we're not backporting it we'll have plenty of opportunities to continue testing it, including resolving https://github.com/konveyor/forklift-ui/issues/694. I think it's good to merge.

PR #680, opened 2021-07-02
Bug 1977657 - Concatenate Service Signing CA to global CA bundle

fdupont-redhat:
In OpenShift 4.8, the `service-ca.crt` file contains only one CA certificate, the one for Service Serving Certificates, and this breaks the ability for NodeJS to verify Kubernetes API certificate. Previously, all the internal CA certificates were present in `service-ca.crt`. Now, they are only present in `ca.crt`.

This pull request creates an entrypoint script that concatenates the `service-ca.crt` to the `ca.crt` bundle, if it exists. A parallel pull request, https://github.com/konveyor/forklift-operator/pull/131, changes the `NODE_EXTRA_CA_CERTS` environment variable of the UI container to point to `ca.crt`, so that it uses the CA bundle with all the internal CAs and the Service Service CA.


fbladilo (approved):
LGTM , glad this one turned to be not so complicated.

PR #684, opened 2021-07-02
Bug 1978785: Use history.replace() instead of history.push() when redirecting on first load of /providers route

### LINKED ISSUE #543: [BZ#1978785] Navigating to `/providers` route redirects to `/providers/:providerType` as a `history.push()`, should be `history.replace()` so that the back button works
If you click the Providers nav link, you get two entries pushed into your browser history (first the root providers page, then a redirect to the tab for a specific provider type). Clicking Back in the browser takes you back to the root providers page which then redirects you forward again, there's no way to go back two levels.

If we change that `history.push` to a `history.replace` on this line: https://github.com/konveyor/forklift-ui/blob/main/src/app/Providers/ProvidersPage.tsx#L80 the root providers page will not end up in the history, and the Back button will work as expected.

https://bugzilla.redhat.com/show_bug.cgi?id=1978785

mturley:
Fixes #543 (https://bugzilla.redhat.com/show_bug.cgi?id=1978785)

### DIFF
--- a/src/app/Providers/ProvidersPage.tsx
+++ b/src/app/Providers/ProvidersPage.tsx
@@ -77,7 +77,7 @@ const ProvidersPage: React.FunctionComponent = () => {
       (!activeProviderType && availableProviderTypes.length > 0) ||
       (activeProviderType && !availableProviderTypes.includes(activeProviderType))
     )
-      history.push(`/providers/${availableProviderTypes[0]}`);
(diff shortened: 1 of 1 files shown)


seanforyou23 (approved):
Nice!

fdupont-redhat (approved):
LGTM

PR #685, opened 2021-07-02
[Backport] Use history.replace() instead of history.push() when redirecting on first load of /providers route (#684)

mturley:
Backports #684 (https://bugzilla.redhat.com/show_bug.cgi?id=1978785)


fdupont-redhat (approved):
LGTM

PR #686, opened 2021-07-05
Use nodejs-14-minimal as base image

fdupont-redhat:
The `ubi8/nodejs-14-minimal` container image only weighs 253 MB, against 659 MB for the `ubi8/nodejs-14` image. With this pull request, we reduce the total size of the `forklift-ui` container image, hence reduce download time and save space on the nodes where the container runs.

I have built the image and tested it in a live cluster. No error, after a few tests.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ COPY . .
 RUN npm install && npm run build
 
 # Runner image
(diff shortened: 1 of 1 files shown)


mturley (approved):
LGTM

PR #691, opened 2021-07-07
Aligned tests with changes in MTV 2.1, fixed VMWare provider, mapping…

ibragins:
Aligned tests with changes in MTV 2.1, fixed VMWare provider, mappings and plan creation

PR #696, opened 2021-07-08
[Backport] Aligned tests with changes in MTV 2.1, fixed VMWare provider, mappings and plan creation (#691)

mturley:
Backports #691 

@ibragins let's open backport PRs for all 2.1 related tests to the `release-v2.1.0` branch from here on out so we don't have to do something like https://github.com/konveyor/forklift-ui/pull/695 again later. Here is a backport for the only outstanding one so far.

### DIFF
--- a/qe-tests/cypress.json
+++ b/qe-tests/cypress.json
@@ -1,4 +1,4 @@
 {
-  "viewportWidth": 1280,
-  "viewportHeight": 800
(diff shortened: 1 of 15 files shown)


mturley:
Self-merging because this is existing qe-tests code and it is not modifying any source code used in the UI.

PR #697, opened 2021-07-09
Bug 1980585: PlanDetailsModal: look up sourceProvider of any type instead of just vsphere

mturley:
https://bugzilla.redhat.com/show_bug.cgi?id=1980585


mturley:
Oops, pushed to main branch in my fork instead of named branch

PR #698, opened 2021-07-09
Bug 1980585: PlanDetailsModal: look up sourceProvider of any type instead of just vsphere

mturley:
https://bugzilla.redhat.com/show_bug.cgi?id=1980585

Replaces #697 which was using the wrong branch


seanforyou23 (approved):
LGTM

Issue #701, opened 2021-07-12
Can we improve the existing route guard UI?


seanforyou23:
Currently, if you're in the plan creation wizard flow and you attempt to navigate to another route/page, a native browser prompt asks to confirm before proceeding.

<img src="https://user-images.githubusercontent.com/5942899/125327937-b543ca00-e311-11eb-8bc3-2cbdb5d42f7d.png" width="400" alt="screenshot" />

It would be nice if we had a custom implementation for this so that the styles match that of the rest of the UI.

mturley:
Note to self: even after #702, the route guard does not stop the user from closing or reloading the tab, since those operations circumvent react-router. We could address that in a future PR with clever use of the `beforeunload` event. I may follow up after this is closed to try that out.

PR #702, opened 2021-07-12
Add a prompt component for route blocking

seanforyou23:
This PR replaces our usage of react-router-dom's `Prompt` component with a custom one that recycles our existing `ConfirmModal`. Should help close https://github.com/konveyor/forklift-ui/issues/701

screenshot

### DIFF
--- a/src/app/Plans/components/Wizard/PlanWizard.tsx
+++ b/src/app/Plans/components/Wizard/PlanWizard.tsx
@@ -11,11 +11,11 @@ import {
   WizardStep,
   WizardStepFunctionType,
 } from '@patternfly/react-core';
-import { Link, Prompt, Redirect, useHistory, useRouteMatch } from 'react-router-dom';
+import { Link, Redirect, useHistory, useRouteMatch } from 'react-router-dom';
 import { UseQueryResult } from 'react-query';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import { useFormField, useFormState } from '@konveyor/lib-ui';
-
+import { RouteGuard } from '@app/common/components/RouteGuard';
 import WizardStepContainer from './WizardStepContainer';
 import GeneralForm from './GeneralForm';
 import FilterVMsForm from './FilterVMsForm';
@@ -410,9 +410,10 @@ const PlanWizard: React.FunctionComponent = () => {
         <Redirect to="/plans" />
       ) : (
         <>
-          <Prompt
+          <RouteGuard
             when={forms.isSomeFormDirty && mutationStatus === 'idle'}
-            message="Leave this page? All unsaved changes will be lost."
+            title="Leave this page?"
+            message="All unsaved changes will be lost."
           />
           <PageSection
             title={`${!planBeingEdited ? 'Create' : 'Edit'} Migration Plan`}
--- a/src/app/common/components/RouteGuard.tsx
+++ b/src/app/common/components/RouteGuard.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { useHistory } from 'react-router';
 import ConfirmModal from '@app/common/components/ConfirmModal';
-import { RouteGuardOptions } from '@app/common/constants';
+import { unBlockRoute } from '@app/common/constants';
 interface IRouteGuard {
   when: boolean;
   title: string;
@@ -20,25 +20,24 @@ const RouteGuard: React.FunctionComponent<IRouteGuard> = ({
   const history = useHistory();
   const [showRouteGuard, setShowRouteGuard] = React.useState(when);
   const [currentPath, setCurrentPath] = React.useState('');
-  const unBlock = () => RouteGuardOptions.permit;
 
   React.useEffect(() => {
     if (when) {
       history.block((prompt) => {
         setCurrentPath(prompt.pathname);
         setShowRouteGuard(true);
-        return RouteGuardOptions.prevent;
+        return false;
       });
     } else {
-      history.block(unBlock());
+      history.block(unBlockRoute);
     }
     return () => {
-      history.block(unBlock());
+      history.block(unBlockRoute);
     };
   }, [history, when]);
 
   const handleOk = React.useCallback(() => {
-    history.block(unBlock());
+    history.block(unBlockRoute);
     history.push(currentPath);
   }, [currentPath, history]);
 
--- a/src/app/common/components/RouteGuard.tsx
+++ b/src/app/common/components/RouteGuard.tsx
@@ -1,0 +1,60 @@
+import * as React from 'react';
+import { useHistory } from 'react-router';
+import ConfirmModal from '@app/common/components/ConfirmModal';
+import { RouteGuardOptions } from '@app/common/constants';
+interface IRouteGuard {
+  when: boolean;
+  title: string;
+  message: string | JSX.Element;
+  okText?: string;
+  cancelText?: string;
+}
+
+const RouteGuard: React.FunctionComponent<IRouteGuard> = ({
+  when,
+  title,
+  message,
+  okText = 'OK',
+  cancelText = 'Cancel',
+}) => {
+  const history = useHistory();
+  const [showRouteGuard, setShowRouteGuard] = React.useState(when);
+  const [currentPath, setCurrentPath] = React.useState('');
+  const unBlock = () => RouteGuardOptions.permit;
+
+  React.useEffect(() => {
+    if (when) {
+      history.block((prompt) => {
+        setCurrentPath(prompt.pathname);
+        setShowRouteGuard(true);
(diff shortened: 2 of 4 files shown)


src/app/Plans/components/Wizard/PlanWizard.tsx
```diff
@@ -11,11 +11,11 @@ import {
   WizardStep,
   WizardStepFunctionType,
 } from '@patternfly/react-core';
-import { Link, Prompt, Redirect, useHistory, useRouteMatch } from 'react-router-dom';
+import { Link, Redirect, useHistory, useRouteMatch } from 'react-router-dom';
 import { UseQueryResult } from 'react-query';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import { useFormField, useFormState } from '@konveyor/lib-ui';
-
+import { Prompt } from '@app/common/components/Prompt';
```
mturley, line 18: Since this is not coming from react-router, do you think "Prompt" is a descriptive enough name? With no knowledge of the component I wouldn't guess it has anything to do with navigation.

src/app/common/components/Prompt.tsx
```diff
@@ -0,0 +1,58 @@
+import * as React from 'react';
+import { Modal, Button } from '@patternfly/react-core';
```
mturley, line 2: Can you remove these unused imports?

src/app/common/components/Prompt.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  React.useEffect(() => {
+    if (when) {
+      history.block((prompt) => {
+        setCurrentPath(prompt.pathname);
+        setShowPrompt(true);
+        return 'true';
+      });
+    }
+  }, [history, when]);
```
mturley, line 34: This `useEffect` successfully sets up the block when `when` becomes true, but if `when` becomes false again it doesn't remove the block. That doesn't matter much for this use case, but the prior behavior is that you're only blocked if some form is dirty (e.g. I can open a plan for editing, step through it without changing anything, and close it with no warning). I think if this will be reusable in lib-ui it should also clean up this block if `when` becomes false.

Looking at [this stackoverflow answer](https://stackoverflow.com/a/65527222) it seems like the cleanest way to do that might be to have the effect always call `history.block` with `if (when)` inside its callback, and then have the effect return a cleanup function that calls `unblock` (stored from the return value of `history.block()`). That way the block is always in place as long as the component is mounted but it checks the `when` condition every time it is hit.

src/app/common/components/Prompt.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const handleOk = React.useCallback(() => {
+    history.block(noop);
```
mturley, line 37: Maybe if we store the `unblock` function from above in a `useRef`, we can just call it here instead of having to pass a noop to clear the block which doesn't look like documented behavior as far as I can tell from the docs.

src/app/index.tsx
```diff
@@ -20,7 +21,7 @@ const App: React.FunctionComponent = () => (
     <PollingContextProvider>
       <LocalStorageContextProvider>
         <NetworkContextProvider>
-          <Router>
+          <Router getUserConfirmation={noop}>
```
mturley, line 24: This seems odd, do we need this to return a default message if/when a native alert is used? (see my general comment)

mturley (changes requested):
Hey Mike. A couple of small things here. Also, one big observation: this successfully blocks navigation within the app, but it does not stop the user from closing or reloading the tab with unsaved changes. I was about to say the prior implementation did this properly, but in testing it now I guess it doesn't! I could have sworn it did originally. Maybe that broke with a react-router upgrade or something.

Up to you if you want to try and fix that here or open a separate issue for it. I noticed [a caveat about it in their docs here](https://github.com/ReactTraining/history/blob/master/docs/blocking-transitions.md#caveats).

src/app/Plans/components/Wizard/PlanWizard.tsx
seanforyou23: Probably not, I didn't think too much into the name so far - was mostly just working toward a PF style direct replacement. How about a generic `RouteGuard`?

mturley: I like RouteGuard!

src/app/common/components/Prompt.tsx
seanforyou23: Ok, yea that makes sense. I was trying to find more info on how to best utilize that return function, and had actually gone down that road a bit but I wasn't quite sure so I left it minimal for that first pass. I'll tidy this up  👍

src/app/index.tsx
mturley: Nevermind, I see that this prevents history.block from defaulting to a native confirm dialog.

src/app/common/components/Prompt.tsx
seanforyou23: Ok I pushed some updates, what's strange is that the block function doesn't actually return anything usable as mentioned in the docs. For now I stored a similar noop style function within the component scope that serves the same purpose and is used in the same way.

src/app/index.tsx
seanforyou23: Yea took me a few to grok what was happening here. Turns out whatever the callback to block() returns is provided here so I suppose if you need to do further processing to determine if the route should be blocked or not, you have the ability to do that in a more global manner. I set up an enum that we can use to control the passage - but it's not required. By setting a function here, we're overriding the default behavior of react-router block implementation, so we need to at least set something, even if it's just a noop like it was before. I left it more verbose just for discussion sake so people could see it in action and in case it ties in with preventing the tab close as you mentioned above.

seanforyou23:
@mturley pushed another update - now the old Prompt from react-router-dom should still work in addition to the new RouteGuard.. After thinking about it more it seemed weird that the old solution wouldn't work anymore, so hopefully this makes sense. Let me know what you think!

src/app/common/components/RouteGuard.tsx
```diff
@@ -0,0 +1,60 @@
...
+  when,
+  title,
+  message,
+  okText = 'OK',
+  cancelText = 'Cancel',
+}) => {
+  const history = useHistory();
+  const [showRouteGuard, setShowRouteGuard] = React.useState(when);
+  const [currentPath, setCurrentPath] = React.useState('');
+  const unBlock = () => RouteGuardOptions.permit;
```
mturley, line 23: Why make unBlock a function? you could just pass RouteGuardOptions.permit in the places you call it, or make it a regular variable here if you don't want to repeat that

src/app/index.tsx
```diff
@@ -20,7 +21,19 @@ const App: React.FunctionComponent = () => (
...
-          <Router>
+          <Router
+            getUserConfirmation={(message: string, callback: (ok: boolean) => void) => {
+              const isAllowed = message === RouteGuardOptions.permit || message === '';
+              if (
+                message !== RouteGuardOptions.prevent &&
+                message !== RouteGuardOptions.permit &&
+                message !== ''
+              ) {
+                confirm(message) ? callback(true) : callback(false);
```
mturley, line 32: Ah, is this logic just here so that we could use the regular <Prompt> if we really want to? If we make RouteGuard into a reusable component in lib-ui or PF, does the consumer need to make sure they put this code in their router as well?

mturley (commented):
@seanforyou23 this is looking great, I just have one super tiny nitpick and a question.

src/app/common/components/RouteGuard.tsx
seanforyou23: Good call, doesn't need to be a function :)

src/app/index.tsx
seanforyou23: Yes exactly - it could be cleaned up if we don't care to have support for `Prompt` from react-router-dom (I don't know why we'd ever use the old version given it doesn't automatically prevent browser tab close like you mentioned). I'd personally prefer to remove that old support and just require devs to always use the newer version, but I'm happy to implement whatever everyone thinks is best.

It does look like any consumer will have to modify this `getUserConfirmation` prop in some way in order to customize the prompt behavior. If I can find a way to avoid it I surely will, but as of this moment it looks like some type of modification will be required.

seanforyou23:
@mturley I pushed another round of updates. With the latest changes you can see that we don't really need an enum at all, and simply returning something from history.block prevents the route change. I was thinking prevent/permit strings would be more readable, but it has the trade off of slightly more code/setup. I removed this and distilled it down to about as simple as I think it can be in case it makes it easier to contribute to PF or lib-ui.. let me know what you think!

mturley (approved):
I think you've done it! This looks great to me.

mturley:
@seanforyou23 I'm marking this `v.2.1.0/no`, do you agree?

Issue #703, opened 2021-07-12
[BZ#1980587] Filtering plans by 'Ready' status returns 'Succeeded' plans [NEEDINFO]


seanforyou23:
Filtering plans by 'Ready' status returns 'Succeeded' plans  and 'Ready' plans.
See attached screen shot.

Actual results:
---------------
Filtering plans by 'Ready' status returns 'Succeeded' plans  and 'Ready' plans

Expected results:
-----------------
Filtering plans by 'Ready' status should only return 'Ready' plans.

![migration_plan_type](https://user-images.githubusercontent.com/5942899/125332761-6c8f0f80-e317-11eb-9a5c-26d9cf0ea2bf.png)

https://bugzilla.redhat.com/show_bug.cgi?id=1980587

PR #707, opened 2021-07-14
Bug 1980587: Update plans table status filter logic to account for various plan states

seanforyou23:
The options in the plans table status filter dropdown were out of date, so this PR brings those options back in alignment with the states known for a plan (driven by the same logic that drives the plan status column in the table). Other than fixing the existing filter logic, another thing this does is introduce a few new statuses the user can filter by. Previously it was just Failed, Ready, Running, and Succeeded. Now we have addition options of Not Ready, Canceled, and Finished Incomplete. @vconzola are these new states ok/provide value? Let me know if we need to tweak it further.

Screen Shot 2021-07-14 at 3 40 52 PM

drive status filter options from same state log as a plan state
distinguish between not started because of issues vs simply not having tried to start yet
should help close https://github.com/konveyor/forklift-ui/issues/703

### DIFF
--- a/src/app/Plans/components/PlanDetailsModal.tsx
+++ b/src/app/Plans/components/PlanDetailsModal.tsx
@@ -3,7 +3,6 @@ import * as React from 'react';
 import {
   IHook,
   IMetaObjectMeta,
-  InventoryProvider,
   IPlan,
   MappingType,
   SourceInventoryProvider,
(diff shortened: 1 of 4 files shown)


vconzola (commented):
@seanforyou23 Yes, if a state can show up in the Status column user should be able to filter by it.

mturley (approved):
LGTM! This is a really nice loose end to tie up after the PlanState refactor, that stuff is looking real slick now. 👌

seanforyou23:
Thx for all the help @mturley!

Issue #706, opened 2021-07-14
Iterate on confirmation modal text


mturley:
See post-merge review comments on https://github.com/konveyor/forklift-ui/pull/700. We can probably backport as part of http://bugzilla.redhat.com/show_bug.cgi?id=1981801.

PR #708, opened 2021-07-14
chore(copy): update confirmation microcopy to match best practices

seanforyou23:
This PR applies some further cleanup to the language used in various confirmation modals throughout the app.

Screen Shot 2021-07-14 at 4 16 54 PM

Should help close https://github.com/konveyor/forklift-ui/issues/706


mturley:
@seanforyou23 as part of this PR would you mind updating all the `errorTitle` / `errorTitles` strings across our usages of `ResolvedQuery` / `ResolvedQueries` to match? They are all also in that format e.g. "Error loading providers" and should probably all say e.g. "Could not load providers"

mturley (approved):
LGTM!

PR #715, opened 2021-07-21
RouteGuard: fix console warning, block tab from being closed or reloaded

### LINKED ISSUE #714: Enhance RouteGuard to block closing or reloading the browser tab, and fix warning about more than one history prompt at a time
Following up on https://github.com/konveyor/forklift-ui/issues/701#issuecomment-883648905, I think we can carefully introduce a `beforeunload` event handler to RouteGuard to make it act the same way even if the user is circumventing react-router via closing or reloading the tab.

Also, I noticed after merging #701 that we get this warning when the `history.block()` function is called multiple times:
Screen Shot 2021-07-21 at 3 21 34 PM

mturley:
Resolves #714.

* The `Warning: A history supports only one prompt at a time` message was coming up because `history.block` was being called more than once without properly unblocking it. Looking at [the docs for `history.block`](https://github.com/ReactTraining/history/blob/master/docs/blocking-transitions.md), it seems the correct way to unblock the route is to call the callback that gets returned from the initial call to `block` rather than calling `block` again with a noop function. The first commit in this PR accomplishes this by storing that returned callback in a ref and calling it where necessary via the ref.
* In order to block the tab from being reloaded or closed while we have our route guard in place, we need to use the `beforeunload` event. Unfortunately there is no way to prevent the unload without showing the native confirmation dialog provided by the browser, so we can't show our custom ConfirmModal in these corner cases. The second commit in this PR implements this event handler.

There was also an `} else { unblock(); }` case in the useEffect that we didn't need. useEffects call their cleanup function every time before running the effect again, so each time it runs the route will already be unblocked. So if `when` is false when the effect runs again it is sufficient to just not call `history.block`. The only places where we need to unblock are in that useEffect cleanup function and the `handleOk` function.

cc @seanforyou23

### DIFF
--- a/src/app/common/components/RouteGuard.tsx
+++ b/src/app/common/components/RouteGuard.tsx
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import { useHistory } from 'react-router';
+import { UnregisterCallback } from 'history';
 import ConfirmModal from '@app/common/components/ConfirmModal';
-import { unBlockRoute } from '@app/common/constants';
+
 interface IRouteGuard {
   when: boolean;
   title: string;
@@ -21,23 +22,27 @@ const RouteGuard: React.FunctionComponent<IRouteGuard> = ({
   const [showRouteGuard, setShowRouteGuard] = React.useState(when);
   const [currentPath, setCurrentPath] = React.useState('');
 
+  const unblockFnRef = React.useRef<UnregisterCallback | null>(null);
+  const unblock = () => {
+    if (unblockFnRef.current) {
+      unblockFnRef.current();
+      unblockFnRef.current = null;
+    }
+  };
+
   React.useEffect(() => {
     if (when) {
-      history.block((prompt) => {
+      unblockFnRef.current = history.block((prompt) => {
         setCurrentPath(prompt.pathname);
         setShowRouteGuard(true);
         return false;
       });
-    } else {
-      history.block(unBlockRoute);
     }
-    return () => {
-      history.block(unBlockRoute);
-    };
+    return unblock;
   }, [history, when]);
 
   const handleOk = React.useCallback(() => {
-    history.block(unBlockRoute);
+    unblock();
     history.push(currentPath);
   }, [currentPath, history]);
 
--- a/src/app/common/components/RouteGuard.tsx
+++ b/src/app/common/components/RouteGuard.tsx
@@ -3,6 +3,13 @@ import { useHistory } from 'react-router';
 import { UnregisterCallback } from 'history';
 import ConfirmModal from '@app/common/components/ConfirmModal';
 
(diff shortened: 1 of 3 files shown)


mturley:
Looking closer at the docs for history, I'm actually a little confused... [under `caveats` there](https://github.com/ReactTraining/history/blob/master/docs/blocking-transitions.md#caveats), it says that the `history.block` function registers its own `beforeunload` handler... but I didn't observe that behavior. Maybe we need to update our version of `history`, and then the handler introduced in this PR is unnecessary?

src/app/common/components/RouteGuard.tsx
```diff
@@ -21,23 +29,33 @@ const RouteGuard: React.FunctionComponent<IRouteGuard> = ({
...
+      unblockFnRef.current();
+      unblockFnRef.current = null;
+    }
+  };
+
   React.useEffect(() => {
     if (when) {
-      history.block((prompt) => {
+      // When browser history routing happens (clicking links, back/forward) we can handle it with our custom modal.
+      unblockFnRef.current = history.block((prompt) => {
```
seanforyou23, line 43: Can you help me understand why storing this return value in a ref works, but setting a state property with the same callback doesn't? I'm sure it's just a gap in my understanding of the lifecycle differences between useRef and useState.. I was essentially trying..

```
const [unblockFn, setUnblockFn] = React.useState<UnregisterCallback | null>(null);
...
    if (unblockFn) {
      unblockFn();
      setUnblockFn(null);
    }
...
const unBlockHandler = history.block((prompt) => {...
setUnblockFn(unBlockHandler);
```

But unblockFn is always null in that case.

seanforyou23 (approved):
This looks great! Had a question about the blocking implementation, but end result is exactly what we need. Nice job

src/app/common/components/RouteGuard.tsx
mturley: I was also trying that and getting frustrated. Something dumb about closures and function scope. I think it's because the useEffect cleanup function has to reference `unblockFn` from the function scope where the cleanup function was declared, which is the scope from the render before `setUnblockFn` was called. Since you're using `unblockFn` inside the useEffect it is in the dependencies for the effect, which means when you `setUnblockFn` it will cause a re-render with a new `unblockFn`, so the effect will cleanup before that render, before it has access to the new `unblockFn`. 

Anyway, since we don't need to render any JSX differently based on the unblockFn, putting it in a useState is sorta overkill. It's the same reason you put return values of setTimeout / setInterval in refs, they need to sort of exist as a persistent thing outside the render lifecycle.

My brain hurts.

seanforyou23: Thx for explaining! I was speculating something of the sort, just wasn't so confident so wanted to get your take on it. Nicely done, sir.

PR #719, opened 2021-07-23
Bug 1985475: SelectVMsForm: Ensure the latest versions of already-selected VMs are used when re-rendering from new query data

mturley:
When the SelectVMsForm is mounted, if there are already any selected VMs in form state (because either you are editing an existing plan, or you have moved back and forward again in the wizard), we include those VMs in the list even if they don't match the tree-based filter. This is done by capturing the copy of `selectedVMs` from when the component is first mounted in a ref and including them when calling `getAvailableVMs` (as opposed to always including the latest selected VMs, because that way if you deselect them they can suddenly disappear from the list... it's a whole thing).

The bug with this implementation is that when new VM data comes in via polling, we were not showing the latest data for these extra VMs being included this way since we were using the copy of them from the first mount. This PR fixes it by using selfLinks as the source of truth for which VMs to include, and always getting the latest VMs from the query data for all of the selfLinks.

I hate these two wizard steps.

### DIFF
--- a/src/app/Plans/components/Wizard/VMConcernsDescription.tsx
+++ b/src/app/Plans/components/Wizard/VMConcernsDescription.tsx
@@ -33,7 +33,7 @@ const VMConcernsDescription: React.FunctionComponent<IVMConcernsDescriptionProps
   if (vm.revisionValidated < vm.revision) {
     return (
       <TextContent className={spacing.myMd}>
-        <Text component="p">Completing migration Analysis. This might take a few minutes.</Text>
+        <Text component="p">Completing migration analysis. This might take a few minutes.</Text>
       </TextContent>
     );
   } else {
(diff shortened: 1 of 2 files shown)


vconzola (commented):
@mturley "I hate these two wizard steps." LOL! Me too. Since we're not really pressed right now should we revisit the single step, tree table for VM selection?

mturley:
@vconzola I was discussing that with Mike S, and the tree table is quite a huge can of worms that will raise a bunch of other corner cases. He and I are both thinking a tree table is a bad idea here, we can discuss why if you want some time.

mturley:
That said, I don't have much of a suggestion about doing it better 🙃 my comment about hating them was mostly lamenting the trouble this code has been. but it's in great shape now.

seanforyou23 (approved):

vconzola:
> That said, I don't have much of a suggestion about doing it better 🙃 my comment about hating them was mostly lamenting the trouble this code has been. but it's in great shape now.

Alright. If the code is rock solid now, I'll stop bugging you about the tree table. ;-)

mturley:
No worries! I had started thinking it was a good idea and Mike scared me straight 😆

PR #720, opened 2021-07-23
[Backport] Bug 1985475: SelectVMsForm: Ensure the latest versions of already-selected VMs are used when re-rendering from new query data

mturley:
Backports #719.

Note: due to conflicts #675 (this code was totally reworked), this is a totally different (but similar) branch-specific implementation of the fix.


mturley:
Closing this in favor of backporting #675 to prevent conflicts in backporting #719.

Issue #723, opened 2021-07-23
Regression: Creating a new provider tab (adding the first provider of that type) no longer navigates the user to that tab


mturley:
In https://github.com/konveyor/forklift-ui/pull/513 we added a callback after adding a provider that will navigate you to the tab containing the provider you just added. This works as expected in 2.0.0.

However, in https://github.com/konveyor/forklift-ui/pull/654 we fixed a bug where removing the last provider of a type would leave you at an invalid URL since the tab no longer exists after that removal (because our provider tabs are dynamic and only exist if there are providers of that type present). As a result, if you navigate to a route for a provider type that doesn't exist, you will be redirected to the first tab.

This breaks the #513 behavior because when you add a provider, it navigates to that tab before the providers query catches up and includes the new provider, so the tab doesn't quite exist yet and you get redirected away.

To fix this, we need to either revert #654 and have some kind of empty/404 page for a blank providers tab where the URL won't change (so when the new provider shows up the page will fix itself), or we need to do something clever where we don't navigate to that tab until the new provider is actually in the query cache (possibly from both the cluster and inventory APIs?).

If we go with the 404-style approach, it would be weird to show the tabs since none of them would be active, so we may want to render a full-page empty state instead if you're on a route that doesn't have any providers. It should have a button that takes you to the default providers tab in this case, so it is still a good experience when you delete the last provider of a type (#513).

mturley:
@seanforyou23 if you're looking for a fun little one this might be good

PR #729, opened 2021-07-28
fix(723): ensure user is navigated to new provider tab on adding a new provider

seanforyou23:
This PR updates the logic around when a first or last provider is added/deleted to ensure the user is always sent to a correct provider tab.

- Adding a new provider always switches you to that provider tab
- Navigating to an unknown provider (for example /providers/foo) should redirect you to `/providers` (or `/providers/availableProviderType` if one exist)
- Removing the last provider of a given type should also redirect you to `/providers` (or `/providers/availableProviderType` if one exist)

Should help close https://github.com/konveyor/forklift-ui/issues/723

### DIFF
--- a/src/app/Providers/ProvidersPage.tsx
+++ b/src/app/Providers/ProvidersPage.tsx
@@ -134,7 +134,12 @@ const ProvidersPage: React.FunctionComponent = () => {
         <ResolvedQueries results={allQueries} errorTitles={allErrorTitles} errorsInline={false}>
           <Card>
             <CardBody>
-              {!clusterProvidersQuery.data || !inventoryProvidersQuery.data || areProvidersEmpty ? (
+              {!clusterProvidersQuery.data ||
+              !inventoryProvidersQuery.data ||
+              areProvidersEmpty ||
+              !clusterProvidersQuery.data?.items
+                .map((provider) => provider.spec.type)
+                .includes(activeProviderType) ? (
                 <EmptyState className={spacing.my_2xl}>
                   <EmptyStateIcon icon={PlusCircleIcon} />
                   <Title headingLevel="h2" size="lg">
--- a/src/app/Providers/ProvidersPage.tsx
+++ b/src/app/Providers/ProvidersPage.tsx
@@ -18,7 +18,7 @@ import { PlusCircleIcon } from '@patternfly/react-icons';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import { useHistory } from 'react-router-dom';
 
-import { ProviderType, PROVIDER_TYPE_NAMES } from '@app/common/constants';
+import { ProviderType, PROVIDER_TYPE_NAMES, PROVIDER_TYPES } from '@app/common/constants';
 import { useClusterProvidersQuery, useInventoryProvidersQuery, usePlansQuery } from '@app/queries';
 
 import ProvidersTable from './components/ProvidersTable';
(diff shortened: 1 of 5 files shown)


src/app/Providers/ProvidersPage.tsx
```diff
@@ -72,13 +72,13 @@ const ProvidersPage: React.FunctionComponent = () => {
...
   const activeProviderType = match?.params.providerType || null;
 
+  const isValidProviderType = !!(activeProviderType && PROVIDER_TYPES.includes(activeProviderType));
+
   React.useEffect(() => {
-    if (
-      (!activeProviderType && availableProviderTypes.length > 0) ||
-      (activeProviderType && !availableProviderTypes.includes(activeProviderType))
-    )
+    if ((!isValidProviderType || !activeProviderType) && availableProviderTypes.length > 0) {
```
mturley, line 78: @seanforyou23 I thought of one more weird corner case here that I would be ok with not handling: If you go to the URL for e.g. `/providers/openshift` when there are no openshift providers, that is a valid provider type so it will still render the page (with a blank body). The tab won't exist, so users could only get there if they like bookmarked it or something and then removed a provider. If we want to handle this we should maybe render a real empty state on that page instead of a blank body in that case.

I think I'm really splitting hairs here though, so I'm totally fine leaving that alone.

mturley (approved):
Looking good!! I'll wait to merge until you have a chance to look at my comment here though.

src/app/Providers/ProvidersPage.tsx
seanforyou23: Ah, it's a good point! Yea that should be a nicely formatted page for sure. I don't mind doing that as part of this - I'll take a best guess at what the empty state should say. @vconzola any preferences?

vconzola:
@seanforyou23 there should already be an empty state design for VMware providers. just riff off that.

mturley:
@vconzola I think when we decided to make provider tabs only exist if you've added a provider of that type, we made it impossible to hit the empty state on this page since there will always be at least one openshift provider (the host provider). It is in the code though @seanforyou23 in `ProvidersPage`, the body says "Add source and target providers for migrations."

I think we need to just alter the boolean expression for that empty state so it also appears if there are no providers of the `activeProviderType` present in the `clusterProvidersQuery.data.items`.

mturley:
@seanforyou23 I think you pushed https://github.com/konveyor/forklift-ui/pull/729/commits/2f1d7af94683f76550249c4e02306e91c75f51f8 to the wrong branch :)

seanforyou23:
@mturley nope - it was intentional 😉 - the backend service was down for a bit when I went back to address the follow up comments, and I noticed the Alerts that are shown where this component would normally render don't follow best practices so just doctored that up while I was in there. I'll update the PR description to explain the changes once I'm ready for another review. Gonna circle back to the rest of remaining changes soon.. thx for being patient :)

mturley:
sounds good! sorry to assume.

seanforyou23:
Hey @mturley - I'm tied up at the moment working on another issue, I think this PR is in good shape for a merge in case you wanna go and head get it in? I'm definitely planning on fixing the empty state issue next, will send it as a separate PR in case you wanna get this in just to keep it moving along. Your call, happy Friday :D

mturley:
Thanks! Actually I think there's no rush in getting this in, we can wait until you're ready.

src/app/Providers/ProvidersPage.tsx
seanforyou23: https://github.com/konveyor/forklift-ui/pull/729/commits/4ea3d7dd3bd1a6dca52c5a09c302cf3d3a458bd7

Screen Shot 2021-08-02 at 9 42 16 AM

mturley (approved):
Good stuff!

PR #730, opened 2021-07-29
Centralize PR-BZ map

rayfordj:
https://github.com/konveyor/bz-github-action/issues/10

This will need to be picked to any active release branches, such as `release-v2.1.0`.

PR #731, opened 2021-07-29
[Backport] Centralize PR-BZ map (#730)

mturley:
Backports #730. BZ automation only, does not affect builds.


rayfordj (approved):
/lgtm

Thanks, @mturley !

PR #732, opened 2021-07-30
Block the plan name 'create' to prevent a URL path conflict

mturley:
Here's a dumb one. In our UI if you navigate to `/plans/create` you get the plan wizard, and if you navigate to `/plans/myPlan` you get the migration details for "myPlan". I only just realized that this means if you create a plan with the name "create", you can never view its migration details because the wizard route matches first.

There's probably a better way to fix this, but just for now I'm blocking the name "create" in the plan wizard.

Screen Shot 2021-07-30 at 11 23 37 AM


seanforyou23 (approved):
Good catch @mturley!

PR #733, opened 2021-07-30
Bug 1985787: Fix blank columns in Select VMs step for RHV VMs

### LINKED ISSUE #724: [BZ#1985787] Select VMs step: For RHV VMs, the Datacenter and Host columns are blank
https://bugzilla.redhat.com/show_bug.cgi?id=1985787

RHV-VMs-missing-columns

mturley:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1985787.

Resolves #724

* In the RHV/oVirt tree data, the `Datacenter` kind is spelled differently: `DataCenter` with a capital C. Our `getVMTreePathInfo` helper was not accounting for this, so that column was showing up empty for RHV VMs. This PR adds a condition for that in `getVMTreePathInfo` and updates the mock data to match.
* RHV VMs that are powered off are not associated with a host, so that column was blank for those VMs. Instead of leaving it blank, this PR changes it to say 'N/A' so as to be less broken-looking (based on discussion with @vconzola).

### DIFF
--- a/src/app/Plans/components/Wizard/helpers.tsx
+++ b/src/app/Plans/components/Wizard/helpers.tsx
@@ -308,7 +308,9 @@ export const getVMTreePathInfo = (
         .map((node) => node.object) as ICommonTreeObject[]) || null;
   }
   return {
-    datacenter: hostTreeAncestors?.find((node) => node.kind === 'Datacenter')?.object || null,
+    datacenter:
+      hostTreeAncestors?.find((node) => node.kind === 'Datacenter' || node.kind === 'DataCenter')
+        ?.object || null,
     cluster: hostTreeAncestors?.find((node) => node.kind === 'Cluster')?.object || null,
     host: hostTreeAncestors?.find((node) => node.kind === 'Host')?.object || null,
     folders,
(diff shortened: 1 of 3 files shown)


src/app/Plans/components/Wizard/helpers.tsx
```diff
@@ -308,7 +308,9 @@ export const getVMTreePathInfo = (
         .map((node) => node.object) as ICommonTreeObject[]) || null;
   }
   return {
-    datacenter: hostTreeAncestors?.find((node) => node.kind === 'Datacenter')?.object || null,
+    datacenter:
+      hostTreeAncestors?.find((node) => node.kind === 'Datacenter' || node.kind === 'DataCenter')
```
seanforyou23, line 312: I wonder, could we just coerce these with `toLowerCase()`?

mturley: hmm, good point. I do like how TypeScript enforces that both the strings I compare against here are part of the string union type for that field though. It's such a tiny little thing that it's probably fine.

mturley: Good call @seanforyou23, while fixing that I realized there were 2 other places where I should have been checking it case-insensitively too. Pushed a commit.

seanforyou23: Sweet!

seanforyou23 (approved):

PR #739, opened 2021-08-05
Bug 1984471: Fix passing wrong bearer token to inventory API, clean up NetworkContext usage of auth token

mturley:
In https://github.com/konveyor/forklift-ui/pull/94 we set the `Authorization` header to pass the `oauth.clientSecret` from meta.json as the bearer token to the inventory API. That has been incorrect all along, and was just fine because the inventory ignores that token currently. @jortel is testing enabling auth in the inventory API and this is causing errors.

This PR passes the oauth `access_token` correctly in this header to resolve the new issues with the authenticated inventory requests.

I also cleaned up how we handle the `currentUser` value in our `useNetworkContext`. Instead of passing around the JSON-stringified value and parsing it where we need it, it gets parsed in that hook and the parsed value is available everywhere.

https://bugzilla.redhat.com/show_bug.cgi?id=1984471

### DIFF
--- a/src/app/client/helpers.ts
+++ b/src/app/client/helpers.ts
@@ -200,10 +200,9 @@ export const checkIfResourceExists = async (
 
 export const useClientInstance = (): KubeClient.ClusterClient => {
   const { currentUser } = useNetworkContext();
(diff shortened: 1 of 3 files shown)


seanforyou23 (approved):
😃

PR #746, opened 2021-08-18
fix(fetch): relax query/polling logic

seanforyou23:
This PR makes a few changes to how and when react-query goes about fetching data for the UI, mostly just relaxing the amount of refetching that's going on and attempts to better cache data where possible. Please double check for stale data in the UI as you test, we may need to manually invalidate a cache somewhere now that I've flipped the overall strategy to keep the previous data around by default.

Remove faster polling strategy
Reduce global polling interval
Introduce 15 second stale time
Enforce caching with keepPreviousData
Reduce amount of refetching/reindexing for vms trees

### DIFF
--- a/src/app/common/context/PollingContext.tsx
+++ b/src/app/common/context/PollingContext.tsx
@@ -1,14 +1,9 @@
-import {
-  AFTER_MUTATION_WINDOW,
-  POLLING_INTERVAL,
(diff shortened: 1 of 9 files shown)


fdupont-redhat (approved):
Confirmed by real world tests.

Issue #749, opened 2021-08-24
Disambiguate RHV objects with same name but from different datacenters


seanforyou23:
This is an upstream issue for tracking [bz-1994146](https://bugzilla.redhat.com/show_bug.cgi?id=1994146).

Description of problem:

When creating a migration plan from RHV, if the selected VMs belong to more than one datacenter and that two networks have the same name in each datacenter, the dropdown for the network mapping creation become ambiguous (see attachment). This will also happen for storage domains.

It will likely happen also when creating a network or storage mapping out of the Plan creation wizard. At least, this should be checked.

Steps to Reproduce:
1. In RHV provider with two datacenters, create two networks with the same name, one in each datacenter.
2. In RHV provider, create two virtual machines, one in each datacenter, using the previously created networks.
2. In MTV, add the RHV provider and wait for the inventory to complete.
3. In MTV, create a Plan with the wizard and select the previously created VMs.

Actual results:
When creating the network mapping, the dropdown will show two lines with the same network name and no other information to disambiguate them.

Expected results:
The dropdown shows the datacenter name along with the network name.

PR #750, opened 2021-08-24
Bug: 1994146 feat(mappings) add support for datacenter/directory meta for storage/network mappings

seanforyou23:
This PR adds some description text, when available, to the select dropdown for selecting storage/network maps. For ovirt I added information about the corresponding datacenter, and for vsphere I added information about the associated folder.

Also added a new query hook for polling datacenter data.

Screen Shot 2021-08-24 at 8 57 56 AM

Screen Shot 2021-08-24 at 8 59 04 AM

Screen Shot 2021-08-24 at 8 57 18 AM

Screen Shot 2021-08-24 at 8 58 14 AM

Should help close https://github.com/konveyor/forklift-ui/issues/749

### DIFF
--- a/src/app/Mappings/components/AddEditMappingModal.tsx
+++ b/src/app/Mappings/components/AddEditMappingModal.tsx
@@ -252,6 +252,7 @@ const AddEditMappingModal: React.FunctionComponent<IAddEditMappingModalProps> = 
                 >
                   <MappingBuilder
                     mappingType={mappingType}
+                    sourceProvider={form.values.sourceProvider}
                     sourceProviderType={form.values.sourceProvider?.type || 'vsphere'}
                     availableSources={mappingResourceQueries.availableSources}
                     availableTargets={mappingResourceQueries.availableTargets}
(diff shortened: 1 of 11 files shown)


fdupont-redhat:
@seanforyou23 is there a reason why Datacenter is minimized to DC? Not enough space on the line?

seanforyou23:
@fdupont-redhat nope - I'll expand it to use the full word. Didn't know if those names get long but looks like not.

vconzola (commented):
LGTM.

seanforyou23:
Pushed some updates - worked with @jortel and found an even better way to accomplish this than querying for datacenter data. Changes the format slightly, but I think its better. This metadata is now provided as part of the source object payload and gives even more useful info to the user. Screenshots updated, let me know what you think!

jortel:
LGTM

fdupont-redhat:
So, you're exposing the full path of the resource in vCenter, right?
Looks good to me as long as @vconzola is fine with it.
It would be great to also have approval from Amos, Avital and Ilanit, but I can't simply poke them here.

mturley:
@seanforyou23 nice work!!

I wonder if we should also change the OpenShift networks to have their namespace in a description under the items like you have here. They are currently shown inline, so if the namespace or network name is long it can get ugly:

Screen Shot 2021-08-30 at 4 27 11 PM

It might be nice to have just the network name in the main text and the namespace/name ( or just namespace) in the space below.

Issue #745, opened 2021-08-06
Implement must gather in UI


seanforyou23:
The "must gather" feature is a functionality that allows a user to collect a migration plan logs for diagnostics/analysis purposes. Let's implement it!

[REST API](https://github.com/aufi/must-gather-rest-wrapper)
[Designs](https://docs.google.com/presentation/d/19KuCgeRQUYPrkXmxiLCpdZ88NuikbN8Ajm16ZGAHYSo)

PR #757, opened 2021-08-30
Must gather UI

seanforyou23:
This PR introduces some new UI elements to make use of [must gather rest wrapper](https://github.com/aufi/must-gather-rest-wrapper) so we can kick that service off from within the UI.

Part of this work requires toast notifications to be displayed for certain actions, so in addition to hooks and types related to must gather, we've also added similar elements for a generic Notification service.

Screen Shot 2021-08-30 at 1 36 35 PM

Steps toward https://github.com/konveyor/forklift-ui/issues/745

### DIFF
--- a/src/app/queries/fetchHelpers.ts
+++ b/src/app/queries/fetchHelpers.ts
@@ -17,14 +17,13 @@ export const useFetchContext = (): IFetchContext => {
   return { history: useHistory(), checkExpiry, currentUser };
 };
 
-export const authorizedFetch = async <T>(
+export const authorizedFetch = async <TResponse, TData = unknown>(
   url: string,
   fetchContext: IFetchContext,
   extraHeaders: RequestInit['headers'] = {},
   method: 'get' | 'post' = 'get',
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  data?: any // TODO: type this
-): Promise<T> => {
+  data?: TData
+): Promise<TResponse> => {
   const { history, checkExpiry } = fetchContext;
   try {
     const response = await fetch(url, {
--- a/src/app/Plans/components/PlanActionsDropdown.tsx
+++ b/src/app/Plans/components/PlanActionsDropdown.tsx
@@ -182,7 +182,7 @@ export const PlanActionsDropdown: React.FunctionComponent<IPlansActionDropdownPr
             <ConditionalTooltip
               key="Restart"
               isTooltipEnabled={isPlanGathering}
-              content={'This plan cannot be restarted because it is running must gather service'}
+              content="This plan cannot be restarted because it is running must gather service"
             >
               <DropdownItem
                 isDisabled={isPlanGathering}
--- a/src/app/Plans/components/PlansTable.tsx
+++ b/src/app/Plans/components/PlansTable.tsx
@@ -60,7 +60,6 @@ import StatusCondition from '@app/common/components/StatusCondition';
 import MigrateOrCutoverButton from './MigrateOrCutoverButton';
 import PlanStatusNavLink from './PlanStatusNavLink';
 import { MustGatherBtn } from '@app/common/components/MustGatherBtn';
-import { useAppLayoutContext } from '@app/common/context';
 
 export type PlanActionButtonType = 'Start' | 'Cutover' | 'MustGather';
 interface IPlansTableProps {
@@ -73,7 +72,6 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
   errorContainerRef,
 }: IPlansTableProps) => {
   const [showArchivedPlans, toggleShowArchivedPlans] = React.useReducer((show) => !show, false);
-  const appLayoutContext = useAppLayoutContext();
   const providersQuery = useInventoryProvidersQuery();
   const migrationsQuery = useMigrationsQuery();
   const filterCategories: FilterCategory<IPlan>[] = [


src/app/client/types.ts
```diff
@@ -39,3 +39,23 @@ export interface IKubeStatus extends IMetaTypeMeta {
   };
   metadata: Partial<IMetaObjectMeta>;
 }
+
+export type mustGatherStatus = 'error' | 'inprogress' | 'completed' | 'new';
+
+export interface IMustGatherResponse {
+  ['archive-name']: string;
```
gildub, line 46: Well actually in that case I found out that not using the brackets is valid too. Up to you if you'd like.

src/app/common/components/MustGatherModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+    >
+      <TextContent>
+        <p>
+          The migration logs will be consolidated into a single archive file named{' '}
+          <strong>must-gather-{mustGatherContext.activeMustGather?.customName}.tar.gz</strong>.
+        </p>
+        <p>
+          The log collection process can take several minutes. You will be notified when the process
+          is complete and the tar archive file is ready to download.
+        </p>
```
gildub, line 91: Would it be possible to use PF here with <Text> and variants?

src/app/queries/fetchHelpers.ts
```diff
@@ -29,34 +32,33 @@ export const authorizedFetch = async <T>(
...
+  } catch (error: unknown) {
     checkExpiry(error, history);
     throw error;
   }
 };
 
-export const useAuthorizedFetch = <T>(url: string): QueryFunction<T> => {
+export const useAuthorizedFetch = <T>(
+  url: string
+  // extraHeaders: RequestInit['headers'] = {},
```
gildub, line 54: Do we need to keep those commented lines?

gildub (changes requested):
LGTM besides minors comments

src/app/queries/fetchHelpers.ts
seanforyou23: Nope! I'll remove them

src/app/client/types.ts
seanforyou23: I can drop the brackets for the definitions, mainly just the property access that requires it.

src/app/Plans/components/helpers.ts
```diff
@@ -235,3 +243,16 @@ export const canPlanBeStarted = (plan: IPlan): boolean => {
     false;
   return !hasEverStarted || hasSomeIncompleteVM;
 };
+
+export const latestAssociatedMg = (name: string, mustGathers: IMustGatherResponse[]) =>
```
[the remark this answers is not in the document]
mturley: I think it's intuitive enough as long as we expect someone reading the code to know that Mg means MustGather, which is clear from the types of the params.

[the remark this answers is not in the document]
gildub: No drama.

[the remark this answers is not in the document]
seanforyou23: Ah I see, I'll update it to spell out mustGather so it isn't misunderstood - good catch

src/app/common/components/MustGatherBtn.tsx
```diff
@@ -0,0 +1,105 @@
...
+}) => {
+  const {
+    setMustGatherModalOpen,
+    setActiveMustGather,
+    mustGatherWatchList,
+    setMustGatherWatchList,
+  } = React.useContext(MustGatherContext);
+
+  return (
+    <>
```
mturley, line 30: I don't think you need this wrapped in a `<>` fragment, you could just have the ternary be directly what you return:

```ts
return mg?.status === 'completed' && mg?.['archive-name'] ? (
  <Tooltip
  ...
) : (
  ...
);
```

src/app/common/components/MustGatherBtn.tsx
```diff
@@ -0,0 +1,105 @@
...
+            component={'a'}
+            target="_blank"
+            download
+            href={getMustGatherApiUrl(`must-gather/${mg?.['id']}/data`)}
+          >
+            Download logs
+          </Button>
+        </Tooltip>
+      ) : (
+        <>
```
mturley, line 51: Same here, no fragment needed if it only has a single child.

src/app/common/components/MustGatherBtn.tsx
```diff
@@ -0,0 +1,105 @@
...
+  } = React.useContext(MustGatherContext);
+
+  return (
+    <>
+      {mg?.status === 'completed' && mg?.['archive-name'] ? (
+        <Tooltip
+          content={
+            !mustGathers.isSuccess
+              ? `Could not reach must gather service.`
+              : `${mg?.['archive-name']} available for download`
```
mturley, line 36: TIL, I didn't know you could use optional chaining with bracket syntax like that! neat!

src/app/common/components/MustGatherBtn.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { WarningTriangleIcon } from '@patternfly/react-icons';
+import { MustGatherContext } from '@app/common/context';
+import { getMustGatherApiUrl } from '@app/queries/helpers';
+import { IMustGatherResponse } from '@app/client/types';
+import { UseQueryResult } from 'react-query';
+import { IPlan } from '@app/queries/types';
+
+interface IMustGatherBtn {
+  mg: IMustGatherResponse | undefined;
+  mustGathers: UseQueryResult<IMustGatherResponse[], Response>;
```
mturley, line 11: I wonder if this `mustGathers` prop should be `mustGathersQuery` or `mustGathersResult`. The name implies it's a plain array of MGs, so below when you use `mustGathers.isSuccess` it reads weird.

src/app/common/components/MustGatherBtn.tsx
```diff
@@ -0,0 +1,105 @@
...
+          content={
+            !mustGathers.isSuccess
+              ? `Could not reach must gather service.`
+              : `${mg?.['archive-name']} available for download`
+          }
+        >
+          <Button
+            isAriaDisabled={!mustGathers.isSuccess}
+            variant="secondary"
+            component={'a'}
```
mturley, line 42: Can be a regular string prop `component="a"` (but you can ignore that level of nitpick if you want lol)

src/app/common/components/MustGatherBtn.tsx
```diff
@@ -0,0 +1,105 @@
...
+          >
+            Download logs
+          </Button>
+        </Tooltip>
+      ) : (
+        <>
+          <Tooltip
+            content={
+              !mustGathers.isSuccess
+                ? `Could not reach must gather service.`
```
mturley, line 55: Just an observation, not a criticism: this is a perfect use case for nested ternaries and would be a mess without them, but SonarCloud is yelling at you to not do that... I really want to figure out how to turn off some of those rules.

src/app/common/components/MustGatherBtn.tsx
```diff
@@ -0,0 +1,105 @@
...
+                : mg?.status === 'error'
+                ? `Could not complete must gather for ${mg['custom-name']}`
+                : `Collects the current ${
+                    type === 'plan' ? 'migration plan' : 'VM migration'
+                  } logs and creates a tar archive file for download.`
+            }
+          >
+            <Button
+              icon={mg?.status === 'error' ? <WarningTriangleIcon /> : null}
+              isLoading={
```
mturley, line 69: I really like this new `isLoading` prop of button, there are definitely other places in our UI where we just disable a button and show a spinner below it or whatever. Pretty much everywhere we use `<ResolvedQuery />` to handle a mutation... that would be neat to clean up later.

src/app/common/components/MustGatherBtn.tsx
```diff
@@ -0,0 +1,105 @@
...
+                mustGatherWatchList.map((mg) => mg.name).includes(customName) &&
+                !mustGathers.isError &&
+                (mg?.status === 'inprogress' || mg?.status === 'new')
+              }
+              isAriaDisabled={
+                (mustGatherWatchList.map((mg) => mg.name).includes(customName) &&
```
mturley, line 75: Is it worth factoring out `mustGatherWatchList.map((mg) => mg.name).includes(customName) && (mg?.status === 'inprogress' || mg?.status === 'new')` into some variable, or am I overthinking this? Not even sure what we'd call it.

src/app/common/components/MustGatherBtn.tsx
```diff
@@ -0,0 +1,105 @@
...
+              variant="secondary"
+              onClick={() => {
+                setMustGatherModalOpen(true);
+                setActiveMustGather({
+                  type,
+                  customName: customName,
+                  status: 'new',
+                });
+
+                setMustGatherWatchList([
```
mturley, line 88: I think maybe we should add a code comment here describing what you explained to me yesterday about why we need to `setMustGatherWatchList` here instead of relying on the query. We might also want to put a comment in the context provider where the `[mustGatherWatchList, setMustGatherWatchList]` state is defined, because without seeing this usage, it looks like unnecessary duplication of state when you look at the provider code alone.

It might also be worth opening an issue and following up if we can get the backend to reflect this state properly when we reload the list, so we can clean this part up.

src/app/common/components/MustGatherModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+    }
+  };
+
+  const registerMustGather = useMustGatherMutation(
+    'must-gather',
+    handleMustGatherSuccess,
+    handleMustGatherError
+  );
+
+  const handleMustGatherRequest = ({ customName, type }: mustGatherObjType) => {
```
mturley, line 41: Nitpick, but it feels weird to have `mustGatherObjType` not be capitalized. At first glance it seems like a value and not a type (despite the name and syntax... take it or leave it lol)

src/app/common/components/MustGatherModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { MustGatherContext, AppContext, mustGatherObjType } from '@app/common/context';
+import { useMustGatherMutation } from '@app/queries';
+
+export const MustGatherModal: React.FunctionComponent = () => {
+  const mustGatherContext = React.useContext(MustGatherContext);
+  const { pushNotification } = React.useContext(AppContext);
+
+  const handleMustGatherSuccess = () => {
+    if (mustGatherContext.activeMustGather) {
+      pushNotification({
```
mturley, line 12: Just thinking out loud, because I like this: in other places where we have a modal that calls a mutation, we show the result of the mutation by keeping the modal open until the mutation resolves succesfully, and showing any errors as inline alerts in the modal (by using `<ResolvedQuery />`... incidentally we may want to factor out its error-parsing bits for use in push notifications).

Here though, the modal closes immediately and the result is shown as a push notification. I think it works here because you're kicking off a long-running process, and because there is no input in the modal. But it's inconsistent with our other mutation-modals... I'm trying to determine if we should open an issue to make them consistent later or if they should be different. I like the push notifications a lot and was about to suggest we switch to them everywhere, but in the other places I'm referring to, we want to preserve the form data in the modal so they can make a correction and try again if the mutation fails. 🤔 

It does make me think though that we should use a push notification for kicking off other long running tasks, like when a migration starts or is canceled, or maybe even when we detect a VM failing (yikes, that's a whole future project. Don't tell Miguel yet)

src/app/common/components/MustGatherWatcher.tsx
```diff
@@ -0,0 +1,94 @@
...
+  name: string;
+  listStatus: mustGatherStatus;
+}
+
+export const MustGatherWatcher: React.FunctionComponent<IMustGatherWatcherProps> = ({
+  name,
+  listStatus,
+}: IMustGatherWatcherProps) => {
+  const { pushNotification } = React.useContext(AppContext);
+  const completedPreviously = listStatus === 'completed';
```
mturley, line 18: Trying to figure out if these variables could have better names, since the behavior isn't obvious right away... I think maybe they're as good as they'll get. I was thinking of like `completedOnMount` / `erroredOnMount` but those might be more confusing, not less. Maybe a comment could help, but it's also fine as is.

This watcher pattern is so slick, man. Nice work.

src/app/common/components/Notifications.tsx
```diff
@@ -0,0 +1,43 @@
...
+export interface INotification {
+  title: string;
+  message: string | JSX.Element;
+  variant: notificationType;
+  key: string;
+  actionClose?: boolean;
+  timeout?: number | boolean;
+}
+
+export const Notifications: React.FunctionComponent = () => {
```
mturley, line 16: Hell yeah. Gonna use the crap out of these notifications.

src/app/common/context/AppContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+const appContextDefaultValue = {} as IAppContext;
+export const AppContext = React.createContext<IAppContext>(appContextDefaultValue);
+
+interface IAppContextProvider {
+  children: React.ReactNode;
+}
+
+export const AppContextProvider: React.FunctionComponent<IAppContextProvider> = ({
```
mturley, line 17: Okay, I know I've said this before and backed down, but part of me is worried that if we introduce an `AppContext`, in the future we'll end up shoving a bunch of unrelated stuff together in there any time we have a need for something to be globally available. Maybe that's fine, but I feel like it could become a mess. I bet you if we'd called it `AppContext` when we started out with `NetworkContext`, we'd have ended up lumping in the stuff that is now in `LocalStorageContext` and `PollingContext` etc.

It's not too late to call this `NotificationContext` :) but also I know I'm being paranoid and it's fine.

src/app/common/context/MustGatherContext.tsx
```diff
@@ -0,0 +1,118 @@
...
+export type mustGatherListType = mustGatherObjType[];
+interface IMustGatherContext {
+  mustGatherModalOpen: boolean;
+  setMustGatherModalOpen: (isOpen: boolean) => void;
+  mustGatherList: mustGatherListType;
+  setActiveMustGather: (mustGatherObj: mustGatherObjType) => void;
+  activeMustGather: mustGatherObjType | undefined;
+  mustGatherWatchList: mgWatchListType;
+  setMustGatherWatchList: (list: mgWatchListType) => void;
+  mustGathers: UseQueryResult<IMustGatherResponse[], Response>;
```
mturley, line 28: Ah, I didn't realize it comes all the way from up here. Per my earlier comment, consumer usage might be more readable if we called this `mustGathersQuery` or `mustGathersResult`.

src/app/common/context/MustGatherContext.tsx
```diff
@@ -0,0 +1,118 @@
...
+);
+
+interface IMustGatherContextProvider {
+  children: React.ReactNode;
+}
+
+export const MustGatherContextProvider: React.FunctionComponent<IMustGatherContextProvider> = ({
+  children,
+}: IMustGatherContextProvider) => {
+  const appContext = React.useContext(AppContext);
```
mturley, line 43: Hm... okay, AppContext does make sense as a name if it means it's the only context we guarantee all other context providers should have access to. Hmm.

src/app/common/context/MustGatherContext.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const [mustGatherModalOpen, setMustGatherModalOpen] = React.useState(false);
+  const [mustGatherList, setMustGatherList] = React.useState<mustGatherListType>([]);
+  const [activeMustGather, setActiveMustGather] = React.useState<mustGatherObjType>();
+  const [mustGatherWatchList, setMustGatherWatchList] = React.useState<mgWatchListType>([]);
+  const [errorCount, setErrorCount] = React.useState(0);
```
mturley, line 48: Can we make this a boolean like `errorNotified` or something? It's fine if you want to leave it as a number, it just makes me want to scan the code to see what happens when errors add up.

src/app/common/context/MustGatherContext.tsx
```diff
@@ -0,0 +1,118 @@
...
+        activeMustGather,
+        setActiveMustGather,
+      }}
+    >
+      <>
+        {children}
+        {mustGatherList.map((mg, idx) => {
+          return (
+            mg.customName && (
+              <div data-watcher={mg.customName} data-status={mg.status} key={idx}>
```
mturley, line 105: I'm on the fence here, but I think we might want to remove this so we don't have unnecessary DOM elements. Considering our other performance issues this is a drop in the bucket though, so I'm fine with leaving it. Either way, I think you can also keep track of watchers by looking at the `must-gather-entity` queries in the react-query dev tools.

src/app/index.tsx
```diff
@@ -30,13 +32,17 @@ const App: React.FunctionComponent = () => (
...
       <LocalStorageContextProvider>
         <NetworkContextProvider>
-          <Router getUserConfirmation={noop}>
-            <AppLayoutContextProvider>
-              <AppLayout>
-                <AppRoutes />
-              </AppLayout>
-            </AppLayoutContextProvider>
-          </Router>
+          <AppContextProvider>
```
mturley, line 35: One day I'll write a `<ProviderProvider />`, I feel like every time I say that I'm joking a little less.

src/app/queries/fetchHelpers.ts
```diff
@@ -20,7 +20,10 @@ export const useFetchContext = (): IFetchContext => {
 export const authorizedFetch = async <T>(
   url: string,
   fetchContext: IFetchContext,
-  extraHeaders: RequestInit['headers'] = {}
+  extraHeaders: RequestInit['headers'] = {},
+  method: 'get' | 'post' = 'get',
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  data?: any // TODO: type this
```
mturley, line 26: I think maybe the correct way to type this is to have it be another generic type param that defaults to unknown, e.g. instead of `<T>` we could have `<TResponse, TData = unknown>`. That way if we really need to we can restrict the type of the object we pass into a particular call. That's how react-query handles this type of thing internally.

If TS doesn't like passing `unknown` into `JSON.stringify()` we could use `any` as the default instead, but it'd be nice to not have to disable eslint here. And this feels like a good use case for `unknown`... it's not that the param can be anything, it's that this function doesn't care what it is.

src/app/queries/helpers.ts
```diff
@@ -86,6 +86,9 @@ export const useMockableMutation = <
 export const getInventoryApiUrl = (relativePath: string): string =>
   `/inventory-api/${relativePath}`;
 
+export const getMustGatherApiUrl = (relativePath?: string): string =>
+  relativePath ? `/must-gather-api/${relativePath}` : `/must-gather-api/`;
```
mturley, line 90: Nitpick, but you could avoid repeating the base path by using this instead:

```ts
export const getMustGatherApiUrl = (relativePath?: string): string =>
  `/must-gather-api/${relativePath || ''}`;
```

src/app/queries/mustGather.ts
```diff
@@ -0,0 +1,100 @@
...
+      },
+    },
+    MOCK_MUST_GATHERS
+  );
+  return result;
+};
+
+// monitor a single must gather
+export const useMustGatherQuery = (
+  planName: string,
```
mturley, line 77: Now that this can be VMs too, should this be `customName` instead of `planName`?

src/app/Plans/components/VMMigrationDetails.tsx
```diff
@@ -290,6 +295,16 @@ const VMMigrationDetails: React.FunctionComponent = () => {
               vmStatus.warm?.precopies?.length || 0,
               { title: <VMWarmCopyStatus vmStatus={vmStatus} isCanceled={isCanceled} /> },
             ]),
+        {
+          title: (
+            <MustGatherBtn
+              type="vm"
+              customName={getVMName(vmStatus)}
```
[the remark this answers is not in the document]
seanforyou23: Hmm, good point!

src/app/common/components/MustGatherModal.tsx
seanforyou23: ha it's a good point. Both strategies have a different UX and which one to use probably just depends on the use case. The notifications are a new feature, so I'm thinking not go too fast and put them everywhere all at once, but we should definitely make use of them anywhere it improves the flow with no downside.

src/app/common/context/AppContext.tsx
seanforyou23: I like NotificationContext :)

You convinced me at the point we created AppLayoutContextProvider 😆

src/app/common/context/MustGatherContext.tsx
seanforyou23: this was part of the thought process, but I suppose it's ok for Providers to depend on other arbitrary providers - we just have to keep the component nesting in order

src/app/common/components/MustGatherBtn.tsx
seanforyou23: Yea, it starts to feel like the plan state/status.. it's mainly just the two scenarios for now so if you're cool with it we can leave as it, and if another condition comes into play we can step back and think about what they have in common - should help the naming of the abstraction.

src/app/common/components/MustGatherBtn.tsx
seanforyou23: Agree!

src/app/common/components/MustGatherWatcher.tsx
seanforyou23: Yea naming these was tricky.. happy to update it if you can think of something more descriptive. Might be good enough for a first pass..

src/app/common/components/MustGatherBtn.tsx
mturley: That sounds good. We'll have plenty of opportunities to abstract better later.

src/app/Plans/components/helpers.ts
```diff
@@ -85,6 +85,17 @@ export const getMigStatusState = (state: PlanState | null, isWarmPlan: boolean)
   };
 };
 
+export const canBeRestarted = (planState: PlanState | null) => {
+  return (
+    !planState ||
```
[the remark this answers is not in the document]
seanforyou23: This that was just leftover from when I was getting the button to display in the right scenarios, removed it and seems to be working as intended. Good catch!

src/app/common/context/AppContext.tsx
mturley: Ha, fair. That one is maybe a bit rough lol

gildub (approved):
Nice work!

mturley:
Looks like I caused some minor merge conflicts with https://github.com/konveyor/forklift-ui/pull/777, sorry :(

seanforyou23:
ha! All good, I've a couple minor things to clean up anyway, will rebase shortly

mturley (approved):
Nice work man!

aufi:
Nice work Michael!

Issue #760, opened 2021-08-31
Convert Express server (`deploy/server.js`) to TypeScript


mturley:
With the http-proxy-middleware usage getting more complex now that we're working towards websocket support, it would be very helpful to have TS in place for the NodeJS server code (currently the only plain JS in our repo).

@seanforyou23 and I discussed an approach here involving a simple `tsc` build step for prod mode and `nodemon`'s TS support for a smooth experience in dev mode.

gildub:
For our monorepo organization we can take benefit from [Typescript project references] (https://www.typescriptlang.org/docs/handbook/project-references.html) and also [NPM workspaces](https://docs.npmjs.com/cli/v8/using-npm/workspaces).

Meanwhile to be able to use NPM workspaces we need to bump to NPM version 7 or 8 and therefore use a more recent Nodejs LTS version.

yaacov:
closing as new development is moving to:
https://github.com/kubev2v/forklift-console-plugin

PR #762, opened 2021-09-01
feat(types): convert server to typescript

seanforyou23:
This PR converts the primary express server app to typescript as well as some of the other development scripts. I added a config file for nodemon so all the options can be stored in a single place. With these changes we get both type safety as well as hot-reloading of the express server when running locally in development mode.

I fixed a couple minor issues that were discovered once ts was enabled for the api, one however I'm not sure what the best fix would be so I left that as a commonjs import for now. Let me know if anyone has an idea on the best fix for it.

Screen Shot 2021-09-01 at 10 45 04 AM

Should help close https://github.com/konveyor/forklift-ui/issues/760


mturley:
@seanforyou23 for the error you point out in the screenshot here, it looks like the problem is coming from `const { code } = req.query;`, where `code` can be `string | string[] | qs.ParsedQs | qs.ParsedQs[] | undefined`. I bet we can pass some generic type params to `app.get<...>(...)` to specify what query params are there and what their types are, so that `code` gets inferred to just be a `string`. Worst-case, we could instead have `const code = req.query.code as string`. We should probably also either put the `AuthorizationTokenConfig` type on `const options`, or just inline that object inside the `clusterAuth.getToken(...)` call.

seanforyou23:
Yea I've been going through and making small adjustments here and there as the ts encounters issues. As a first/initial pass I'd like to try and find a way to use that custom tsconfig for the api/nodemon to reference instead of the one that's imported from webpack which is what's used for the UI.. these configs are generally incompatible.If we can keep them totally separate we'll be able to relax the typescript settings in the api pkg independent from the UI and can make these types of enhancements in followup efforts.

mturley:
I think it makes sense to keep the tsconfigs completely separate, since one has to worry about a bunch of React stuff that the other won't care about. I think it's ok if they are a little repetitive (not sure if there is a mechanism in tsconfig to extend a common one, though, that could be neat).

I wonder if it would help to combine the `deploy` and `scripts` directories and then put a `tsconfig.json` and `nodemon.json` in there? treat them like totally separate projects essentially?

mturley:
> Can we try to get the configs sorted first and then circle back to actually typing it out?

Yeah I think that makes sense. So maybe just moving nodemon into `deploy` and leaving `scripts` alone for now is good.

seanforyou23:
Ok so the build target now works - feel free to test it. The dev server isn't as happy with the settings though. If you add `"type": "module"` to package.json we go down a different path where imports work but require doesn't. Else if it's commonjs (the default - and what I believe it should be) the dev server **either** gets hung at `[nodemon] clean exit - waiting for changes before restart` or compiles with errors like `SyntaxError: Cannot use import statement outside a module` or 

```sh
[EXPRESS] Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /path/fo/forklift-ui/node_modules/node-fetch/src/index.js
[EXPRESS] require() of ES modules is not supported.
```
or
```sh
[EXPRESS] /path/to/forklift-ui/deploy/server.ts:1
[EXPRESS] import { __assign, __awaiter, __generator } from "tslib";
[EXPRESS] ^^^^^^
[EXPRESS] SyntaxError: Cannot use import statement outside a module
```
depending on which module format you specify in package.json and tsconfig.json.

I've explored all these paths and you'll see I have commented out in the code where you can easily convert these files between commonjs/es6 for testing purposes.. But regardless of which path I go down, I either way run into the problem of inoperability between the modules. `esModuleInterop` should allow for this and I'm stumped as to why it isn't working.

I upgraded ts and webpack which helped give some better error messages but in the end after I've fixed all the errors I'm not sure how to get past any version of this problem (short of introducing lerna and making separate ui and api sub-packages - which I'm happy to do if people want). Anyway I've probably already spent too much time on this, but if anyone wants to try to help I think the main thing left is figuring out how to prevent webpack/typescript from trying to parse everything in the dependency tree the same in dev mode. Ideally we could leave all those files (except server.ts) as .js files and they should continue to work as they do now. The methods I've tried to exclude those directories don't seem to work.

seanforyou23:
I took a stab at moving toward a [monorepo](https://github.com/seanforyou23/forklift-ui/tree/monorepo) structure, still hitting a similar issue but it gives us more options none the less. I think we could explore using some other settings for the [nodemon config](https://github.com/seanforyou23/forklift-ui/commit/3c51ef18ec77ed0feed37453fbc698bbe8deba35#diff-620a573ed1ed80153a3e5eae368c8306b58765d6dea305ff633b0608be1ffee9R5), like specifying `--loader ts-node/esm`? Anyway, just wanted to mention before I pause this work for a bit.

PR #765, opened 2021-09-06
Make sure prettier is properly plugged with eslint

gildub:
Having eslint autosave not working with my VS Code environment, I verified the eslint/prettier config and noticed the prettier plugin wasn't full configured. 

Also:
Add `.eslintignore`file and changed lint rule to apply across the project.
The latter helps detect eslint issues in webpack configuration files.
Remove `.prettierignore` file which contained only package.json and seemed not relevant anymore.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,8 +1,8 @@
 {
-  // tells eslint to use the TypeScript parser
+  "root": true,
   "parser": "@typescript-eslint/parser",
-  // tell the TypeScript parser that we want to use JSX syntax
   "parserOptions": {
+    "ecmaVersion": 2015,
     "tsx": true,
     "jsx": true,
     "js": true,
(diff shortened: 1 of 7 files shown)


seanforyou23 (approved):

.eslintrc
```diff
@@ -31,8 +32,7 @@
       "version": "^16.11.0"
     }
   },
-  // includes the typescript specific rules found here: https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin#supported-rules
-  "plugins": ["@typescript-eslint", "react-hooks", "eslint-plugin-react-hooks"],
+  "plugins": ["prettier", "@typescript-eslint", "react-hooks", "eslint-plugin-react-hooks"],
```
mturley, line 35: Hm, I'm surprised that adding this makes a difference. According to https://github.com/prettier/eslint-plugin-prettier#recommended-configuration, the line `"extends": ["plugin:prettier/recommended"]` has the same effect as:

```
{
  "extends": ["prettier"],
  "plugins": ["prettier"],
  "rules": {
    "prettier/prettier": "error",
    "arrow-body-style": "off",
    "prefer-arrow-callback": "off"
  }
}
```.

So the only effect this change has is the addition of `plugin:@typescript-eslint/eslint-recommended` to the `extends` array.

File: .eslintrc [hunk 2]
```diff
@@ -42,7 +42,13 @@
     "import/no-unresolved": "off",
     "import/extensions": "off",
     "react/prop-types": "off",
-    "prettier/prettier": "error"
+    "prettier/prettier": [
+      "error",
+      {
```
mturley, line 47: This now duplicates the options that are also defined in `.prettierrc`. I wonder if there's a way to make sure it's reading that file instead here?

mturley (commented):
Thanks for this @gildub ! It's nice to have linting for everything.

I noticed a couple things about the changes here, but I'm not sure we need to do anything about it. Just curious about your thoughts.

Issue #35, opened 2020-09-09
Add "expand/collapse all" toggle to the tree view in step 2 of the plan wizard


mturley:
Currently this is not possible due to a limitation of the PatternFly TreeView. This needs to be solved in PatternFly itself, but I'm opening this issue so we remember to follow up and add that support here once it is available.

mturley:
Upstream issue to track: https://github.com/patternfly/patternfly-react/issues/4797

mturley:
This has now been implemented in PatternFly: https://github.com/patternfly/patternfly-react/pull/5503

gildub:
Following up with https://github.com/patternfly/patternfly-react/issues/5933

gildub:
@mturley , @seanforyou23 ,

Because for the Plan's wizard the top level of Treeview's FilterVMsForm is always a single node "All datacenters" there is no need for a collapse/expand all toggle. 
The default behavior (not sure that was always the case for prior PF core version, but I can't see it in current master and following versions) automatically collapses/expands all sub-nodes when clicking a given node arrow (> or v).
So clicking "All datacenters" node arrow fullfills the requirement to collapse/expand all nodes.

Let me know what you think.

mturley:
Ha! That's a good point, I hadn't considered that we basically already have this functionality. @vconzola do you have an opinion on this? A separate expand/collapse all button for the tree might be overkill.

vconzola:
@mturley I don't like the default behavior if it means that clicking an arrow expands all sub-nodes below it. That's a terrible idea because it doesn't scale well for large trees. I'll bring it up with Matt. I'd like to keep (or go back to) our current design where clicking an arrow only expands the nodes immediately below the selected node. And keep the select all toggle. Can we do this?

mturley:
@vconzola looking into it more closely, it looks like when we passed the `defaultAllExpanded` prop (to make the whole tree start out fully expanded... if we're worried about larger trees, do we still want this behavior?) that causes the expand action to always expand all sub-nodes. That sounds to me like a bug in the TreeView.

However, @gildub now that the TreeView supports the `allExpanded` prop that we can use to implement the expand/collapse-all toggle, I think we can drop `defaultAllExpanded` and just default the `allExpanded` state to true. I'm not sure what the point of that prop is now that the other exists anyway, it seems there's some weird stuff going on with TreeView.

mturley:
[...]

PR #766, opened 2021-09-07
TreeView: Bump PF core to move search field to toolbar and adjust expandAll behavior

gildub:
I realized a toggle seems redundant, [See comment](https://github.com/konveyor/forklift-ui/issues/35#issuecomment-915082796)

Therefore this patch helps with following:
- Bumps to latest Treeview and by doing so moves the search field to the toolbar.
- Removes `defaultAllExpanded` 
- Defaults `allExpanded` to true
- Adds `hasGuides` 

Edited by @mturley : removed annotation about fixing #35, since we still want the expand/collapse-all after https://github.com/patternfly/patternfly-react/issues/6314 is resolved.


gildub:
@vconzola, could you please validate?

vconzola:
@gildub I'm not sure what I'm looking for. What page(s) does this impact?

gildub:
@vconzola this is for Plan Wizard step 2 of VMSelection during filtering

vconzola:
@gildub Sorry, but I don't see what the difference is? Is the change something below the presentation layer. The Search bar looks the same to me?

mturley:
@vconzola yes, this is a refactor under the hood due to a change in PatternFly. It gives us a slot in the toolbar to add the expand/collapse control for https://github.com/konveyor/forklift-ui/issues/35 that we didn't have before, but this PR doesn't put anything there.

mturley:
Hmm, @gildub seeing your fix for the CI, I realize that I didn't notice the tree was collapsed by default. It looks like `allExpanded={true}` isn't working as expected. We may need to fix that and put the line back in the test that you removed.

mturley:
@gildub I realized that this is a bug in PF, which I've opened here: https://github.com/patternfly/patternfly-react/issues/6314

Maybe we should just go back to `defaultAllExpanded` for now until this is resolved. Sorry to send you down the wrong path.

gildub:
@mturley, no worries, bug happens ;)
I've restored `defaultAllExpanded` for now.

mturley (approved):
LGTM!

PR #769, opened 2021-09-13
fix(deps): bump tslib

seanforyou23:
Fixes minor type issues.

Screen Shot 2021-09-13 at 3 33 29 PM

Screen Shot 2021-09-13 at 4 31 31 PM

### DIFF
--- a/src/app/queries/hooks.ts
+++ b/src/app/queries/hooks.ts
@@ -38,7 +38,7 @@ export const playbookSchema = yup
   .test('valid-yaml', 'Playbook must be valid YAML', (value, context) => {
     try {
       yaml.load(value || '');
-    } catch (e) {
+    } catch (e: any) {
       if (e.reason && e.mark) {
         return context.createError({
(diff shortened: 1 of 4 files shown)


src/app/queries/hooks.ts
```diff
@@ -38,7 +38,7 @@ export const playbookSchema = yup
   .test('valid-yaml', 'Playbook must be valid YAML', (value, context) => {
     try {
       yaml.load(value || '');
-    } catch (e) {
+    } catch (e: any) {
```
mturley, line 41: I was about to say we can use `YAMLException` from `js-yaml` as the type here, but disappointingly they use `any` for the types of `reason` and `mark` anyway. So not much would be gained there.

mturley (approved):
LGTM, thanks!

src/app/queries/hooks.ts
seanforyou23: Ah good call - we can make that change in a followup PR.. every little bit helps I suppose :)

Issue #770, opened 2021-09-14
Allow to archive a plan from the UI


seanforyou23:
As a user, I want to be able to archive a Plan from the MTV UI, in order to stay in the MTV UI for all migration related operations.

A plan can be archived at any stage, but running. If a plan is running, it must be cancelled before being archived, so that all assets / dependencies are at rest.

When a plan is archived, the only available options are "Clone to new plan".

We want to consider moving archived plans out of the main plan view, to avoid cluttering it with not actionable items.

PR #771, opened 2021-09-14
feat(plans): add ui bits for archiving a plan

seanforyou23:
This PR updates the UI to account for archiving plans. I also added an AppLayoutContext component in order to track/notify arbitrary parts of the UI when we're in a mobile sized viewport - this was needed because the Switch/toggle for showing archived plans label doesn't fit well on smaller viewports, so for those cases we hide the label altogether.

@mturley I did try out that [`YAMLException`](https://github.com/konveyor/forklift-ui/pull/769#discussion_r707773435) you mentioned for the hooks error type but it wasn't working for me so I ignored them for now.

Screen Shot 2021-09-14 at 7 20 42 PM

Steps toward closing https://github.com/konveyor/forklift-ui/issues/770 and [MTV-110](https://issues.redhat.com/browse/MTV-110).


mturley:
LGTM! @seanforyou23 do you think we should hold off on merging until we have backend support for the true archived flag, or should we merge and patch later? I think the latter would be fine to avoid a long-running PR, because the UX is in place and doesn't feel broken.

mturley (approved):

seanforyou23:
@mturley either way works for me. I think I'll need to make a minor tweak to some of the tooltips and also the available actions for archived plans - maybe we can leave it open for now til the UX for all that is settled.

fdupont-redhat:
@mturley @seanforyou23 go ahead and merge. The backend is not ready yet, but the PR won't break anything.

fdupont-redhat (approved):

mturley:
Agreed on leaving it open for the little UX quirks but not waiting for the backend.

mturley:
@seanforyou23 can we put the archived annotation on at least one plan in the mock data, so you can see the toggle in action in mock mode? Also, do we have any way of visually distinguishing an archived plan in the list? Maybe we should show "Archived" in the status column or something.

seanforyou23:
Good call @mturley - keepin me honest 😄

vconzola (commented):
A couple comments and a question.

1. The placement and spacing of the "Show archived" toggle looks a little funky. Lemme take a look at PF guidelines. We may go rogue.
2. I don't know if we need an "unarchive" (probably should be called "restore") action. Need to ask Miguel / Fabien.
3. What causes a "Finished - Incomplete" state? Is that if user cancels? I don't know if those plans should be archivable or not.

fdupont-redhat:
> A couple comments and a question.
> 
> 1. The placement and spacing of the "Show archived" toggle looks a little funky. Lemme take a look at PF guidelines. We may go rogue.
> 2. I don't know if we need an "unarchive" (probably should be called "restore") action. Need to ask Miguel / Fabien.

I don't think we should unarchive for now. To me, it's an action to save space by cleaning up the pods for a plan that won't be used anymore. The user has the ability to clone it if needed. And I guess it's easy to add the unarchive later if we want to.

> 3. What causes a "Finished - Incomplete" state? Is that if user cancels? I don't know if those plans should be archivable or not.

It can also be a plan in which some VMs have failed. The user knows why and doesn't care about the logs anymore. So, it can be archived to clear pods, while keeping metadata for history.

seanforyou23:
Good stuff! I think it's all addressed with the latest on this thread. Updated screenshots and rebased, let me know if you need anything else on this run!

vconzola:
LGTM. The only thing I'm not sure of are the Incomplete plans. I thought if any VM failed to migrate, we Failed the plan and that Incomplete plans are still considered in-progress. @mturley can you confirm what Finished - Incomplete means in the ui?

seanforyou23:
I think it comes from this [condition](https://github.com/konveyor/forklift-ui/blob/main/src/app/Plans/components/helpers.ts#L147-L149), which checks that a plan has been started, and yet is still able to be started (likely because [not all the vms contained "Succeeded" conditions](https://github.com/konveyor/forklift-ui/blob/main/src/app/Plans/components/helpers.ts#L222-L226) so while the migration plan ran and the run was considered finished, it was incomplete from a success point of view. That's my best guess anyway :)

PR #772, opened 2021-09-14
Revert to package-lock.json lockfileVersion@1 after #766

mturley:
See context in https://github.com/konveyor/forklift-ui/pull/689, we should keep our lockfile at v1 for now. It was upgraded to v2 in #766 and I didn't catch it before merging.

@gildub I know it's a hassle, but would you mind downgrading your node/npm to LTS versions (node 14.17.6 which includes npm 6.14.15) for this repo? If you need the newer versions for another project, I recommend using https://github.com/tj/n to manage multiple versions (I find it more reliable than nvm, but that's personal preference)


gildub:
@mturley, yes I notice the error message recently while building my own image.
I've downgraded my node/npm version. Thank you very much for taking care of this.

gildub (approved):

mturley:
No problem! Thanks @gildub

PR #775, opened 2021-09-15
Websocket support

seanforyou23:
Figured I'd join the club and post my draft PR for websocket support for comparison. It's pretty similar in implementation to https://github.com/konveyor/forklift-ui/pull/768/ - I've exhausted a few more options with the `http-proxy-middleware` and `ws` packages, but while I can send events between the UI and node app, I can't seem to receive events from the inventory on the node app (although I was able to get same type confirmation logs for "Upgrading to websocket" as observed in https://github.com/konveyor/forklift-ui/pull/774.

I think next step is to try swapping usage of either of those two npm packages (just for the ws endpoint) and see if we can get message events going from the inventory to the node app - once we are there we're home free as communication from the react app and the websocket server is already working great.

The new WebSocket constructor in the browser seems to require `ws://` schema to be specified, which prevents us from being able to use the dev server proxy config, so in the mean time I'm just hitting the express server directly for now. I did get a tip from a fellow UXD member to check into the webpack [webSocketServer](https://webpack.js.org/configuration/dev-server/#devserverwebsocketserver) config, so I'm thinking that will help move things along on that front. I'll give it a spin as soon as I get back to this work.

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -169,9 +169,6 @@ module.exports = (env) => {
       }),
     ],
     resolve: {
-      alias: {
-        'react-dom': '@hot-loader/react-dom',
-      },
       extensions: ['.js', '.ts', '.tsx', '.jsx'],
       plugins: [
         new TsconfigPathsPlugin({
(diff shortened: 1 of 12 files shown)


seanforyou23:
Closing for now, we can revive this work later.

PR #776, opened 2021-09-16
Add tar to the final container image

fabiendupont:
To be able to use `kubectl scp`, the container needs to have the `tar` command. This PR adds it.


gildub (approved):
LGTM

Issue #779, opened 2021-09-27
Namespace custom names for must gather


seanforyou23:
Currently we use the plan name as is for the customName in must gather. It could lead to conflicts as nothing prevents someone from using the same name for a plan/vm. Let's ensure these customNames are unique by namespacing them as necessary.

PR #780, opened 2021-09-27
chore(mg): ensure must gather custom names are namespaced

seanforyou23:
This PR prefixes the customName used for must gather service to help avoid a case where users may have the same name for a plan/vm must gather. Was able to get rid of the inprogress glue code we used before to determine if a must gather was being executed.

Should help close https://github.com/konveyor/forklift-ui/issues/779.

### DIFF
--- a/src/app/common/components/MustGatherModal.tsx
+++ b/src/app/common/components/MustGatherModal.tsx
@@ -11,9 +11,9 @@ export const MustGatherModal: React.FunctionComponent = () => {
   const handleMustGatherSuccess = () => {
     if (activeMustGather) {
       pushNotification({
-        title: `Started must gather for ${activeMustGather.customName}`,
+        title: `Started must gather for ${activeMustGather.displayName}`,
         message: '',
-        key: activeMustGather.customName,
+        key: activeMustGather.displayName,
         variant: 'info',
         actionClose: true,
         timeout: 4000,
@@ -24,9 +24,9 @@ export const MustGatherModal: React.FunctionComponent = () => {
   const handleMustGatherError = () => {
     if (activeMustGather) {
       pushNotification({
-        title: `Could not run must gather for ${activeMustGather.customName}`,
+        title: `Could not run must gather for ${activeMustGather.displayName}`,
         message: '',
-        key: activeMustGather.customName,
+        key: activeMustGather.displayName,
         variant: 'danger',
         actionClose: true,
       });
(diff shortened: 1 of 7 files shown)


src/app/common/components/MustGatherModal.tsx
```diff
@@ -39,32 +40,31 @@ export const MustGatherModal: React.FunctionComponent = () => {
...
   const handleMustGatherRequest = ({ customName, type }: MustGatherObjType) => {
+    const namespacedName = withNs(customName, type);
     registerMustGather.mutate({
-      'custom-name': customName,
+      'custom-name': namespacedName,
       command:
         type === 'plan'
-          ? `PLAN=${customName} /usr/bin/targeted`
-          : `VM=${customName} /usr/bin/targeted`,
+          ? `PLAN=${namespacedName} /usr/bin/targeted`
```
mturley, line 48: Should this be the `customName` instead of `namespacedName`? Wouldn't the service need the bare plan/VM name?

src/app/common/components/MustGatherModal.tsx
```diff
@@ -39,32 +40,31 @@ export const MustGatherModal: React.FunctionComponent = () => {
   );
 
   const handleMustGatherRequest = ({ customName, type }: MustGatherObjType) => {
+    const namespacedName = withNs(customName, type);
     registerMustGather.mutate({
-      'custom-name': customName,
+      'custom-name': namespacedName,
```
mturley, line 45: The fact that `custom-name` gets set to `namespacedName` in the backend, but in the UI `customName` is a different non-namespaced thing, feels a bit awkward to me. We can leave it, but I think we might want to invert it. We could call your current `customName` something like `displayName` or `resourceName` or `planOrVmName` or something, and change your current `namespacedName` to `customName` so it matches what the backend calls `custom-name`?

I'm overthinking it again.

src/app/common/components/MustGatherModal.tsx
seanforyou23: The must gather service will use whatever you supply, it's not required to use any specific format. If we don't include the namespacing info we'll have to re-stitch the values back together elsewhere in the UI where we do lookups. If there's no issue with including the namespace, I think it may be better to keep it in lock step with what the backend receives.

src/app/common/components/MustGatherModal.tsx
seanforyou23: It's a good point though. Yea gimme a few and I'll make it more readable

mturley (approved):

PR #783, opened 2021-09-30
feat(vms): use minimal payload for vms query

seanforyou23:
This PR makes corresponding changes in the UI to support the slimmed down api payloads work (https://github.com/konveyor/forklift-controller/pull/343).

The vms payloads are cut in half and the nic/disk data is no longer duplicated across vm objects and instead fetched in similar manner to other service requests we make. Had to re-map the references to disks and nics, seems to be working but that's definitely the main thing to verify in review. Thx for all the help @mturley!

Related to [MTV-139](https://issues.redhat.com/browse/MTV-139)

### DIFF
--- a/src/app/Plans/components/Wizard/MappingForm.tsx
+++ b/src/app/Plans/components/Wizard/MappingForm.tsx
@@ -17,7 +17,6 @@ import {
 } from '@patternfly/react-core';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import { ValidatedTextInput } from '@konveyor/lib-ui';
-
 import { OptionWithValue } from '@app/common/components/SimpleSelect';
 import {
   MappingType,
@@ -29,7 +28,13 @@ import {
   SourceInventoryProvider,
 } from '@app/queries/types';
 import { MappingBuilder, IMappingBuilderItem } from '@app/Mappings/components/MappingBuilder';
-import { filterSharedMappings, useMappingResourceQueries, useMappingsQuery } from '@app/queries';
+import {
+  filterSharedMappings,
+  useMappingResourceQueries,
+  useMappingsQuery,
+  useDisksQuery,
+  useNicProfilesQuery,
+} from '@app/queries';
 import { PlanWizardFormState } from './PlanWizard';
 import {
   getBuilderItemsFromMapping,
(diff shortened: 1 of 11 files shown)


src/app/Plans/components/Wizard/MappingForm.tsx
```diff
@@ -63,16 +68,27 @@ const MappingForm: React.FunctionComponent<IMappingFormProps> = ({
...
+
   const { availableSources, availableTargets } = mappingResourceQueries;
 
   const hasInitialized = React.useRef(false);
   React.useEffect(() => {
-    if (!hasInitialized.current && mappingResourceQueries.status === 'success') {
+    if (
+      !hasInitialized.current &&
+      mappingResourceQueries.status === 'success' &&
+      sourceProvider?.type === 'ovirt' &&
```
mturley, line 89: This will prevent the effect from running when you're using a vsphere source provider. Instead we want something like `&& (sourceProvider?.type !== 'ovirt' || rhvResourcesLoaded)`

src/app/Plans/components/Wizard/helpers.tsx
```diff
@@ -351,32 +355,45 @@ export const filterSourcesBySelectedVMs = (
...
+  const sourceIds: (string | undefined)[] = Array.from(
     new Set(
       selectedVMs.flatMap((vm) => {
         if (mappingType === MappingType.Network) {
           if (sourceProviderType === 'vsphere') {
             return (vm as IVMwareVM).networks.map((network) => network.id);
           }
           if (sourceProviderType === 'ovirt') {
-            return (vm as IRHVVM).nics.map((nic) => nic.profile.network);
+            const networkIds = (vm as IRHVVM).nics.map((nic) => {
```
mturley, line 370: Nitpick, but if you want to be consistent with storages in terms of doing this in 2 steps for readability, you could do:

```ts
const vmNicProfiles = (vm as IRHVVM).nics.map(nic) =>
  nicProfiles.find((nicProfile) => nicProfile.id === nic.profile)
);
const networkIds = vmNicProfiles.map((nicProfile => nicProfile?.network);
return networkIds;
```

mturley (changes requested):
Looking good, but I noticed one problem and one nitpick.

src/app/Plans/components/Wizard/MappingForm.tsx
seanforyou23: Good catch!

mturley (approved):
🎉

Issue #782, opened 2021-09-30
Bearer token is undefined in must-gather-api request


aufi:
Fabien pointed me to not working Forklift UI (redirecting to/from welcome page periodically). That seems to be caused by 401 errors when loading the UI and must-gather API.

It looks there is ```Authorization: Bearer undefined``` header in must-gather requests. The must-gather-api pod seems to work okay when using its console from token from UI.

Also it would be nice to not "break" Forklift UI when must-gather-api is missing or not working (or returning 401).

Example from Chrome Inspector:
![image](https://user-images.githubusercontent.com/555381/135483148-082374be-4e3e-43e6-9ce7-8e758b6570fc.png)

aufi:
Both PR were merged! closing this issue.

PR #787, opened 2021-10-05
fix(must-gather): ensure MG command uses plan/vm name, ensure access_token exist

seanforyou23:
This PR makes two notable changes.

The first change is that we add a check to ensure the relevant access token exists before attempting to hit the must gather service endpoint.

The second change is to make sure that the `command` property for must gather references the actual plan/vm name and not one with any namespace prefix.

Steps toward closing https://github.com/konveyor/forklift-ui/issues/782

### DIFF
--- a/src/app/queries/fetchHelpers.ts
+++ b/src/app/queries/fetchHelpers.ts
@@ -25,12 +25,14 @@ export const authorizedFetch = async <TResponse, TData = unknown>(
   data?: TData
 ): Promise<TResponse> => {
   const { history, checkExpiry } = fetchContext;
+  const headersObj = {
+    Authorization: `Bearer ${fetchContext.currentUser.access_token}`,
+    ...extraHeaders,
+  };
+
   try {
     const response = await fetch(url, {
-      headers: {
-        Authorization: `Bearer ${fetchContext.currentUser.access_token}`,
-        ...extraHeaders,
-      },
+      headers: headersObj,
       method,
       ...(data &&
         method !== 'get' && {
--- a/src/app/AppLayout/AppLayout.tsx
+++ b/src/app/AppLayout/AppLayout.tsx
@@ -23,7 +23,6 @@ import './AppLayout.css';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import HelpDropdown from './HelpDropdown';
 import { Notifications } from '@app/common/components/Notifications';
-import { MustGatherModal } from '@app/common/components/MustGatherModal';
 
 interface IAppLayout {
   children: React.ReactNode;
@@ -150,7 +149,6 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({ children }) => {
     >
       <>
         {children}
-        <MustGatherModal />
         <Notifications />
       </>
     </Page>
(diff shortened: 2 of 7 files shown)


aufi:
I'm not a UI specialist, but looks good to me and works well (except download finished must-gather archive) for me :+1:

src/app/common/components/MustGatherBtn.tsx
```diff
@@ -31,12 +34,27 @@ export const MustGatherBtn: React.FunctionComponent<IMustGatherBtn> = ({ display
...
+            'blob'
+          )
+            .then((tarBall) => {
+              const file = new File([tarBall], mustGather['archive-name'], {
+                type: 'text/plain;charset=utf-8',
+              });
+              saveAs(file);
+            })
+            .catch((error) => {
+              console.error(error);
```
mturley, line 55: We may want to pop a push notification in this catch statement in case it fails to download the tarball or there is some exception in file-saver (malformed blob)

src/app/common/components/MustGatherWatcher.tsx
```diff
@@ -44,10 +46,26 @@ export const MustGatherWatcher: React.FunctionComponent<IMustGatherWatcherProps>
           <div>
             You can download the migration logs for the {type}&nbsp;
             <Button
-              component="a"
-              href={getMustGatherApiUrl(`must-gather/${data?.['id']}/data`)}
-              download
-              target="_self"
+              onClick={() => {
+                authorizedFetch<Blob>(
```
mturley, line 50: We could probably abstract these two fetch-then-save code snippets into a helper function, but it's not too much code so I'm fine either way.

src/app/queries/fetchHelpers.ts
```diff
@@ -22,23 +22,26 @@ export const authorizedFetch = async <TResponse, TData = unknown>(
   fetchContext: IFetchContext,
   extraHeaders: RequestInit['headers'] = {},
   method: 'get' | 'post' = 'get',
+  returnMode: 'json' | 'blob' = 'json',
```
mturley, line 25: Nice.

mturley (commented):
Looking good, two super minor things. If you disagree I'm fine with merging it.

src/app/common/components/MustGatherBtn.tsx
seanforyou23: Good call - done!

src/app/common/components/MustGatherWatcher.tsx
seanforyou23: Not a problem, I went ahead and created a few helper functions for fetching, downloading, and complaining about the must gather output file. I've realized an issue with the notifications for when there are many at a single time, but we can follow up on that outside of this PR.

aufi:
I've succesfully tested this PR (custom forklift-ui image build) with quay.io/konveyor/forklift-must-gather-api:latest \o/

Steps:
- open forklift-ui (with exiting providers and plans) - OK
- get logs for cancelled plan (I don't have currently any finished plan) - 10-20 seconds to get the download link, completed status - downloaded the archive, name contains plan - OK; checked its content - OK
- get logs for VM (actually not yet started plan which is not very useful since VM's CRs are not yet present) - same 10-20 seconds fot the download link with name including VM name, the archive doesn't contain vm-related CRs (since it doesn't exist), but few logs with the VM were present (nit: maybe hide the get logs button on new plan?) - OK

aufi (approved):

PR #797, opened 2021-10-07
Bug 2011785: Show Validating instead of Unknown for the first 30 seconds of missing status on a plan CR

mturley:
https://bugzilla.redhat.com/show_bug.cgi?id=2011785

Currently when the `plan.status` property is missing, we assume there is a problem and show Unknown in the status column on the plans table. It appears that in some circumstances this can be normal, if the controller is taking a few seconds to catch up and validate the newly created plan CR. This PR shows a spinner with "Validating" if there is no status data available for a plan at first.

Screen Shot 2021-10-07 at 12 19 01 PM

However, the status field can also be empty because of some unknown problem, and I didn't want it to be possible for a plan to get stuck showing "Validating" forever in that case. After 30 seconds of status data continuing to be missing, the message will change to "Unknown" with a warning icon as is displayed today.


mturley:
@vconzola are you ok with the word "Validating" being used here? that's what is happening when a plan is first created and not yet known to be Ready.

PR #800, opened 2021-10-07
fix(must-gather): ensure must gather btn is disabled until migrations finish

seanforyou23:
This PR adds a check to ensure that a completed property exist on the vm/plan before enabling the must gather button.

Screen Shot 2021-10-07 at 5 00 42 PM

### DIFF
--- a/src/app/Plans/components/PlanActionsDropdown.tsx
+++ b/src/app/Plans/components/PlanActionsDropdown.tsx
@@ -76,6 +76,9 @@ export const PlanActionsDropdown: React.FunctionComponent<IPlansActionDropdownPr
     !planState?.toLowerCase().includes('failed') &&
     !planState?.toLowerCase().includes('canceled');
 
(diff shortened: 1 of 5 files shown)


aufi:
I'm okay with this as far it allows download logs for Plans completed/finished in whatever state (success, failure, cancelation, ..). I don't have env to test all these combinations, so will trust you :+1:

apinnick:
@mturley  Is there any way to see the forklift-ui-preview with the "Get logs" button enabled?

mturley:
@apinnick unfortunately no, that API is a little hard to mock so it's not supported in preview mode right now. If you need screenshots I can send them, or I can find a cluster for you to look at

apinnick:
@mturley Thanks. I thought that might be the issue. No need. 

I wrote up a "flow" and asked Maayan to check it, since the final UI is different from Vince's mockups.

PR #810, opened 2021-10-19
Bug 2012795: feat(edit): allow editing for plans where no individual VM migration succeeded

seanforyou23:
This PR updates the logic around the PlansTable and PlanActionsDropdown to allow for editing a plan assuming none of the VMs have been migrated yet. We anticipate there may be some changes needed elsewhere in the UI, so I'm posting this as a draft first to work out any kinks.

### DIFF
--- a/pkg/web/src/app/Plans/components/PlanActionsDropdown.tsx
+++ b/pkg/web/src/app/Plans/components/PlanActionsDropdown.tsx
@@ -32,12 +32,14 @@ interface IPlansActionDropdownProps {
   plan: IPlan;
   planState: PlanState | null;
   canRestart: boolean;
+  numVMsDone: number;
 }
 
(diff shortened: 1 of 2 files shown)


mturley:
The complexity of this one that we still need to address is making sure all our status displays / lists of VMs in a plan still behave properly when the VM list in a plan spec has been edited and no longer matches the VM list in that plan's status (from the last migration). If that turns out to be a pain for us we could ask the backend guys if it's ok to require deleting existing migration CRs when editing a plan, if that would even help.

I may be anxious over nothing and it might be easy to manage. I think we use the status list as the source of truth for the VMs that exist in the migration details page, and we just need to use the spec instead and find the matching VMs from status when rendering? maybe?

seanforyou23:
Closing this for now, can revive the effort later if need be.

PR #831, opened 2021-11-11
Bug 2022651: Add explanation of leftover resources when deleting a non-archived plan, don't allow deleting while a plan is being archived

mturley:
This is the confirm modal we currently show when deleting a plan:
Screen Shot 2021-11-11 at 3 19 53 PM

With this PR, that text is still used for deleting any plans that have never been started or are fully archived. Otherwise (for plans that are done running but have not yet been archived), this is the new confirm modal:

Screen Shot 2021-11-11 at 3 20 12 PM

This PR also prevents triggering the delete action (even opening the modal) if the plan is currently in the process of being archived.

### DIFF
--- a/src/app/Plans/components/PlanActionsDropdown.tsx
+++ b/src/app/Plans/components/PlanActionsDropdown.tsx
@@ -152,7 +152,8 @@ export const PlanActionsDropdown: React.FunctionComponent<IPlansActionDropdownPr
             isTooltipEnabled={
               hasCondition(conditions, 'Executing') ||
               deletePlanMutation.isLoading ||
-              isPlanGathering
+              isPlanGathering ||
+              planState === 'Archiving'
             }
             content={
               hasCondition(conditions, 'Executing')
@@ -161,6 +162,8 @@ export const PlanActionsDropdown: React.FunctionComponent<IPlansActionDropdownPr
                 ? 'This plan cannot be deleted because it is deleting'
                 : isPlanGathering
                 ? 'This plan cannot be deleted because it is running must gather service'
+                : planState === 'Archiving'
+                ? 'This plan cannot be deleted because it is being archived'
                 : ''
             }
           >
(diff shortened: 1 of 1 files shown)


mturley:
@apinnick can you review this text change for us?

fdupont-redhat (approved):

apinnick:
@mturley 
I understand what you are saying but it might be confusing to the user. Normally, one does not need to archive something before deleting it in order to clean it up. If this is a permanent change, I think it should also be documented with a brief note in the section on "migration plan options" in the UI.

Also, I think the text could be a little more concise. I will add my suggestion in a comment on the file.

src/app/Plans/components/PlanActionsDropdown.tsx
```diff
@@ -229,7 +233,20 @@ export const PlanActionsDropdown: React.FunctionComponent<IPlansActionDropdownPr
         mutateResult={deletePlanMutation}
         title="Permanently delete migration plan?"
         confirmButtonText="Delete"
-        body={`All data for migration plan "${plan.metadata.name}" will be lost.`}
+        body={
+          isPlanStarted && !isPlanArchived ? (
+            <TextContent>
+              <Text>
+                Migration plan &quot;{plan.metadata.name}&quot; will be deleted. However, some
```
apinnick, line 240: I think this could be more concise. Suggestion:

Migration plan &quot;{plan.metadata.name}&quot; will be deleted. However, deleting a migration plan does not remove temporary resources such as failed VMs and data volumes, conversion pods, importer pods, secrets, or config maps.

To clean up these resources, archive the plan before deleting it.

mturley:
@apinnick thanks for the revision. I think this is a permanent change due to the limitations of CR deletion handling in k8s/OpenShift. There is a 2.3-targeted BZ for it here which sounds like it's going to be closed: https://bugzilla.redhat.com/show_bug.cgi?id=2018974

PR #829, opened 2021-11-01
Hooks Infra and New changes as per the UI requirement

kpunwatk:
Added Updated New Hook Infra for running QE UI hook Tests

PR #844, opened 2021-11-18
Duplicate Feature Test Case also with the Fix for Cancel_Restart

kpunwatk:


mturley:
@kpunwatk this PR includes the changes from your other PR https://github.com/konveyor/forklift-ui/pull/829. We should try to only include each commit in one PR so we don't have redundant changes to review and to reduce merge conflicts. 

* If the changes here depend on the changes in the other PR, we can leave this as-is for now but we should prioritize merging the first PR so we can rebase this one with a clean history.
* If the changes here do not depend on the other PR, you could cherry-pick only the new commits onto a fresh branch.

In either case, let me know if you'd like help with it!

kpunwatk:
@mturley You are right this PR should includes the changes from previous PR. Actually, I created this as Draft PR because Meanwhile I wanted Igor to go through the Duplicate Feature test code and needed his suggestion and help.
I don't think we will need this PR anymore (has new modified code is available) and also it doesn't contains the changes of the previous PR.
Will wait till the Previous PR is merged.
Thanks a lot ! Mike I will definitely let you know if I need any help :)

mturley:
Sounds good! Sorry to assume.

Issue #833, opened 2021-11-15
[BZ#1897276] Vsphere provider: Fetch and show certificate to get user validation


gildub:
https://bugzilla.redhat.com/show_bug.cgi?id=1897276

The vCenter certificate can be retrieved by the Express server through a specific route and returned to the client UI in order to present it to the end user for validation.
 
The reason for the latter approach is because JavaScript on the browser side is not designed to deal with certificate (it's handled by the browsers themselves). Although we have envisaged using client side approach such as using a [native version of TLS](https://www.npmjs.com/package/node-forge) meanwhile we'd still have to handle a CORS situation.

mturley:
This was resolved by #845.

PR #845, opened 2021-11-18
Bug 1897276: Vsphere provider creation/edition certificate validation

gildub:
Addresses https://github.com/konveyor/forklift-ui/issues/833

### DIFF
--- a/pkg/web/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
+++ b/pkg/web/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
@@ -9,6 +9,17 @@ import {
   Stack,
   Popover,
   FileUpload,
+  Spinner,
+  Text,
+  Checkbox,
+  DescriptionList,
+  DescriptionListTerm,
+  DescriptionListDescription,
+  DescriptionListGroup,
+  Title,
+  Panel,
+  PanelMain,
+  PanelMainBody,
 } from '@patternfly/react-core';
 import {
   useFormState,
@@ -21,7 +32,6 @@ import { SimpleSelect, OptionWithValue } from '@app/common/components/SimpleSele
 import {
   fingerprintSchema,
   hostnameSchema,
-  PRODUCT_DOCO_LINK,
   ProviderType,
   PROVIDER_TYPES,
   PROVIDER_TYPE_NAMES,
@@ -34,10 +44,11 @@ import {
   useCreateProviderMutation,
   usePatchProviderMutation,
   useClusterProvidersQuery,
+  useCertificateQuery,
 } from '@app/queries';
 
-import { IProviderObject } from '@app/queries/types';
 import HelpIcon from '@patternfly/react-icons/dist/esm/icons/help-icon';
+import { IProviderObject } from '@app/queries/types';
 import { QuerySpinnerMode, ResolvedQuery } from '@app/common/components/ResolvedQuery';
 import { useEditProviderPrefillEffect } from './helpers';
 import { LoadingEmptyState } from '@app/common/components/LoadingEmptyState';
@@ -129,7 +140,7 @@ const useAddProviderFormState = (
     vsphere: useFormState({
       ...sourceProviderFields,
       fingerprint: useFormField('', fingerprintSchema.required()),
-      fingerprintFilename: useFormField('', yup.string()),
+      isCertificateValid: useFormField(false, yup.boolean()),
     }),
     ovirt: useFormState({
       ...sourceProviderFields,
@@ -194,6 +205,10 @@ export const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModal
   const mutateProviderResult = !providerBeingEdited
     ? createProviderMutation
     : patchProviderMutation;
+
+  const [hostname, setHostname] = React.useState('');
+  const [isCertificateQueryEnabled, setCertificateQueryEnabled] = React.useState(false);
+  const { status, error, certificate } = useCertificateQuery(hostname, isCertificateQueryEnabled);
 
   return (
     <Modal
@@ -284,6 +299,14 @@ export const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModal
                     label={getLabelName('hostname', brandPrefix(providerType))}
                     isRequired
                     fieldId="hostname"
+                    inputProps={{
+                      onBlur: () => {
+                        fields.hostname?.setIsTouched(true);
+                      },
+                      onFocus: () => {
+                        fields.hostname?.setIsTouched(false);
+                      },
+                    }}
                   />
                 ) : null}
                 {fields?.username ? (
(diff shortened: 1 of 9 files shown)


src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
```diff
@@ -195,6 +199,39 @@ export const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModal
     ? createProviderMutation
     : patchProviderMutation;
 
+  const [hostname, setHostname] = React.useState(fields?.hostname ? fields.hostname.value : '');
+
+  function assertIsCertificate(
+    certificate: ITLSCertificate
+  ): asserts certificate is ITLSCertificate {
```
mturley, line 206: whoa!! I was not aware of this TypeScript feature, I really wish I had discovered this sooner! That's really nice. I think we could use it in a few places where we have logic based on providerType and mappingType with a lot of `as` assertions. Future cleanup :)

src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
```diff
@@ -195,6 +199,39 @@ export const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModal
...
+    assertIsCertificate(certificate);
+
+    return certificate;
+  };
+
+  const [isCertificateQueryEnabled, setCertificateQueryFlag] = React.useState(false);
+
+  const [isCertificateValid, setCertificateValid] = React.useState(false);
+
+  const { status, error, data } = useQuery<ITLSCertificate, Error>(
```
mturley, line 227: As you suggested on Slack, I do think we should probably move this stuff to a `useCertificateQuery` hook so we're not cluttering this component. `assertIsCertificate` and `getCertificate` could be moved into there. I think we should also have the query result as a single variable instead of destructuring it, so we can use `certificateQuery.status` instead of `status` so the usage is more readable below.

src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
```diff
@@ -279,14 +316,54 @@ export const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModal
...
+                  <>
+                    <ValidatedTextInput
+                      field={fields.hostname}
+                      label={getLabelName('hostname', brandPrefix(providerType))}
+                      isRequired
+                      fieldId="hostname"
+                      inputProps={{
+                        onFocus: () => setCertificateQueryFlag(false),
+                        onBlur: () => {
+                          setHostname(forms.vsphere.fields.hostname.value);
```
mturley, line 328: I think we can simplify things and reduce the amount of state we're using here.

* I think we should define `isCertificateValid` as a required field in the `vsphere` section of `useAddProviderFormState` rather than in its own `useState`. We may also be able to just remove the `fingerprint` and `fingerprintFilename` fields from there, but we'd need to pass the fingerprint as an additional param to `mutateProvider` (which comes from `useCreateProviderMutation` and `usePatchProviderMutation`).
* By default (without the custom `onBlur` here), the `fields.hostname.isTouched` boolean will not be true until the blur event happens (because that's what triggers the validation errors to appear). Currently the validation on the hostname field is broken here because you're not calling `setIsTouched` in your custom `onBlur`. But I think we can restore the default behavior by removing the custom `onBlur` and `isCertificateQueryEnabled` state entirely, because:
  * If we replace the `setCertificateQueryFlag(false)` in `onFocus` with `fields.hostname.setIsTouched(false)`, we can use `fields.hostname.isTouched` as our value for whether the query should be enabled.
* If we do that, we also don't need the `hostname`/`setHostname` state, because we already have `fields.hostname.value`.

So in short, I think all we really need is the `isCertificateValid` to be part of the form state (so `isFormValid` works properly), and to manage the `isTouched` value of the hostname field as our indicator of whether the query should run, and we can get rid of all the other new state you've introduced here. Everything below that uses `isCertificateValid` can use `fields.isCertificateValid.value`, and your `useCertificateQuery` can take `fields.hostname.isTouched` as an `enabled` param.

src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
```diff
@@ -299,47 +376,14 @@ export const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModal
                     fieldId="username"
                   />
                 ) : null}
-                {fields?.password ? (
+                {fields?.password && status === 'success' && isCertificateValid ? (
```
mturley, line 379: Same here.

mturley (changes requested):
Looking good, but there are a few state things we can clean up here.

Also, I wonder if we should be showing other certificate details in addition to just the fingerprint, like issuer and expiration date. That would give the user more info to go by to know if the cert is valid. @vconzola what do you think? I left a similar comment on slide 7 of the mockup. The cert details could be in a little info box and the checkbox could say something more generic like "I trust the authenticity of this certificate" if we are showing more details.

vconzola:
> Looking good, but there are a few state things we can clean up here.
> 
> Also, I wonder if we should be showing other certificate details in addition to just the fingerprint, like issuer and expiration date. That would give the user more info to go by to know if the cert is valid. @vconzola what do you think? I left a similar comment on slide 7 of the mockup. The cert details could be in a little info box and the checkbox could say something more generic like "I trust the authenticity of this certificate" if we are showing more details.

@mturley I really like that idea. I think it'd be more helpful to the user than just seeing the fingerprint. @gildub Lemme update the mockups this afternoon and send them to you.

mturley:
Another note: when it comes time to deal with editing, I think we just need to be sure to modify `useEditProviderPrefillEffect` to match the changes you make to `useAddProviderFormState` here (including making sure the hostname's `isTouched` gets set to true so the form fetches the certificate right away). Then I think we will need something that runs after the fingerprint is loaded (a useEffect, or maybe just an onSuccess param we pass to the certificate query?) with the following logic:
* if the fingerprint we loaded matches the fingerprint on the `providerBeingEdited`, we can check the validated box.
* else, we need to leave that box unchecked and display an alert that the fingerprint changed and the user should make sure they trust the new certificate.

mturley:
One more thought: @vconzola was mentioning that it might be weird to have the cert fetch on blur, because what if the user types and doesn't click away from the field because the rest of them are disabled, and they are confused about why they have to click away? We could keep the custom `onBlur` you have and have it clear/set a timeout (stored in a ref) so that it sets `isTouched` when they stop typing for X amount of time rather than when they blur the field.

mturley:
For reference, these are the default text field props that you have to replicate if you override them with additional custom behavior. We should document that better in lib-ui maybe.

```ts
export const getTextFieldProps = (
  field: IValidatedFormField<string> | IValidatedFormField<string | undefined>
): Pick<TextInputProps | TextAreaProps, 'value' | 'onChange' | 'onBlur' | 'validated'> => ({
  value: field.value,
  onChange: field.setValue,
  onBlur: () => field.setIsTouched(true),
  validated: field.isValid ? 'default' : 'error',
});
```

gildub:
In reply of your latest comment in regards of `getTextFieldPropsdo` we have to override all the props?

mturley:
> In reply of your latest comment in regards of `getTextFieldProps`, do we have to override all the props?

No, those are just the default values provided by `ValidatedTextInput`, so if you do override `onBlur` for example you need to make sure you are calling `field.setIsTouched(true)` because it will no longer happen for you automatically. I think maybe this indicates a flaw in my component 😆

src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
gildub: @mturley, 

Thank you for your thorough review.
With latest commit is as close I could get to the expected behavior. 
The reason I keep using a state hostname is to make benefit of having a button triggering the fetch and avoid trying to fetch the certificate at each key stroke when hostname is entering.

gildub:
@vconzola, please review.

gildub:
@mturley, I have no idea why but following line fails the tests even though the `hostname` field contains a valid url:
https://github.com/konveyor/forklift-ui/pull/845/files#diff-a4f147238f961def753e901eb8e5ef56a8a77abd05ecf5f3595e66e25cb3d222R343

mturley:
@gildub that's quite strange. Looking into it now

mturley:
See slack for my findings :)

gildub:
@vconzola, per your question on Slack, 

> Not sure there’s anything we can do about this, but the Verify button is hidden below the fold when either the Add or Edit modals opened. Can we auto-scroll the form when the user clicks into either the user name or password field?

Moving the focus between different parts of the modal could be done but what about moving the hostname to be after the other input fields?
This way it's just before the "verify certificate" and then the certificate information and "Validate  certificate" checkbox are next to each others.
Latest commit shows such approach. Please let me know what you think.
If that feel wrong then we can try moving focus or look at other solutions such as a dedicated pop-up for the certificate.

gildub:
Having the hostname is really weird and unusual. So I reverted it.
Let's discuss further.

mturley:
Hey Gilles. I pushed a few commits here, feel free to revert if you need to, but I had it working in troubleshooting so I figured I may as well commit.

* Fixed the verify button enable issue in edit mode by calling `setIsTouched` on the hostname field when prefilling the form
* Replaced destructuring with a single variable for `certificateQuery` to make later code more readable (having variables simply called `status` and `data` was unclear imo)
* Added a warning when editing a provider if the loaded fingerprint is different from the one stored in the provider
* Per @vconzola's request, made the modal scroll down to show the verify button if it is offscreen when focusing the username/password fields

gildub:
Mike, 

That's great, no need to revert ! 

Thanks for troubleshooting the validation field issue. Fixing it will be great although we'll a fair amount of testing to do ;)
The rest is nice and makes sense.

I added a patch to mock `scrollIntoView` to be able to pass the tests. This is because `scrollIntoView` is missing from jsdom: https://github.com/jsdom/jsdom/issues/1695

mturley:
@seanforyou23 @gildub are we happy with this?

gildub:
@mturley, I'm happy with this now.

mturley (dismissed):

pkg/web/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
```diff
@@ -301,44 +330,105 @@ export const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModal
...
+                              }
+                            } else {
+                              fields.fingerprint?.setValue('');
+                            }
+                            fields.isCertificateValid?.setValue(!fields.isCertificateValid.value);
+                          }}
+                        />
+                      </>
+                    ) : (
+                      <Text>Error: Cannot retrieve certificate</Text>
```
seanforyou23: Might be good to use an [`<Alert />`](https://www.patternfly.org/v4/components/alert/#static-live-region-alert) here.

seanforyou23 (commented):
Looking good! The only thing I was wondering about is the error path. I noticed that when I give bad credentials the standard react query 3 retries with increasing backoff ends up going for minutes before an error state is produced in the UI. Maybe we should lower the number of retries, or reduce the retry time/delay so that if the UI fails to fetch the certs it will go ahead and report it? 

Last question - I don't suppose it's worth putting a retry button in case the fetch attempt fails? User just needs to close and reopen the modal, but a re-attempt button could be a nice little enhancement. This can totally be done in follow up though.

pkg/web/src/app/queries/providers.ts
```diff
@@ -377,3 +379,34 @@ export const getProviderNameSchema = (
...
+
+function assertIsCertificate(certificate: ITLSCertificate): asserts certificate is ITLSCertificate {
+  if (!('fingerprint' in certificate)) {
+    throw new Error('Not certificate');
+  }
+}
+const getCertificate = async (hostname: string) => {
+  const response = await fetch(`get-certificate?url=${hostname}`);
+  if (!response.ok) {
+    throw new Error('Problem fetching data');
```
seanforyou23, line 391: Could we forward an error from the backend here?

pkg/web/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
```diff
@@ -301,44 +331,112 @@ export const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModal
...
+                        Verify certificate
+                      </Button>
+                    ) : certificateQuery.status === 'loading' ? (
+                      <div className="pf-c-empty-state__icon">
+                        <Spinner
+                          aria-labelledby="loadingPrefLabel"
+                          size="sm"
+                          className={spacing.mrSm}
+                        />
+                        Retrieving SHA-1 certificate fingerprint
```
seanforyou23, line 363: This could maybe also be an [inline plain](https://www.patternfly.org/v4/components/alert#inline-plain-types) static live region info alert, if you wanted to communicate to screen reader users the loading state - since this one can take a while to resolve if there are issues with the network request.

seanforyou23 (approved):
LGTM - and nice job with the test suite and mock mode support. That extra effort really goes a long way!

mturley:
Discussing things with Mike S, I pushed a few more commits to surface error messages from fetching the certificate all the way to the UI. I think we can merge this and follow up in another PR perhaps, but the main thing remaining here I think is that if the certificate request fails, the button should reappear so it can be retried.

mturley (approved):
Good to merge, we can address the above in a followup.

mturley:
Followup issue: https://github.com/konveyor/forklift-ui/issues/868

mturley:
Hmm... CI is failing on `createEvent` here again. Not sure why

mturley:
@seanforyou23 @gildub I'm pretty baffled as to why this error is coming up when running tests in CI but not locally. Any ideas?

gildub:
@mturley, yeah I know, really frustrating I already looked at it and I've no idea why.

gildub:
@mturley, I've started investigation...

gildub:
@mturley, 

I was able to reproduce CI error locally using Ubuntu image and following same pipeline.
Meanwhile I'm also getting the same error when running CI locally on my machine using Nodejs 16 LTS.
So maybe my env was out of sync (forgot to run `npm install`).
So the positive side is that if anything like that (a CI mismatch) occurs again then we can simply run the Ubuntu env to compare. Although there is no reason this should happen because Ubuntu or Fedora or <place another OS here> are all running on same Node.js LTS latest version. 

Also I worked around the error itself by using scrollIntoView with `React.useRef()`.

mturley:
Good find, thanks @gildub ! The ref solution is interesting, I guess that just means something is different about that global document in jsdom or something like that. Glad it's working!

PR #846, opened 2021-11-19
Treeview: expands only top level

gildub:
- Only first level children are expanded by default
- Added a distinct React `key` to TreeView to avoid tree behavior overlap between Cluster and Folder views. For example (without a key), expend/collapse one tree would also expend/collapse the other one.

Follow up from https://github.com/konveyor/forklift-ui/issues/35 (which becomes obsolete after this one)

### DIFF
--- a/src/app/Plans/components/Wizard/helpers.tsx
+++ b/src/app/Plans/components/Wizard/helpers.tsx
@@ -170,6 +170,7 @@ const convertInventoryTreeNode = (
       node.kind === 'Cluster' ? <ClusterIcon /> : node.kind === 'Folder' ? <FolderIcon /> : null,
     customBadgeContent: badge,
     hasBadge: !!badge,
+    defaultExpanded: false,
   };
 };
 
@@ -235,6 +236,7 @@ export const filterAndConvertInventoryTree = (
       ),
       customBadgeContent: badge,
       hasBadge: !!badge,
+      defaultExpanded: true,
     },
   ];
 };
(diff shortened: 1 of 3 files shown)


src/app/Plans/components/Wizard/helpers.tsx
```diff
@@ -235,6 +236,7 @@ export const filterAndConvertInventoryTree = (
       ),
       customBadgeContent: badge,
       hasBadge: !!badge,
+      defaultExpanded: true,
```
mturley, line 239: Ha! I didn't realize it would be so simple.

mturley (approved):
Fantastic! Thanks @gildub .

mturley:
Ah, but I'm not sure this resolves #35 exactly. The default state works great, but the "All datacenters" toggle no longer expands all its descendants (which is correct, I think that was the wrong behavior and it was caused by `defaultAllExpanded`). But I think the issue calls for a separate "Expand all" toggle that will toggle the `allExpanded` prop and cause all descendants of the tree to expand. @vconzola thoughts?

Maybe we should merge this but just remove the "fixes #35" text, and follow up with another PR?

src/app/Plans/components/Wizard/helpers.tsx
gildub: Yeah once the `defaultAllExpanded` and `allExpanded` props are out of the picture its easy ;)

gildub: But devil hides in details, as you mentioned this doesn't solve the initial requirement.

gildub:
@mturley,
you're right, there is more work to be done.
Also it breaks the tests because the children is not visible.
Let's me try to do better.

vconzola:
@mturley As I said back in September, I'm not sure we want an expand all for the tree because it doesn't scale well for large trees. I played around with demo and it behaves as expected, though I wasn't exactly sure what's changed.

mturley:
Ah right. Ok, then I suppose we can just close the issue.

The change in this demo is that the tree defaults to only one level expanded instead of fully expanded.

gildub:
I clarified the change in the description and also reference this PR in the aforementioned issue.

Issue #55, opened 2020-09-17
Replace @patternfly/react-icons imports with direct deep-path imports to reduce bundle size


mturley:
According to @redallen we should be importing our icons like this:

```ts
import CogIcon from '@patternfly/react-icons/dist/js/icons/cog-icon';
```

instead of:

```ts
import { CogIcon } from '@patternfly/react-icons';
```

As-is, we're bundling and shipping all of those SVG icons in our app no matter which ones we use.

We should also add https://github.com/patternfly/patternfly-react/tree/master/packages/eslint-plugin-patternfly-react to catch these issues in the future.

gildub:
@mturley,
The recommend way doesn't work.
I tried with current @patternfly/react-icons version and latest 4.7.11.
I've created an issue on Pattern Fly project: https://github.com/patternfly/patternfly-react/issues/5028

mturley:
Thanks @gildub . Not sure what changed, I think this way of importing was working for a few icons in mig-ui.

mturley:
Also I agree it would be optimal to continue importing them the way we are, but we definitely don't want to bundle the entire icons package with our apps. They bundle the entire Font Awesome collection in there, it's huge.

gildub:
@mturley, following up on this issue, 

Using the shaking tree now works for `@patternfly/react-icons`, adding a PR to address those.

I was wondering too about the bundle size from other icons we import from `@patternfly/react-core` and `@konveyor/lib-ui`.
For example `src/app/NotFound.tsx` uses `EmptyStateIcon` from react-core.
Or `StatusIcon` from lib-ui.
I don't think there are issues because they're imported as React components.

mturley:
`StatusIcon` from lib-ui should probably be refactored to use the tree-shakeable imports itself as well. I didn't realize `EmptyStateIcon` was coming from react-core, if that's available in react-icons maybe we should be importing it from there instead.

mturley:
Oh, I forgot about `EmptyStateIcon`, that is also not an icon itself but a React component that renders an icon in a container (similar to lib-ui's `StatusIcon`). As you say it's not an issue.

PR #847, opened 2021-11-19
Import react icons specifically for tree shaking bundle

gildub:
- Imports every Patternfly icons file individually
- No sources maps for production (reduce build from 11Mb to 7.4Mb)
- Moves `css-minimizer-webpack-plugin`" to devDependencies
- Changes `AddEditProviderModal.test.tsx` from `findByRole()` to `getByLabelText()` to catch select due to a change in PF

Fixes https://github.com/konveyor/forklift-ui/issues/55


src/app/common/components/ValidatedPasswordInput.tsx
```diff
@@ -7,7 +7,8 @@ import {
   TextInput,
   TextInputProps,
 } from '@patternfly/react-core';
-import { EyeIcon, EyeSlashIcon } from '@patternfly/react-icons';
+import EyeIcon from '@patternfly/react-icons/dist/esm/icons/eye-icon';
+import EyeSlashIcon from '@patternfly/react-icons';
```
seanforyou23, line 11: I think this one can be just like the others, maybe the linter bit ya?

`import EyeSlashIcon from '@patternfly/react-icons/dist/esm/icons/eye-slash-icon';`

seanforyou23 (commented):
Looks good! Only thing I was sort of confused by was that the size of dist didn't seem to change much. I get ~11M with this branch and same on main. Is it expected?

seanforyou23:
I played around with it a bit, and found that if you add the `optimize-css-assets-webpack-plugin` pkg, then attach an instance to the production optimization configurations like `new OptimizeCSSAssetsPlugin({})` it seems to work great with the rest of these changes and drops the bundle size to ~9.5M

src/app/common/components/ValidatedPasswordInput.tsx
gildub: I left that one aside because it initially didn't work for me but I must have done a typo because it's working now.
Thanks for insisting! ;)

gildub:
@seanforyou23, sounds really good, do we want to add those to this PR or a separate one?

seanforyou23:
Yea man if you're cool with it I'd say go for it! In case there's any issues with these perf enhancements working in tandem it could make it easy to track it down. Either way works for me though :) Thanks for doing this!

seanforyou23:
For the test suite, there are probably other ways to fix it, but I think you could safely account for them in the moduleNameMapper config setting with something like below in case it helps.

```sh
'@patternfly/react-icons/dist/esm/icons/': '<rootDir>/__mocks__/fileMock.js'
```

mturley:
If we're messing with the CSS build process we should make sure to double-check that all our stuff still looks right. I've seen things like this happen a few times: https://github.com/patternfly/patternfly-react/issues/5650 (borders disappeared from Select menus)

gildub:
@seanforyou23, I left `optimize-css-assets-webpack-plugin` aside because I would like to avoid any side effects for this PR and isolate them in a separate PR.

Using only direct import for PF icons doesn't seem to reduce much the final build size of dist.
By just adding `nosources-source-map` the build is reduced from 11Mb to 7.4Mb.

mturley (approved):
LGTM, thanks @gildub!

Do you want to open a separate PR for `optimize-css-assets-webpack-plugin` in the short term, or should we open an issue for it so we don't forget?

PR #848, opened 2021-11-22
Use css-minimizer-webpack-plugin for optimizing bundle

gildub:
This is a follow up from https://github.com/konveyor/forklift-ui/pull/847.

This reduces the bundle to 7.2 MB (from 7.4).

I believe we can keep using `css-minimizer-webpack-plugin` instead of restoring `optimize-css-assets-webpack-plugin` to bundle css for production. Effectively the former is more accurate and works by default since Webpack 5. 

The original issue (see https://github.com/konveyor/forklift-ui/pull/646) is related to https://github.com/cssnano/cssnano/issues/675 and the work around for either plugin is to set `mergeLonghand: false` to `cssnano` processing (The flag to be  removed once addressed).

### DIFF
--- a/config/webpack.prod.js
+++ b/config/webpack.prod.js
@@ -4,20 +4,11 @@ const common = require('./webpack.common.js');
 const { stylePaths } = require('./stylePaths');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const TerserJSPlugin = require('terser-webpack-plugin');
+const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
 
 module.exports = merge(common('production'), {
   mode: 'production',
   devtool: 'nosources-source-map',
-  optimization: {
-    minimizer: [new TerserJSPlugin({})],
-    sideEffects: true,
-  },
-  plugins: [
-    new MiniCssExtractPlugin({
-      filename: '[name].css',
-      chunkFilename: '[name].bundle.css',
-    }),
-  ],
   module: {
     rules: [
       {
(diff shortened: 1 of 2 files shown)


gildub:
@seanforyou23, @mturley. 

This is 500KB short from the 6.9MB mentioned in the related PR.
Not sure if that worth moving back to `optimize-css-assets-webpack-plugin`.
What do you think?

mturley:
Wow, that was fast.

Honestly, as long as our CSS output is accurate, the dev experience is the same and everything looks right in prod, I don't really have any strong opinions about our webpack plugins. I'm curious which package is more prominently used by the community (sounds like `css-minimizer-webpack-plugin`?)

IMO the accuracy and stability outweighs a savings of 500KB, but I'm fine with either.

@gildub do you have other work that will depend on this one or are you ok with leaving it open until @seanforyou23 is back?

gildub:
@mturley, 

Thanks for the prompt feedback!
It was fast 'cause I fiddled around it in previous PR ;)
No deps upon it. Can definitely wait.

seanforyou23 (commented):
These changes look good to me! And I agree - don't have a big preference on which minimizer strategy we use, just that it works and doesn't break styles. I ran these changes locally and don't see any styling issues and we drop the distribution size from 7.3 to 7.1 M 😎

The only other thing I'd say may be worth doing as part of this work is to update the browsers list (`npx browserslist@latest --update-db`), the one we have currently is pretty old which can add to the size of required polyfills.

Screen Shot 2021-11-29 at 11 36 11 AM

mturley (approved):
LGTM!

PR #853, opened 2021-12-08
Convert pkg/api to Typescript

gildub:
- Migrates `pkg/api/src/*` and `pkg/api/scripts/remote-dev.js` to Typescript.
- Set `pkg/api/dist` as build target for `server/api`
- Add `npm config set unsafe-perm true` to Dockerfile to avoid mkdir issue during image build phase, i.e:
`12/9/2021, 9:04:28 AM error TS5033: Could not write file '/opt/app-root/src/pkg/api/dist/scripts/remote-dev.js': EACCES: permission denied, mkdir '/opt/app-root/src/pkg/api/dist'`

The dockerfile workaround is also implying https://github.com/konveyor/forklift-ui/issues/857 as when building locally the `unsafe-perm true` is not required.

fixes https://github.com/konveyor/forklift-ui/issues/760


gildub:
@mturley, @seanforyou23, I believe this is ready. 
@fabiendupont comment regarding issue#857 is reassuring as the user used during second build matters so we're good to go.

mturley (approved):
LGTM. @seanforyou23 what do you think?

seanforyou23 (approved):
😎

PR #858, opened 2021-12-10
Resource observability

seanforyou23:
This PR adds support for tailing pod logs from within forklift-ui. The code side isn't quite ready for review (though high-level comments certainly welcome 🙂) but I wanted to get a draft PR up so people could feedback on the UX of this feature since we're still nailing that bit down.

The log viewer is rendered currently as a top level nav item, ignore this - I haven't figured out where to place it just yet. Any specifics on how to handle loading states would be much appreciated!


seanforyou23:
Just realized mock mode is totally busted so unless you're running it against a real backend don't expect to see anything just yet. Working on this now.

seanforyou23:
Mock mode for log viewer is now working. I'm still thinking about how to handle the loading states, and will start working on where exactly we want to place this component within the app.. but it's ready for a round of review when anyone finds time.

seanforyou23:
Closing this for now, can always revive the effort later if need be.

Issue #713, opened 2021-07-20
Upgrade to PatternFly's 2021.09 release as soon as it is stable (available in the `latest` tag)


mturley:
https://github.com/konveyor/forklift-ui/pull/712 installs prerelease versions of PatternFly packages. Just to be safe, we should get back to the stable branch of PF as soon as it includes the fixes we need.

mturley:
Closing #712 for now, we decided to wait for the next stable release anyway. So, resolving this issue will also resolve #256 and #553.

PR #863, opened 2021-12-20
Bug 1992982: fix(dropdown): persist target namespace input filter results

seanforyou23:
This PR fixes an issue where we had a disconnect with the way the namespace selection dropdown behaved, where text was retained but filter state was not. Now the two are in sync, and blurring then refocusing the dropdown gives the same results as it did when you initially type into it.

Most of this work was done upstream so the fix here was a one-liner ;) 

https://bugzilla.redhat.com/show_bug.cgi?id=1992982

I also fixed up the build compilation stats analyzer, it had stopped working at some point. Good to go now.

@mturley can we also close https://github.com/konveyor/forklift-ui/issues/713 now?


mturley (approved):
Yeah I think we can! Thanks @seanforyou23 !

PR #865, opened 2021-12-20
Fix fallback path for meta.dev.json file when META_FILE environment var is undefined

mturley:
Currently when running remote dev mode without an explicit `META_FILE` in your environment, the script attempts to look for a `meta.dev.json` in the repo root rather than in the config directory.


seanforyou23 (approved):
Good catch!

PR #871, opened 2021-12-22
Revert "Convert pkg/api to Typescript (#853)"

mturley:
This reverts #853.

This change appears to have broken Cachito behavior during builds. Until we can determine a better fix, reverting this so we don't block 2.3 builds while some of us are away for year-end recharge.

@gildub after the holidays, let's revisit #853 and try to implement it again with some more scrutiny and make sure we involve Franco and make sure end-to-end builds are still working.

cc @seanforyou23 @fbladilo 

Cachito debug log apparently caused by the change:

```
[2021-12-22 17:56:35,350 cachito.workers.tasks.utils DEBUG utils.get_request_state] Getting the state of request 187413
[2021-12-22 17:56:35,460 cachito.workers.tasks.general INFO general.fetch_app_source] Fetching the source from "https://github.com/konveyor/forklift-ui" at reference "a22e5350ab8964550ce67dec26a8a8d50a244c0e"
[2021-12-22 17:56:35,460 cachito.workers.tasks.utils INFO utils.set_request_state] Setting the state of request 187413 to "in_progress" with the reason "Fetching the application source"
[2021-12-22 17:56:35,595 cachito.workers.scm DEBUG scm.repo_name] Parsed the repository name "konveyor/forklift-ui" from https://github.com/konveyor/forklift-ui
[2021-12-22 17:56:35,595 cachito.workers.paths DEBUG paths.__new__] Ensure directory /var/lib/cachito/sources/konveyor/forklift-ui exists.
[2021-12-22 17:56:35,597 cachito.workers.scm DEBUG scm.fetch_source] The archive already exists at "/var/lib/cachito/sources/konveyor/forklift-ui/a22e5350ab8964550ce67dec26a8a8d50a244c0e.tar.gz"
[2021-12-22 17:56:35,597 cachito.workers.scm DEBUG scm._verify_archive] Verifying the archive at /var/lib/cachito/sources/konveyor/forklift-ui/a22e5350ab8964550ce67dec26a8a8d50a244c0e.tar.gz
[2021-12-22 17:56:36,302 cachito.workers.paths DEBUG paths.__new__] Ensure directory /var/lib/cachito/bundles/temp/187413 exists.
[2021-12-22 17:56:36,302 cachito.workers.paths DEBUG paths.__new__] Ensure directory /var/lib/cachito/bundles/temp/187413/deps exists.
[...]

PR #849, opened 2021-11-26
Tooling for Monorepo

gildub:
- Bumps Nodejs to Latest LTS Version 16.13.0 with npm 8.1.0.
- Creates the following workspaces:
  - forklift-ui-web: `/src` is now `/pkg/web/src`
  - forklift-ui-api: `/deploy` is now `pkg/api/src` and `/scripts` is now `pkg/api/scripts`
  - forklift-ui-qe-tests: `/qe-test` is now `pkg/qe-tests`
- Adds a new package.json to each workspace
- Adds stub for Typescript project references but doesn't leverage it yet.

Notes: 
- Optionally we could spare the `/pkg` folder and have the "workspaces" `api`, `web` and `qe-tests` sitting directly at the top level meanwhile it has the advantage of grouping them in one place. To be discussed.

Helps https://github.com/konveyor/forklift-ui/issues/760

PR #875, opened 2021-12-23
Temporarily remove workspaces and move all dependencies back to the root to fix Cachito

mturley:
Follow-up to https://github.com/konveyor/forklift-ui/pull/871 because Cachito still doesn't like our workspaces. This is a partial revert of changes in https://github.com/konveyor/forklift-ui/pull/849.

This PR removes the package.json files under each workspace, and moves all the dependencies specified there back to the root package.json and updates the scripts there to reference the right paths. It does NOT, however, undo the other changes in https://github.com/konveyor/forklift-ui/pull/849 such as moving files (everything is still under `pkg/*`) or the Node version upgrade.

After break we should try to figure out how to work around Cachito's issues with npm workspaces and revert both this and #871, but we may just need to wait to hear back from that team before we reinstate these changes so we don't block QE's builds for 2.3 testing.

It's possible that the workspaces / TS conversion will need to wait until after 2.3 considering what is left to implement, my responsibilities to MTC/MTX, and the fact that we're losing @seanforyou23 to paternity leave.


mturley:
Self-merging to unblock QE during recharge, since everyone has left. We can revert after if needed.

PR #883, opened 2022-01-06
Bug 1994179: Add support for custom ingress CA chain in OpenShift

fabiendupont:
In OpenShift, it is possible to configure a custom certificate for the ingress routes. The full CA chain is provided to allow certificate verification. By default, the custom CA chain is not exposed in `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`, so pods cannot reach services on their external URL.

The UI uses a callback link with the external URL of the UI, since it will then be passed to the user's browser that can only access external URLs. This forbids the access to the UI when using custom CA.

With konveyor/forklift-operator#201, we ensure that the custom CA is merged into the main CA bundle and mounted in `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`. This change checks
the existence of this file and uses it as the main CA bundle to build the final CA bundle used by NodeJS.

Signed-off-by: Fabien Dupont <fabiendupont@pm.me>

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,8 +1,17 @@
 #!/usr/bin/env bash
 
-# Add service serving CA cert to the global CA bundle
-if [ -f "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt" ]; then
+# Copy the main CA bundle to /opt/app-root/src/ca.crt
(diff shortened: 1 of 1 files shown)


fbladilo (approved):
LGTM

mturley (approved):

PR #884, opened 2022-01-09
Added changes required for running tests from Jenkins

ibragins:
Now external json config file should be used (added such files to `.gitignore`)
Added changes to the config file to take parameters from that JSON
Increased timeout for VMWare cluster list load cause sometimes it works really SLOWLY
Added fix for hooks cause previously it was trying to add hooks even when they were not defined


gildub (approved):
LGTM

gildub (approved):
LGTM

PR #887, opened 2022-01-10
Fix Cannot read property 'createEvent' of null

gildub:
Initially we thought using `scrollIntoView` wasn't playing nicely with our testing libraries and we had to find a solution and in the meantime we remove the feature.

It turned out that it was an async/await issue with userEvent() from react-testing-library (it's community channel helped: https://stackoverflow.com/questions/60504720/jest-cannot-read-property-createevent-of-null).

So finally we don't have to remove `scrollIntoView` and the random error has disappeared where the only way to make sure of it is to run the tests in a loop.

AddEditProviderModal.test.tsx and PlanWizard.test.tsx are the two files involved.

This also:
- adds "eslint-plugin-testing-library" library to help with developments.
- Bump testing libraries to latest

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -32,7 +32,7 @@
       "version": "^16.11.0"
     }
   },
-  "plugins": ["prettier", "@typescript-eslint", "react-hooks", "eslint-plugin-react-hooks"],
+  "plugins": ["prettier", "@typescript-eslint", "react-hooks", "eslint-plugin-react-hooks", "testing-library"],
   "rules": {
     "@typescript-eslint/explicit-module-boundary-types": "off",
     "@typescript-eslint/explicit-function-return-type": "off",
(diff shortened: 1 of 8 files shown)


gildub:
I get the error only when testing the whole using `npm run ci`
There are no errors when testing only `AddEditProviderModal`:
```
cd pkg/web/; DEBUG_PRINT_LIMIT=10000 DATA_SOURCE=mock ../../node_modules/jest/bin/jest.js --rootDir=. --config=../../config/jest.config.js src/app/Providers/components/AddEditProviderModal/__tests__/AddEditProviderModal.test.tsx; cd -
```

mturley:
Ready to go @gildub? or did you still want to add anything?

mturley (approved):

PR #891, opened 2022-01-17
Bug 2041351: Remove all queries and logic based on the old Provisioner CRD

mturley:
Following the removal of the Provisioner CRD in the backend, this PR removes all references to it in the UI. The one major impact of this is that we no longer have a way to display the following warning when selecting target storage classes for a mapping:

Screen Shot 2022-01-17 at 1 39 26 PM

This PR simply removes that warning, so we may want to follow up to reintroduce a similar warning later (https://github.com/konveyor/forklift-ui/issues/892). Relevant discussion with @mansam on Slack:

Screen Shot 2022-01-17 at 1 41 53 PM


gildub (approved):
LGTM

mturley:
Thanks Gilles!

PR #907, opened 2022-02-06
Updated VM to be an instance instead of simple string

ibragins:


yaacov:

PR #909, opened 2022-02-07
Bug 2051508: Add field for VDDK init image to vmware provider form

mturley:
Adds a required form field "VDDK init image" when creating or editing a VMware provider. This value is initially blank for the first VMware provider being created, but will prefill itself from the last used value if another VMware provider is present.

Had to modify the prefill hook logic so it will run both in create and edit modes, since this initial value depends on a query.

Screen Shot 2022-02-07 at 12 24 55 PM


vconzola (approved):
Looks good.

jortel (approved):
lgtm
Visual review with limited UXD understanding.

PR #917, opened 2022-02-16
Correct misspelled word 'Analysing' -> 'Analyzing'

mturley:
cc @gildub


mturley:
@vconzola do you mind giving this an approval since it's a trivial text change, I can go ahead and merge if you agree with the spelling (looks like both are valid but "Analyzing" is more common, I saw someone complain about this on GChat)

vconzola (approved):
LGTM.

PR #927, opened 2022-02-28
Bug 2018983: Ensure k8s names will not be longer than 63 characters when configuring hosts

mturley:
When we configure the migration network for vsphere hosts, we create Secret and Host CRs with names based on the provider name and the host id. If the provider name is very long, as in the BZ example of `vsphere-6.5-rhev-node-05.rdu2.scalelab.redhat.com`, the additional characters of the host id and the `generateName` suffix can put the Secret and/or Host CR name over the 63 character limit for k8s resource names.

This PR adds a `truncateK8sString` helper function which can be used to limit these names to 63 characters while preserving the desired suffix, by truncating the prefix (provider name, in this case). In the case where it is used for a `generateName` field, an additional 5 characters are shaved off to leave room for the generated suffix.


gildub (approved):
LGTM

PR #939, opened 2022-03-04
[QE] Rhv Provider runWizard method

kpunwatk:
modified:   pkg/qe-tests/cypress/integration/models/providerRhv.ts


ibragins (approved):
LGTM

PR #940, opened 2022-03-07
Added some python code.

ibragins:
The task is to parse tarball that was downloaded by getLog and make sure that the content is valid


mturley (approved):
Nice! LGTM

PR #913, opened 2022-02-11
[QE] Added Method Schedule Cutover later and Incremental copies

kpunwatk:
@ibragins Please review this PR! This contains the schedule cutover for later option method and also incremental copies method which waits till the VM completes it's first precopy and status display Idle - Next incremental copy will begin in less than 1 minute.
        modified:   pkg/qe-tests/cypress/integration/models/plan.ts
	modified:   pkg/qe-tests/cypress/integration/types/types.ts
	modified:   pkg/qe-tests/cypress/integration/views/plan.view.ts

PR #942, opened 2022-03-15
[QE] Fixes for warm migration automation

kpunwatk:
1. Removed time-period as an assertion from the text of incremental copy status 
2. In config files added scheduled date and time parameters for both vmware and rhv providers. 
3. Remove "Plan.openList();" extra line from the restart method

modified:   pkg/qe-tests/cypress/integration/models/plan.ts
	modified:   pkg/qe-tests/cypress/integration/tests/Rhv/config_separate_mapping_rhv.ts
	modified:   pkg/qe-tests/cypress/integration/tests/vmware/config_separate_mapping.ts

### DIFF
--- a/pkg/qe-tests/cypress/integration/models/plan.ts
+++ b/pkg/qe-tests/cypress/integration/models/plan.ts
@@ -348,7 +348,8 @@ export class Plan {
       cy.get(`[aria-label="Select row ${i}"]`, { timeout: 30 * SEC })
         .closest(trTag)
         .within(() => {
-          cy.get('[data-label="Status"]', { timeout: 3600 * SEC }).should(
+          //VM's completes the 1 precopy
+          cy.get('[data-label="Status"]', { timeout: 6000 * SEC }).should(
             'contain.text',
             'Idle - Next incremental copy will begin in less than 1 minute'
           );
@@ -440,7 +441,6 @@ export class Plan {
     applyAction(name, restartButton);
     confirm(); //Added Confirm Button
     if (warmMigration) {
-      Plan.openList();
       this.waitForIncrementalCopies(planData);
       this.schedule_cutover(planData);
     }
--- a/pkg/qe-tests/cypress/integration/tests/Rhv/config_separate_mapping_rhv.ts
+++ b/pkg/qe-tests/cypress/integration/tests/Rhv/config_separate_mapping_rhv.ts
@@ -6,6 +6,7 @@ import {
   TestData,
   RhvProviderData,
   HookData,
+  CutoverData,
 } from '../../types/types';
 import { providerType, storageType } from '../../types/constants';
 const url = Cypress.env('url');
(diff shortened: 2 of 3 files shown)


kpunwatk:
@ibragins Please review this PR!

pkg/qe-tests/cypress/integration/models/plan.ts
```diff
@@ -350,7 +350,7 @@ export class Plan {
         .within(() => {
           cy.get('[data-label="Status"]', { timeout: 3600 * SEC }).should(
             'contain.text',
-            'Idle - Next incremental copy will begin in less than 1 minute'
+            'Idle - Next incremental copy will begin in '
```
sshveta, line 353: Why is the message incomplete ?

pkg/qe-tests/cypress/integration/models/plan.ts
```diff
@@ -440,7 +440,6 @@ export class Plan {
     applyAction(name, restartButton);
     confirm(); //Added Confirm Button
     if (warmMigration) {
-      Plan.openList();
```
sshveta, line 443: You don't have to open the plan list anymore ?

pkg/qe-tests/cypress/integration/models/plan.ts
ibragins: I guess that we need to do several pre-copies, at least 2. So please don't remove that time, leave it or update it to be 5-10 minutes, then put another one with for example 50 minutes. So when cutover will run - we will be sure that at least 2 precopies were done

pkg/qe-tests/cypress/integration/models/plan.ts
ibragins: I guess it was left here by mistake, it is not harmful but just not needed here, I talked with Karishma about this on previous review.

pkg/qe-tests/cypress/integration/models/plan.ts
kpunwatk: Thanks @sshveta and @ibragins for reviewing the PR.
In warm migration after it completes its first incremental pre-copy the status displays this message and we are using this status as an assertion in our code.
As the nextPrecopyAt timestamp is not static, it is calculating time until the next copy will start and we can't predict what will be the exact time-period it will start and hence I just implemented it just catching the text without the time-period.
As I observed something like this "begin in 2 minutes" and "begin in 59 minutes".

Thanks @ibragins for this suggestion I'll start working on it.

pkg/qe-tests/cypress/integration/models/plan.ts
kpunwatk: This is related to this merged PR https://github.com/konveyor/forklift-ui/pull/913. It was my mistate I missed to remove the " Plan.openList();" in previous PR. Previously did some modification to the code and hence covered inside it :)

yaacov:

PR #948, opened 2022-03-31
Bug 2067708: Support source VMs, networks and datastores referenced either by id or name in plan and mapping objects

mturley:
Currently, the UI expects all source resources (VMs, networks and datastores) in plan and mapping objects to be referenced by id rather than by name because the UI creates the objects this way. However, the CRDs support using either name or id for these references, and when creating them via the CLI or API with names, various bugs crop up in the UI such as https://bugzilla.redhat.com/show_bug.cgi?id=2067708.

This PR adjusts the TypeScript types for these references so that both `id` and `name` are present as optional fields, and adjusts any code that was relying on ids being present to fall back to names if ids are not present.


mturley:
@gildub @seanforyou23 if either of you guys have a moment it would be good to get this merged, but there's no hurry.

seanforyou23 (approved):
👍

PR #965, opened 2022-06-23
Added local storage preservation functionality

ibragins:
Now, tests can each next test can continue working as is instead of logging in from scratch each time.


mturley (approved):
LGTM

PR #968, opened 2022-06-27
Fixed migration from RHV

ibragins:
There was a mess with selectors when selecting cluster for VMWare or RHV, how selector is chosen depending on cluster type and tests will always work properly.


gildub (approved):
LGTM

PR #969, opened 2022-06-28
Enable minikube: Add AUTH_REQUIRED

mnecas:
Issue: Currently we always log in to the okd, which can be a problem for devs as not everybody has HW for okd.
This patch adds env var `AUTH_REQUIRED` which disables the login.
I have also removed the `Authorization` when communicating with the controller as it's no longer needed because we would be sending empty bearer token.

Example: `NO_AUTH=true DATA_SOURCE=remote npm run start:dev`

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -167,6 +167,7 @@ module.exports = (env) => {
         DATA_SOURCE: 'remote',
         BRAND_TYPE: 'Konveyor',
         NODE_ENV: 'production',
+        NO_AUTH: 'false',
       }),
     ],
     resolve: {
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -167,7 +167,6 @@ module.exports = (env) => {
         DATA_SOURCE: 'remote',
         BRAND_TYPE: 'Konveyor',
         NODE_ENV: 'production',
-        NO_AUTH: 'false',
       }),
     ],
     resolve: {
--- a/pkg/api/scripts/remote-dev.js
+++ b/pkg/api/scripts/remote-dev.js
@@ -3,7 +3,7 @@ const execSync = require('child_process').execSync;
 const helpers = require('../src/helpers');
 
 try {
-  if(process.env['NO_AUTH'] === 'true'){
+  if (process.env['NO_AUTH'] === 'true') {
     process.exit(0);
   }
   execSync('hash oc');
--- a/pkg/api/scripts/remote-dev.js
+++ b/pkg/api/scripts/remote-dev.js
@@ -3,6 +3,9 @@ const execSync = require('child_process').execSync;
 const helpers = require('../src/helpers');
 
 try {
+  if(process.env['NO_AUTH'] === 'true'){
+    process.exit(0);
+  }
   execSync('hash oc');
 } catch (error) {
   console.error(error.stdout.toString());
--- a/pkg/api/scripts/remote-dev.js
+++ b/pkg/api/scripts/remote-dev.js
@@ -3,7 +3,7 @@ const execSync = require('child_process').execSync;
 const helpers = require('../src/helpers');
 
 try {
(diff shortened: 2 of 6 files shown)


mnecas:
[WIP] because it still needs a second patch to the controller to allow empty Authorization: Bearer

pkg/web/src/app/queries/fetchHelpers.ts
```diff
@@ -5,7 +5,9 @@ import { History, LocationState } from 'history';
 import { useClientInstance } from '@app/client/helpers';
 import { KubeResource } from '@konveyor/lib-ui';
 import { IKubeResponse, IKubeStatus } from '@app/client/types';
+import { IEnvVars } from '@app/common/types';
 
+export const ENV: IEnvVars = window['_env'] || {};
```
mturley, line 10: Instead of declaring a new `ENV` in each file where it is used, we should continue to import the centralized `ENV` that is exported by `pkg/web/src/app/common/constants.ts`. You can see where this is already done in `ForkliftAboutModal`. You can import it like this (the `@app` import path is an alias for `pkg/web/src/app` in our tsconfig):

```ts
import { ENV } from '@app/common/constants';
```

Aside from saving you from importing `IEnvVars`, the reason for this is just in case we ever need/want to modify the way these environment variables are provided to / consumed by the UI, since the `window['_env']` global is a bit of a hack. We want to keep that hack isolated 😄

mturley:
One other thing @mnecas , the CI failure is due to Prettier code style being enforced by our linter. You can fix it by running `yarn format` or by running Prettier yourself (I use an extension for VS Code to run it automatically on save).

config/webpack.common.js
```diff
@@ -167,6 +167,7 @@ module.exports = (env) => {
         DATA_SOURCE: 'remote',
         BRAND_TYPE: 'Konveyor',
         NODE_ENV: 'production',
+        NO_AUTH: 'false',
```
mturley, line 170: You don't actually need `NO_AUTH` to be part of the `EnvironmentPlugin` if it's part of `FORKLIFT_ENV`. The other 3 variables are here because we want to be able to statically remove code based on those conditions (for example, webpack replaces `if (process.env['DATA_SOURCE'] === 'mock')` with `if (false)` and then deletes the unreachable code). They're also in `FORKLIFT_ENV` because that's a superset (having all the variables in the same place at run time is convenient).

mturley (commented):
I realized one more thing, sorry. Otherwise this LGTM now.

@jortel @fbladilo how do you feel about this addition? If you run the UI server with `NO_AUTH=true` it won't attempt to log users in and it'll make API requests with no bearer token. The use case of testing forklift on minikube seems valid to me.

pkg/web/src/app/routes.tsx
```diff
@@ -15,6 +15,9 @@ import { VMMigrationDetails } from '@app/Plans/components/VMMigrationDetails';
 import { LoginHandlerComponent } from './common/LoginHandlerComponent';
 import { RedirectToLogin } from './common/RedirectToLogin';
 import { NotFound } from './NotFound';
+import { IEnvVars } from './common/types';
+
+export const ENV: IEnvVars = window['_env'] || {};
```
mturley, line 20: Oh, we also can replace this one with a `import { ENV } from '@app/common/constants';` as well

pkg/api/scripts/remote-dev.js
```diff
@@ -3,6 +3,9 @@ const execSync = require('child_process').execSync;
 const helpers = require('../src/helpers');
 
 try {
+  if (process.env['NO_AUTH'] === 'true') {
```
jortel, line 6: Let's use the environment variable `AUTH_REQUIRED` for consistency with the controller and other konveyor projects.

mturley (approved):

Issue #868, opened 2021-12-21
Vsphere provider create/edit: Verify certificate button should reappear on error fetching certificate


mturley:
Followup from https://github.com/konveyor/forklift-ui/pull/845

mturley:
See also: https://github.com/konveyor/forklift-ui/pull/845#discussion_r772558406

> This could maybe also be an inline plain static live region info alert, if you wanted to communicate to screen reader users the loading state - since this one can take a while to resolve if there are issues with the network request.

We may want to do that as well to resolve this issue.

PR #977, opened 2022-08-04
Add a fetch action for oVirt CA certs in provider add/edit modal

sjd78:
### What is new?
Add a new "fetch the engine CA certificate" link to the Add/Edit provider modal.  Using an enhanced `/get-certificate` server route, the CA cert is fetched and populated to the form field.

ca cert link

The `/get-certificate` server route is enhanced to be able to fetch the CA certificate from an oVirt engine host.  Response data will match the response of the `vsphere` request, and is used to populate the certificate upload form field.  The `useCertificateQuery()` hook has been updated to match the server side enhancements.

None of the existing field or query functionality was changed.  A new query was added for the ovirt CA cert fetching.

Manual upload of the CA certificate is still possible.

### What is missing?
The mock data for the `useCertificateQuery()` hook probably needs to be updated.

### Questions

1. Is the way the CA cert fetch done in the label button click "the right way" for working with the react-query / useQuery libraries? Other suggestions for how to handle the fetch then update a few fields?

2. Does it make sense to show more details about the certificate like the vsphere provider?  Something like showing the cert's subject, issuer and fingerprint?

### Side notes...

The CA cert pulled from the engine is the cert that issues the cert used on the engine SSL connection.  So pulling the SSL cert, matching what is done for the vsphere provider, isn't enough.  This also means that any other forklift code should know the ovirt cert is the CA cert and not a connection's SSL cert.

### DIFF
--- a/pkg/web/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
+++ b/pkg/web/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
@@ -49,6 +49,7 @@ import {
 } from '@app/queries';
 
 import HelpIcon from '@patternfly/react-icons/dist/esm/icons/help-icon';
+import ExternalLinkSquareAltIcon from '@patternfly/react-icons/dist/esm/icons/external-link-square-alt-icon';
 import { IProviderObject } from '@app/queries/types';
 import { QuerySpinnerMode, ResolvedQuery } from '@app/common/components/ResolvedQuery';
 import { useAddEditProviderPrefillEffect } from './helpers';
@@ -214,13 +215,16 @@ export const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModal
     ? createProviderMutation
     : patchProviderMutation;
 
-  const [isCertificateQueryEnabled, setCertificateQueryEnabled] = React.useState(false);
-  const certificateQuery = useCertificateQuery(
+  const [isFingerprintQueryEnabled, setFingerprintQueryEnabled] = React.useState(false);
+  const fingerprintQuery = useCertificateQuery(
+    providerType,
     fields?.hostname?.value || '',
-    isCertificateQueryEnabled
+    isFingerprintQueryEnabled
   );
 
   const certificateConfirmButtonRef = React.useRef<HTMLElement>(null);
+
+  const caCertQuery = useCertificateQuery(providerType, fields?.hostname?.value || '', false);
 
   const scrollVerifyButtonIntoView = () =>
     certificateConfirmButtonRef.current?.scrollIntoView({ behavior: 'smooth' });
@@ -316,7 +320,7 @@ export const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModal
                     fieldId="hostname"
                     inputProps={{
                       onChange: (field: string) => {
-                        setCertificateQueryEnabled(false);
+                        setFingerprintQueryEnabled(false);
                         fields.isCertificateValid?.setValue(false);
                         fields.hostname?.setValue(field);
                       },
@@ -380,7 +384,7 @@ export const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModal
                 ) : null}
                 {fields?.fingerprint ? (
                   <>
-                    {!isCertificateQueryEnabled ? (
+                    {!isFingerprintQueryEnabled ? (
                       <Button
                         id="certificate-confirm-button"
                         key="confirm"
@@ -389,17 +393,17 @@ export const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModal
                         variant="primary"
                         isDisabled={!fields.hostname?.isValid}
                         onClick={() => {
-                          setCertificateQueryEnabled(true);
+                          setFingerprintQueryEnabled(true);
                         }}
                       >
                         Verify certificate
                       </Button>
-                    ) : certificateQuery.status === 'loading' ? (
+                    ) : fingerprintQuery.status === 'loading' ? (
                       <div className="pf-c-empty-state__icon">
                         <Spinner aria-labelledby="loadingPrefLabel" size="sm" />
                         &nbsp;Retrieving SHA-1 certificate fingerprint
                       </div>
-                    ) : certificateQuery.status === 'success' ? (
+                    ) : fingerprintQuery.status === 'success' ? (
                       <>
                         <Panel variant="bordered">
                           <PanelMain>
@@ -411,7 +415,7 @@ export const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModal
                                 <DescriptionListGroup>
                                   <DescriptionListTerm>Issuer</DescriptionListTerm>
                                   <DescriptionListDescription id="issuer">
-                                    {`${certificateQuery.data?.issuer.O} - ${certificateQuery.data?.issuer.OU}`}
+                                    {`${fingerprintQuery.data?.issuer.O} - ${fingerprintQuery.data?.issuer.OU}`}
                                   </DescriptionListDescription>
                                 </DescriptionListGroup>
                                 <DescriptionListGroup>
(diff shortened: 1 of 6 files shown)


mturley:
I may be forgetting the full context here, but isn't the point of asking the user to provide the certificate to prevent a man-in-the-middle attack? How does the user know the certificate that is fetched here is legitimate without going and finding it themselves to compare? cc @jortel @mansam 

Originally, the form for adding a vsphere provider looked similar (it had a field for the user to paste in the fingerprint rather than a button to fetch it). We added the `get-certificate` route as a convenience, but we still wanted the user to confirm that the details shown match the expected certificate, so if a user wants to be lazy and not double check it that's on them.

I think maybe we could use the same pattern here: instead of a button that prefills the certificate field, remove the certificate field and instead have the button behave like the one for vsphere (display details, ask user to confirm they are accurate, use the cached certificate result in submitting the form). cc @vconzola

mturley:
If we're touching that "verify certificate" button, FYI there's an open issue around it: https://github.com/konveyor/forklift-ui/issues/868

sjd78:
> I may be forgetting the full context here, but isn't the point of asking the user to provide the certificate to prevent a man-in-the-middle attack? How does the user know the certificate that is fetched here is legitimate without going and finding it themselves to compare? cc @jortel @mansam

There is always a trade off between convenience and security.  I'm curious as to what that kind of discussion sounded like between the PMs and UX and engineers.

> 
> Originally, the form for adding a vsphere provider looked similar (it had a field for the user to paste in the fingerprint rather than a button to fetch it). We added the `get-certificate` route as a convenience, but we still wanted the user to confirm that the details shown match the expected certificate, so if a user wants to be lazy and not double check it that's on them.
> 
> I think maybe we could use the same pattern here: instead of a button that prefills the certificate field, remove the certificate field and instead have the button behave like the one for vsphere (display details, ask user to confirm they are accurate, use the cached certificate result in submitting the form). cc @vconzola

And that does make sense.  Have the modal show some details of the certificate and make the user click a button that says "yeah what you fetched is correct".  I wouldn't take away the ability to paste in the certificate either.  Fetch, prefill and verify OR upload your own.

I'll adjust things following that pattern.

pkg/web/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
```diff
@@ -496,6 +501,32 @@ export const AddEditProviderModal: React.FunctionComponent<IAddEditProviderModal
...
+                        variant="link"
+                        isDisabled={!fields.hostname?.isValid}
+                        onClick={async () => {
+                          const { data } = await caCertQuery.refetch();
+                          if (data) {
+                            fields.caCert?.setValue(data.pemEncoded);
+                            fields.caCert?.setIsTouched(true);
+                            fields.caCertFilename?.setValue('pki-resource');
+                          }
+                        }}
```
sjd78, line 518: @mturley - Does this look reasonable?  It was the first way I found that made sense to run the cert fetch query and then update the contents of more that one field.

ahadas:
@sjd78 what's the benefit on fetching the certificate from the UI compared to making the certificate field optional and in that case letting the backend to fetch the certificate? the latter could also be useful from the cli

pkg/web/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx
mturley: Hmm, yeah I think this is reasonable. It's hard to say because having an imperative fetch button like this goes a little bit against the design of react-query (declaratively render loaded data, it will handle when to fetch under the hood) but that's necessary for this use case.

We may want to make sure we have a consistent UX and implementation with the way vsphere certificate fingerprints are handled (and maybe that means changing vsphere to behave more like this, bringing back the ability to paste the fingerprint). In the vsphere case, the button simply [sets boolean state](https://github.com/konveyor/forklift-ui/blob/6eb656453f271364211795ded932227bc91bbc20/pkg/web/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx#L387-L389) that enables the certificate query, and when that query succeeds it [renders certificate information and an "I trust this" checkbox](https://github.com/konveyor/forklift-ui/blob/6eb656453f271364211795ded932227bc91bbc20/pkg/web/src/app/Providers/components/AddEditProviderModal/AddEditProviderModal.tsx#L398-L470). When that checkbox is changed, that's when the hidden form field is actually updated with the fingerprint data. There's also a mechanism there when editing an existing provider where it can detect that the fingerprint has changed since the provider was created and show a warning.

Maybe since we're dealing with a fingerprint for vsphere and a whole CA cert for ovirt we can't make this 100% consistent, but I'm just sharing for comparison's sake. In that case since we didn't need to fill in the form field immediately as data was loaded we could get away with using the regular query `enabled` flag to control fetch time. But if you are going to have the field visible here, since you need to fetch and then fill in the field immediately after (and you don't want some `useEffect` overwriting the field later if the user edits it) your implementation makes sense to me.

Another option to imperatively get data from react-query if you know the data has already been fetched is to get a reference to the `queryClient` via [`useQueryClient()`](https://react-query-v3.tanstack.com/reference/useQueryClient) and then call `queryClient.getQueryData(queryKey)`. Similarly not ideal, but won't cause a refetch if you know you don't need one.

pkg/web/src/app/queries/providers.ts
```diff
@@ -398,13 +407,14 @@ const getCertificate = async (hostname: string) => {
...
 export const useCertificateQuery = (
+  providerType: ProviderType | null,
   hostname: string,
   enabled: boolean
 ): UseQueryResult<ITLSCertificate> => {
   const result = useMockableQuery<ITLSCertificate>(
     {
       queryKey: ['certificate', hostname],
-      queryFn: () => getCertificate(hostname),
+      queryFn: () => getCertificate(providerType, hostname),
```
mturley, line 417: One other thing I notice here: Because you're now using `useCertificateQuery` twice in the above component, we should make sure the underlying `queryKey` being used is unique for those 2 calls or else they'll clobber each other's cache. You can add as much unique identifying info to this `queryKey` array as you want, we should probably put `providerType` in there now.

mansam:
Automatic retrieval of a certificate defeats the purpose of trusting it. We had several discussions about automatically populating the certificate for the user in which I raised my objections to doing this sort of thing, and I still think this is a bad idea. It's one thing to have this happen in the UI where at least we can force the user to acknowledge what they are doing, but I am completely opposed to having the back end do automatic certificate retrieval. I'd be much more comfortable with making the user explicitly disable SSL certificate verification if they really don't care about it.

ahadas:
> To @ahadas's point: It's one thing to have this happen in the UI where at least we can force the user to acknowledge what they are doing, but I am completely opposed to having the back end do automatic certificate retrieval. I'd be much more comfortable with making the user explicitly disable SSL certificate verification if they really don't care about it.

I meant to do it as it's done in oVirt where the secured way is enabled by default but users can disable it if they wish to:
vsphere_provider

if you wish to connection via insecure protocol vs. secure protocol that uses the certificate that is automatically retrieved by forklift from ovirt-engine, I don't mind.. I don't think it really matters

mansam:
I think we're talking about the same thing then. I was referring to disabling verification of the certificate as with that screenshot, not communicating over an entirely insecure protocol.

ahadas:
Yeah, I think you didn't parse my original comment correctly.. My point was that we see based on resources we get from the field that it's often the case that users don't need that extra security when migrating workloads from vSphere to RHV, which would probably also be the case for forklift users and for those users we should simplify the process as much as possible - another click for fetching the certificate and showing it on the frontend side would probably not give much to those users

mansam:
@ahadas Agreed then, apologies for the misunderstanding.

mturley:
I agree with @mansam and @ahadas, and it sounds like the best UX for this would be to replicate what you have in that RHV screenshot -- a checkbox for whether to verify, and if the checkbox is unchecked just hide the field. We'd have to adjust the backend to handle the lack of a certificate being passed, rather than being able to fetch the certificate from the UI. I wonder if we should just do the same thing for the vsphere case to be consistent. cc @vconzola

yaacov:
> We'd have to adjust the backend to handle the lack of a certificate being passed,

@mturley @ahadas @mansam hi, I posted a PR to allow the backend to run with oVIrt `insecureSkipVerify`
https://github.com/konveyor/forklift-controller/pull/454 PTAL

yaacov:
konveyor/forklift-controller#454 - merged

with current upstream forklift-controller, we can add a new field to the secret:
`insecureSkipVerify` if it's set to true, the certificate will be ignored ( it can still exist in the secret or can be removed )
so the UI can look like suggested in https://github.com/konveyor/forklift-ui/pull/977#issuecomment-1208305035 
the checkbox will control the `insecureSkipVerify` field in the secret, and if checked the cert value can be ignored.

yaacov:

PR #979, opened 2022-08-10
Upgrade `@konveyor/lib-ui` to 8.3.2, Upgrade PatternFly to latest, replace ValidatedPasswordInput and LocalStorageContext with library versions

mturley:
This PR upgrades our `@konveyor/lib-ui` (https://github.com/konveyor/lib-ui/) dependency from 6.0.0 to 8.3.2 and refactors away some code that has now been abstracted into that library.

The breaking changes in lib-ui 7.0.0 and 8.0.0 (found on the [Releases page](https://github.com/konveyor/lib-ui/releases) of that repo) do not affect forklift-ui:
* [In v7.0.0](https://github.com/konveyor/lib-ui/releases/tag/v7.0.0):

  > ValidatedTextInput: Any direct calls to getTextInputProps or getFormGroupProps that were passing a
boolean for greenWhenValid will now need to instead pass an options object including greenWhenValid
as a property.

  * We're not using the `greenWhenValid` option in forklift-ui.
  
* [In v8.0.0](https://github.com/konveyor/lib-ui/releases/tag/v8.0.0):

  > useFormState: Any forms using the yupOptions argument will need to wrap it in an object with the
property yupOptions instead of passing it directly as a top-level argument.

  * We're not using the `yupOptions` argument in forklift-ui.

This PR makes the following changes to leverage new features:
* The original `ValidatedPasswordInput` component has been replaced by the version added to lib-ui in [v6.4.0](https://github.com/konveyor/lib-ui/releases/tag/v6.4.0).
* In all all instances of `ValidatedTextInput` and `ValidatedPasswordInput` where the `label` prop was redundant with the `.label()` included in the field's yup schema, the prop has been removed. Those components now default the label to the one from the schema if present (added to `ValidatedTextInput` in [v6.2.0](https://github.com/konveyor/lib-ui/releases/tag/v6.2.0) and added to `ValidatedPasswordInput` in [v8.0.1](https://github.com/konveyor/lib-ui/releases/tag/v8.0.1))
  * The one special case here is in `PlanAddEditHookModal`, where we have a field that should have no displayed label since it would be repeating the label on the radio button just above the field. In this case, `label={null}` has been added to retain the intentionally missing label.
* The overly complex `LocalStorageContext` has been replaced by the `useLocalStorage` hook added to lib-ui in [v8.3.0](https://github.com/konveyor/lib-ui/releases/tag/v8.3.0). See the [documentation here](https://konveyor.github.io/lib-ui/?path=/docs/hooks-uselocalstorage--persistent-counter-example) for more information on this hook, and see its [source here](https://github.com/konveyor/lib-ui/blob/main/src/hooks/useLocalStorage/useLocalStorage.ts).
  * React Context is no longer required for this implementation, since multiple instances of `useLocalStorage` which use the same unique `key` will keep their values in sync by piggybacking on the `StorageEvent`.
  * As part of this refactor, the `currentUser` value that was previously serialized to JSON before passing it to `LocalStorageContext` is now passed around in its real object form everywhere, since `useLocalStorage` handles JSON serialization internally. In the process, I realized there was actually a mistake in the TypeScript annotation of `saveLoginToken()`, its `user` param is annotated as a string and is actually the object form (it is parsed from JSON before being passed in, and it is stringified before use). This didn't cause an error because the annotations were consistently incorrect, and it never bit us because we never try to use the value in the scope where its type is incorrect. This has been fixed now since `currentUser` is always an object.
  * These changes are backwards-compatible with any old localStorage values a user might still have in their session after upgrading Forklift. The same keys are still in use, and the stored values were already being serialized to JSON (or the JSON-valid `'true'` and `'false'` strings), that same behavior has just moved to the library.


pkg/web/src/app/queries/fetchHelpers.ts
```diff
@@ -30,7 +30,7 @@ export const authorizedFetch = async <TResponse, TData = unknown>(
   const { history, checkExpiry } = fetchContext;
 
   if (ENV.AUTH_REQUIRED !== 'false') {
-    extraHeaders['Authorization'] = `Bearer ${fetchContext.currentUser.access_token}`;
+    extraHeaders['Authorization'] = `Bearer ${fetchContext.currentUser?.access_token}`;
```
yaacov, line 33: should we throw an error if `fetchContext.currentUser` is null or missing access_token ?

yaacov (approved):
looks great

pkg/web/src/app/queries/fetchHelpers.ts
mturley: Good question. I don't think it should be possible to hit that case here since routes which need to make authorized requests are also `isProtected`, and if `currentUser` is null the UI will kick the user to the login screen when they try to reach those pages. But in the case where a mistake is made and this is called with a null user / missing token, the request will go through with `Authorization: Bearer [blank]` and then fail with a 401, which will be reported in the component as any other API error would.

Maybe instead we should just not try to add that header at all if the token isn't defined?

mturley: Pushed a commit to that effect.

mturley:
Pushed a rebase, looking into CI failure now.

mturley:
@yaacov I added a commit to upgrade PatternFly to latest (we were using versions older than the peerDependencies in the latest version of `@konveyor/lib-ui`), and this broke a test due to the `aria-role` on Tab buttons changing so I fixed the broken test as well. Would you mind re-reviewing?

yaacov (approved):

PR #982, opened 2022-08-23
Add url to secret for the admission webhook

mnecas:
Add the URL for the provider to the secret so that when the controller starts
the provider connection test knows which address to validate.
The webhook is watching specific secrets with specific labels so
we need to post the provider secrets with the label "createdForResourceType=providers".
Related to:
- https://github.com/konveyor/forklift-controller/pull/460
- https://github.com/konveyor/forklift-operator/pull/247


yaacov:
missing `,` in `helpers.ts`, o/w lgtm

yaacov:
IMHO it's ok to merge this before the backend PRs, @mnecas please keep me honest here.

mnecas:
yeah, this can even should go before the others but want to merge them all at the same time.
After everyone agrees on var names etc. (if we would decided to change the name of label would need to change it everywhere)

yaacov:

PR #983, opened 2022-08-24
Allow an optional static bearer token when using dev server

yaacov:
When running on dev environment, sometimes it's useful to allow to provide static bearer token.
For example if running the UI with minikube[1]

In my use case I start minishift with a dev user that has some bearer token, and then run
``` bash
AUTH_REQUIRED=false yarn start:dev:remote
```

with this patch the dev server command will look something like:
``` bash
K8S_AUTH_BEARER_TOKEN=31ada4fd-adec-460c-809a-9e56ceb75269 AUTH_REQUIRED=false yarn start:dev:remote
```

[1] https://github.com/konveyor/forklift-console-plugin/blob/main/scripts/configure-minikube.sh


yaacov:
@mturley @sjd78 @rszwajko please review

pkg/web/src/app/client/helpers.ts
```diff
@@ -203,7 +203,7 @@ export const checkIfResourceExists = async (
 export const useClientInstance = (): KubeClient.ClusterClient => {
   const { currentUser } = useNetworkContext();
   const user = {
-    access_token: currentUser?.access_token || '',
+    access_token: currentUser?.access_token || window['_meta']?.clusterBearerToken || '',
```
mturley, line 206: Instead of directly referencing `window['_meta']`, the properties from this file should be added to the `IMetaVars` type [here](https://github.com/konveyor/forklift-ui/blob/main/pkg/web/src/app/common/types.ts#L4-L17) and fallback defaults for each should be added to the exported `META` variable [here](https://github.com/konveyor/forklift-ui/blob/main/pkg/web/src/app/common/constants.ts#L79-L95). Then this file can `import { META } from '@app/common/constants';` and you can use `META.clusterBearerToken` here.

That way we get strong types for these and don't need the `?.` and the `|| ''`.

mturley (changes requested):

pkg/web/src/app/client/helpers.ts
mturley: Actually, alternatively you could use an environment variable for this. the meta.json file is generated by the operator and generally used for things that also apply to prod but can be overridden in dev. If it's an environment variable it could be overridden on the command line if necessary. I honestly can't remember exactly why we did both (@ibolton336 may remember), and the distinction has probably been broken, but just so you're aware of both, to define a new environment variable you add it to the `IEnvVars` type [here](https://github.com/konveyor/forklift-ui/blob/main/pkg/web/src/app/common/types.ts#L19-L31), use `import { ENV } from '@app/common/constants';` (defined [here](https://github.com/konveyor/forklift-ui/blob/main/pkg/web/src/app/common/constants.ts#L5)), and make sure to add it to the list on the server [here](https://github.com/konveyor/forklift-ui/blob/main/pkg/api/src/helpers.js#L36-L56) so it gets bundled up and sent to the client.

Both these `META` and `ENV` mechanisms could probably be combined and/or replaced, they were hacks that stuck around... but the strong types are useful.

yaacov: Thanks :+1:, I like the environment variable approach better then the meta file for a token parameter,
changed the code to use the ENV, and added the type as recomended/

yaacov: do we document env vars someware ?

yaacov: added documentation to README.md file

rszwajko (approved):

mturley:
Agh sorry, looks like I caused a merge conflict by merging #984

yaacov:
@mturley np :-), rebased

mturley (approved):

PR #986, opened 2022-08-30
Use the K8S_AUTH_BEARER_TOKEN env var for authorizedFetch

yaacov:
Issue:
In https://github.com/konveyor/forklift-ui/pull/983 I added the optional `K8S_AUTH_BEARER_TOKEN` environment variable, but forgot to add it to the `authorizedFetch` method.

Solution:
Adding a check, if the UI is running without OAuth, and the `K8S_AUTH_BEARER_TOKEN` environment variable is provided, use it for authentication in the `authorizedFetch` method.

Signed-off-by: yzamir <yzamir@redhat.com>


mturley (approved):
LGTM

PR #987, opened 2022-09-13
Use --legacy-peer-deps flag in Dockerfile

mturley:
Due to the same issue described in https://github.com/konveyor/forklift-ui/pull/962, image builds are failing when using newer versions of NodeJS / npm because of peer dependency resolution issues. For the time being we need to use the `--legacy-peer-deps` flag to switch to the old peer deps resolution behavior when installing forklift-ui dependencies. #962 did this in the CI scripts but we also need it in the Dockerfile.

cc @fbladilo

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 FROM registry.access.redhat.com/ubi8/nodejs-16 as builder
 COPY . .
 USER root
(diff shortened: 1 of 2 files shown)


fbladilo (approved):
LGTM

ibolton336 (approved):

